PR #491, opened 2026-04-12
fix: remove HLC from snapshotTS to prevent lost-update on concurrent writes

bootjp:
snapshotTS used max(lastCommitTS, HLC.Current()) as the read snapshot timestamp. When a concurrent write advanced the HLC via clock.Next() to obtain its commitTS, the next reader's readTS (used as startTS for conflict detection) could equal that commitTS. The FSM write-conflict check (latestTS > startTS) then missed the conflict because startTS == latestTS, allowing the second write to silently overwrite the first.

Now snapshotTS uses only store.LastCommitTS(), which accurately reflects the committed state visible to readers. This ensures startTS < any concurrent transaction's commitTS, so the > check correctly detects write-write conflicts.

Found by Jepsen: lost-update, G2-item-realtime, strong-PL-1-cycle-exists anomalies on Redis append workload under concurrency=8, rate=10.

### DIFF
--- a/adapter/ts.go
+++ b/adapter/ts.go
@@ -5,19 +5,17 @@ import (
 	"github.com/bootjp/elastickv/store"
 )
 
-// snapshotTS picks a safe snapshot timestamp:
-// - uses the store's last commit watermark if available,
-// - otherwise the coordinator's HLC current value,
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

PR #512, opened 2026-04-15
fix: use safe snapshot timestamp in txnStartTS to prevent lost-write anomaly

bootjp:
clock.Next() can be ahead of store.LastCommitTS() because concurrent dispatchTxn calls advance the HLC before their Raft entry is applied to Pebble. When txnStartTS returned clock.Next() = T, a concurrent RPUSH that had obtained commitTS = T-1 could still be in the Raft pipeline. The transaction would read stale list metadata, pick the same sequence number, and overwrite the concurrent value — a lost write. The FSM conflict check (latestTS > startTS) silently passed because T-1 < T.

Fix: return maxTS (= store.LastCommitTS()) as the startTS directly. store.LastCommitTS() is only advanced after the Pebble batch commit, so every version at <= maxTS is guaranteed visible. Any concurrent write at
> maxTS triggers WriteConflict and a retry via retryRedisWrite.
The Observe(maxTS) call is preserved so dispatchTxn clock.Next() still produces a valid commitTS > maxTS.

This closes the root cause of the Jepsen G1c / lost-write anomalies observed in list-append workloads under concurrent MULTI/EXEC and direct RPUSH operations.

### DIFF
--- a/.github/workflows/jepsen-test-scheduled.yml
+++ b/.github/workflows/jepsen-test-scheduled.yml
@@ -23,6 +23,10 @@ on:
         description: "Maximum writes per key before exhaustion"
         required: false
         default: "150"
+      max-txn-length:
+        description: "Maximum micro-ops per transaction"
+        required: false
+        default: "4"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-jepsen-scheduled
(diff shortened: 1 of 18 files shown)


bootjp:
@copilot apply changes based on the comments in [this thread](https://github.com/bootjp/elastickv/pull/512#pullrequestreview-4113513650)

PR #538, opened 2026-04-19
chore(pebble): pin FormatMajorVersion to ratchet DBs for pebble v2

bootjp:
## Summary

Prep step for #289 (pebble v1.1.5 → v2.1.4 bump).

Pebble v2 sets `FormatMinSupported = FormatFlushableIngest` and refuses to open DBs below that. Today both pebble instances in this repo open with defaults:

- `internal/raftstore/pebble.go` — Raft log / stable store (`raft.db`)
- `store/lsm_store.go` — MVCC FSM store

Pebble v1 treats an unset `FormatMajorVersion` as `FormatMostCompatible`, so any DB written by current binaries is **unreadable by pebble v2**. Merging #289 as-is would brick existing clusters on restart.

This PR pins `FormatMajorVersion` to `pebble.FormatVirtualSSTables` in both opens. That is the newest format pebble v1.1.5 supports and is well above v2's minimum, so:

- Fresh DBs are created at a v2-compatible format.
- Existing DBs are auto-ratcheted by pebble on open.

`FormatVirtualSSTables` was picked as an explicit, stable constant rather than `FormatNewest` so the intent is auditable and doesn't drift with future pebble patch releases.

## Rollout order

1. **Ship this PR on pebble v1.1.5** to every node (rolling restart is fine).
2. Verify that each node has successfully opened its DB at least once after running this build — the format ratchet happens during `pebble.Open`.
3. Only then merge #289 to bump pebble to v2.

If step 2 is skipped on even one node, that node will fail to start after the v2 bump.

## Test plan

- [x] `go test ./internal/raftstore/... ./store/...` — passes
- [x] `go test ./...` — passes
- [x] `golangci-lint run ./internal/raftstore/... ./store/...` — clean
- New regression tests in `internal/raftstore/pebble_test.go`:
  - `TestPebbleStore_FormatRatchetedForV2` — fresh DB opens ≥ `FormatFlushableIngest`
  - `TestPebbleStore_RatchetsExistingDB` — reopening a DB created at `FormatMostCompatible` ratchets it up
  - `TestPebbleStore_PreservesLogsAcrossRatchet` — committed Raft log entries survive the ratchet

## Summary by CodeRabbit

* **Improvements**
  * Enhanced database initialization with stricter format version configuration to ensure consistent and predictable behavior during database format transitions and version updates.

* **Tests**
  * Added test coverage validating that data persists correctly across database format version changes, including verification that existing data remains intact and accessible following format version updates.

### DIFF
--- a/internal/raftstore/pebble.go
+++ b/internal/raftstore/pebble.go
@@ -42,9 +42,11 @@ func NewPebbleStore(dir string) (*PebbleStore, error) {
 // Pinned so existing v1-era DBs are ratcheted above pebble v2's
 // FormatMinSupported (FormatFlushableIngest) before the v2 upgrade lands.
 func pebbleOptions() *pebble.Options {
-	return &pebble.Options{
+	opts := &pebble.Options{
 		FormatMajorVersion: pebble.FormatVirtualSSTables,
 	}
+	opts.EnsureDefaults()
+	return opts
 }
 
 func (s *PebbleStore) FirstIndex() (uint64, error) {
--- a/internal/raftstore/pebble.go
+++ b/internal/raftstore/pebble.go
@@ -32,11 +32,19 @@ func NewPebbleStore(dir string) (*PebbleStore, error) {
 		return nil, errors.WithStack(err)
 	}
 
-	db, err := pebble.Open(dir, &pebble.Options{})
+	db, err := pebble.Open(dir, pebbleOptions())
 	if err != nil {
 		return nil, errors.WithStack(err)
 	}
 	return &PebbleStore{db: db}, nil
+}
+
+// Pinned so existing v1-era DBs are ratcheted above pebble v2's
+// FormatMinSupported (FormatFlushableIngest) before the v2 upgrade lands.
+func pebbleOptions() *pebble.Options {
+	return &pebble.Options{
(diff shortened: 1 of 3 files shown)


bootjp:
Thanks for the reviews. Pushed a357703a addressing the actionable items:

**Applied (Gemini, medium priority):**
- `internal/raftstore/pebble.go`: call `EnsureDefaults()` in `pebbleOptions()` to match the `store/lsm_store.go` site.
- `internal/raftstore/pebble_test.go`: `t.Cleanup` now surfaces `Close()` errors via `require.NoError` instead of swallowing them.

**Not applying (CodeRabbit nitpicks):**
- *Table-driving `TestPebbleStore_FormatRatchetedForV2` + `TestPebbleStore_RatchetsExistingDB`*: the two cases differ in setup (one pre-opens a `FormatMostCompatible` DB), so folding them into a table obscures the intent for only two rows. Preferring to keep them as named behaviors.
- *Extracting an `internal/pebbleopts` package*: the reviewer explicitly marks this non-blocking, and with only two call sites on a one-time migration I would rather not introduce a package just for this. If more Pebble opens appear later, a helper can be added then.

PR #541, opened 2026-04-19
chore(deps): upgrade pebble to v2.1.4

bootjp:
## Summary

Upgrades `github.com/cockroachdb/pebble` from v1.1.5 to v2.1.4.

Supersedes the stale Renovate PR #289, which:
- Only bumped `go.mod` / `go.sum` but kept imports under the v1 module path (pebble v2 moved the module to `github.com/cockroachdb/pebble/v2`).
- Predates #538 and therefore reverts the `FormatMajorVersion` pinning that was the entire point of the two-stage rollout.

## What changed

- `go.mod` / `go.sum`: pin `github.com/cockroachdb/pebble/v2` v2.1.4; drop v1 indirect deps that v2 no longer pulls in; add v2-era indirects.
- All Go source + test imports updated to `github.com/cockroachdb/pebble/v2` and `github.com/cockroachdb/pebble/v2/vfs`.
- `internal/raftstore/pebble_test.go`:
  - Dropped `TestPebbleStore_RatchetsExistingDB` and `TestPebbleStore_PreservesLogsAcrossRatchet` — both opened a DB at `FormatMostCompatible` to simulate pre-migration state, but v2 unexports that constant and refuses to open below `FormatFlushableIngest`. The behavior they guarded (the ratchet) was the job of #538 and has already been validated and deployed.
  - Renamed the remaining test to `TestPebbleStore_OpensAtPinnedFormat` and tightened it to assert equality with `FormatVirtualSSTables`, so an unintended downgrade of the pinned version is caught.

## Migration safety

Rollout prerequisite already satisfied: #538 is merged and deployed cluster-wide. Every node has opened its `raft.db` at least once under the prep binary, so on-disk format is ≥ `FormatVirtualSSTables` — comfortably above v2 `FormatMinSupported` (`FormatFlushableIngest`). v2 will open existing DBs without needing any further ratchet.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (all packages pass locally, including `store/...` MVCC + snapshot tests and `internal/raftstore/...`)
- [x] `golangci-lint run ./... --timeout=5m` — 0 issues
- [x] `go mod tidy` — no further changes

## Follow-up

PR #289 should be closed with a pointer to this PR.


bootjp:
Thanks for the review. Pushed 270c433c addressing both items:

**Applied:**
- Dropped the redundant `GreaterOrEqual(..., FormatFlushableIngest)` assertion in `TestPebbleStore_OpensAtPinnedFormat` — the `Equal(..., FormatVirtualSSTables)` check subsumes it.
- Added `TestPebbleStore_PersistsLogsAcrossReopen` as a replacement for the dropped ratchet persistence test. It does not try to reproduce the pre-migration DB state (v2 cannot create one), but it still exercises `StoreLog` / `GetLog` / `LastIndex` end-to-end across a close/reopen so the wrapper ↔ pebble v2 integration stays covered.

PR #544, opened 2026-04-19
feat(s3): accept aws-chunked streaming payload uploads

bootjp:
## Summary
- Accept aws-chunked framed PUT / UploadPart bodies so aws-sdk-go-v2 default clients stop getting `XAmzContentSHA256Mismatch`.
- Recognise `STREAMING-UNSIGNED-PAYLOAD-TRAILER`, `STREAMING-AWS4-HMAC-SHA256-PAYLOAD`, and its `-TRAILER` sibling as payload sentinels (same skip-the-body-hash treatment as `UNSIGNED-PAYLOAD`).
- Optionally verify `x-amz-checksum-crc32` / `-crc32c` trailers when the client advertises them via `X-Amz-Trailer`; mismatch returns 400 `BadDigest`.
- Strip `aws-chunked` from the stored `Content-Encoding` so GET responses don't advertise a transport-only encoding.

## Why
aws-sdk-go-v2 ≥ ~v1.90 defaults to `Content-Encoding: aws-chunked` + `X-Amz-Content-Sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER` over HTTPS. Until now, butoGPT and every other consumer had to disable SDK streaming (`AWS_REQUEST_CHECKSUM_CALCULATION=when_required`) to upload anything to the Caddy-fronted `s3-elk.bootjp.dev`. Accepting the native format lets clients run with stock defaults.

## What's *not* verified (intentional, documented)
- Per-chunk `chunk-signature=` values: parsed and discarded. HTTPS + request-header SigV4 already protect transport, so matching AWS's strict validation would cost implementation time with little local benefit. A future PR can enforce it behind a flag.
- The optional trailer signature accompanying `STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER` is also ignored; the checksum trailer (`x-amz-checksum-*`) is the authoritative body integrity check.

## Security properties kept
- Per-chunk size is capped at 16 MiB so a malicious peer cannot force an unbounded allocation from a crafted hex size line.
- Decoded byte total is compared for strict equality against `X-Amz-Decoded-Content-Length` when declared.
- Raw transport is bounded by `MaxBytesReader` (decoded-max + 1 MiB framing overhead); decoded bytes are bounded independently inside the chunked reader.
- Truncated streams surface as `io.ErrUnexpectedEOF` wrapped in `awsChunkedError`, so the handler returns 400 `InvalidRequest` instead of silently committing a truncated object.

## Test plan
- [x] `go test ./adapter/... -count=1` (full suite, 40 s)
- [x] `golangci-lint run --config=.golangci.yaml ./adapter/...` → 0 issues
- [x] Unit tests cover: single chunk, multi-chunk, signed-chunk extensions ignored, trailer exposed, decoded-length mismatch (over + under), max-decoded cap, oversized chunk header, malformed hex size, missing CRLF, truncated stream.
- [x] Integration tests run `server.handle()` end-to-end with `encodeAwsChunked(...)` bodies: happy path, good CRC32 trailer, corrupted CRC32 trailer, lying `X-Amz-Decoded-Content-Length`.

## Release Notes

* **New Features**
  * Enabled streaming uploads with automatic checksum verification to ensure data integrity during transfer.
  * Improved request validation and error handling for chunked uploads with better detection of incomplete or malformed requests.
  * Enhanced protection against data corruption from streaming transfer failures.

* **Tests**
  * Added comprehensive test coverage for streaming upload scenarios and edge cases.

### DIFF
--- a/adapter/s3.go
+++ b/adapter/s3.go
@@ -17,6 +17,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/httputil"
+	"net/textproto"
 	"net/url"
 	"strconv"
 	"strings"
@@ -777,7 +778,7 @@ func (s *S3Server) putBucketAcl(w http.ResponseWriter, r *http.Request, bucket s
 	w.WriteHeader(http.StatusOK)
 }
 
-//nolint:cyclop,gocognit,nestif // The S3 PUT flow is intentionally linear and maps directly to protocol steps.
+//nolint:cyclop,gocognit,gocyclo,nestif // The S3 PUT flow is intentionally linear and maps directly to protocol steps.
 func (s *S3Server) putObject(w http.ResponseWriter, r *http.Request, bucket string, objectKey string) {
 	readTS := s.readTS()
 	startTS := s.txnStartTS(readTS)
@@ -809,8 +810,12 @@ func (s *S3Server) putObject(w http.ResponseWriter, r *http.Request, bucket stri
 	hasher := md5.New() //nolint:gosec // S3 ETag compatibility requires MD5.
 	sha256Hasher := sha256.New()
 	expectedPayloadSHA := normalizeS3PayloadHash(r.Header.Get("X-Amz-Content-Sha256"))
-	validatePayloadSHA := expectedPayloadSHA != "" && !strings.EqualFold(expectedPayloadSHA, s3UnsignedPayload)
-	r.Body = http.MaxBytesReader(w, r.Body, s3MaxObjectSizeBytes)
+	validatePayloadSHA := expectedPayloadSHA != "" && !isS3PayloadMarker(expectedPayloadSHA)
+	streamBody, bodyErr := prepareStreamingPutBody(w, r, s3MaxObjectSizeBytes, expectedPayloadSHA)
+	if bodyErr != nil {
+		writeS3Error(w, bodyErr.Status, bodyErr.Code, bodyErr.Message, bucket, objectKey)
+		return
+	}
 	part := s3ObjectPart{PartNo: 1}
 	sizeBytes := int64(0)
 	chunkNo := uint64(0)
@@ -857,6 +862,7 @@ func (s *S3Server) putObject(w http.ResponseWriter, r *http.Request, bucket stri
 					return
 				}
 			}
+			streamBody.writeDecoded(chunk)
 			chunkKey := s3keys.BlobKey(bucket, meta.Generation, objectKey, uploadID, part.PartNo, chunkNo)
 			pendingBatch = append(pendingBatch, &kv.Elem[kv.OP]{Op: kv.Put, Key: chunkKey, Value: chunk})
 			chunkSize, err := uint64FromInt(n)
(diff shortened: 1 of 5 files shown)


bootjp:
Addressed all four gemini review items in ba5848ce:
- HIGH (premature EOF at chunk boundary): switched the unexpected-EOF guard from `r.cur > 0` to `!r.finished` so EOF after a chunk's trailing CRLF but before the terminator also raises. New regression test `TestAwsChunkedReader_TruncatedAtChunkBoundary`.
- SEC HIGH (unbounded line read): replaced `bufio.Reader.ReadString('\n')` with `ReadSlice('\n')` against an 8 KiB pre-sized buffer; `bufio.ErrBufferFull` now short-circuits over-long lines before they allocate. New test `TestAwsChunkedReader_OversizedLineRejectedWithoutUnboundedAlloc` feeds 32 KiB without a newline.
- MEDIUM (bufio buffer sizing): `bufio.NewReaderSize(r, s3ChunkedReaderBufBytes)`.
- MEDIUM (hashesEqual): dropped the hand-rolled comparator in favor of `bytes.Equal`. Constant-time comparison isn't needed here since the trailer values are non-secret checksums.

bootjp:
Addressed in 484577fb: X-Amz-Trailer is now split on commas (per AWS docs), each entry is trimmed, and we bind to the first one whose checksum algorithm we support. Added `TestS3Server_PutObjectStreamingWithMultipleAdvertisedTrailers` with the list "x-amz-checksum-crc32, x-amz-checksum-sha256" as a regression.

bootjp:
Round 3 addressed in 878fdb6d:

**coderabbit**
- MAJOR (signed streaming markers): `isS3StreamingPayloadMarker` now only accepts `STREAMING-UNSIGNED-PAYLOAD-TRAILER`. Added `isS3SignedStreamingPayloadMarker` and `prepareStreamingPutBody` returns 501 `NotImplemented` for the signed variants — safer default until chunk-signature verification is actually implemented.
- MAJOR (aws-chunked without streaming marker + framing overhead): added `contentEncodingContains` helper; requests with `Content-Encoding: aws-chunked` but no streaming sentinel now get 400 `InvalidRequest`. Replaced the fixed 1 MiB framing cap with `maxDecoded/256 + 4 KiB` (~0.4% headroom that scales with object size), clamped down to `r.ContentLength` when tighter.
- MINOR (tooLargeMessage): threaded `tooLargeMessage` through `prepareStreamingPutBody`; PutObject passes "object...", UploadPart passes "part...".
- MINOR (assertion): `TestS3Server_PutObjectStreamingRejectsBadDecodedLength` now asserts 400 + `<Code>InvalidRequest</Code>` exactly.
- nit (UploadPart coverage): added `TestS3Server_UploadPartStreamingWithTrailerChecksum` exercising CreateMultipartUpload → UploadPart with 5 MiB aws-chunked body + good and corrupted CRC32 trailers.

**gemini**
- HIGH (NopCloser leak): replaced with `closeForwardingReader{Reader: decoder, Closer: originalBody}` so `net/http` still calls Close on the underlying MaxBytesReader at request end.

Also added negative tests: `TestS3Server_PutObjectStreamingRejectsSignedPayloadMarker` and `TestS3Server_PutObjectRejectsAwsChunkedWithoutStreamingMarker`. test + lint green.

bootjp:
This review looks stale — it was submitted at 16:13:20Z against commit `484577fb`, but `878fdb6d` (pushed at 16:12:04Z, ~1m before the review landed) already replaced `io.NopCloser` with a `closeForwardingReader` wrapper that delegates `Close()` to the underlying `MaxBytesReader`. 

Current code at adapter/s3.go:2448:
```go
// Preserve Close() on the outer body so net/http still drains and
// frees the underlying MaxBytesReader + transport when the handler
// returns (io.NopCloser would have dropped that on the floor).
r.Body = closeForwardingReader{Reader: reader, Closer: r.Body}
```

which is the exact pattern gemini suggested. No further change needed.

PR #548, opened 2026-04-19
perf(lua): eliminate redundant ZSet member-prefix scan in zsetState

bootjp:
## Summary
- `zsetState` previously scanned `ZSetMemberScanPrefix` twice — once via `keyTypeAt → rawKeyTypeAt → detectWideColumnType`, then again to choose wide-column vs legacy-blob format.
- New `zsetStorageHintAt` (with `rawZSetPhysTypeAt` + `zsetMetaOrDeltaExistsAt` helpers) probes the member prefix once and carries the `memberFound` signal back, also folding the TTL-expired-key detection into the same hint.
- Common ZSet-with-members path: 2 `ScanAt` -> 1 `ScanAt`.
- TTL-expired ZSet path: 3 probe calls -> 1-2 probe calls.

## Motivation
Addresses Copilot review comment on PR #547 (line 689) about redundant I/O in `zsetState`.

## Test plan
- [ ] `go test ./adapter/... -run ZSet`
- [ ] Existing TTL-expired recreation and delta-commit-on-existing-wide-column tests still pass

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal type detection logic to reduce redundant operations and improve efficiency.
  * Enhanced caching mechanisms for type resolution to support faster lookups.
  * Improved TTL-aware type handling for expired keys.

### DIFF
--- a/adapter/redis_compat_helpers.go
+++ b/adapter/redis_compat_helpers.go
@@ -61,6 +61,23 @@ func normalizeStartTS(ts uint64) uint64 {
 // correctly detect collections whose fields were all deleted (metadata key exists but
 // no member keys) or newly created collections that only have delta keys.
 func (r *RedisServer) detectWideColumnType(ctx context.Context, key []byte, readTS uint64) (redisValueType, error) {
+	if typ, err := r.detectWideColumnTypeSkipZSet(ctx, key, readTS); err != nil || typ != redisTypeNone {
+		return typ, err
+	}
+	if found, err := r.wideColumnTypeExists(ctx, key, readTS, store.ZSetMemberScanPrefix, store.ZSetMetaKey, store.ZSetMetaDeltaScanPrefix); err != nil {
+		return redisTypeNone, err
+	} else if found {
+		return redisTypeZSet, nil
+	}
+	return redisTypeNone, nil
+}
+
+// detectWideColumnTypeSkipZSet runs the wide-column hash / set probes
+// only. Callers that have already eliminated ZSet (e.g.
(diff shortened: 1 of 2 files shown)


bootjp:
Rebased onto latest main and addressed gemini-code-assist's review (force-push, commit `5b0c56de`):

**Correctness (regression caught during rebase)**
- `zsetState` now consults `c.cachedType(key)` before calling the new `zsetStorageHintAt`. Without this, a pre-script storage probe leaked stale pre-mutation state: `SET k x; ZSCORE k m` returned the old score instead of WRONGTYPE; `DEL k; ZSCORE k m` returned the old score instead of nil. New `TestLua_ZSCORE_SetThenZScoreReturnsWrongType` / `TestLua_ZSCORE_DelThenZScoreReturnsNil` cover both. Mirrors the legacy `keyType()` fallback order: cachedType first, storage second.

**Addressed review comments**
- **Medium — redundant I/O in `rawZSetPhysTypeAt` fallback**: added `rawKeyTypeAtSkipZSet` and `detectWideColumnTypeSkipZSet` helpers so the non-ZSet fallback no longer re-scans the three ZSet-side prefixes (`ZSetMemberScanPrefix`, `ZSetMetaKey`, `ZSetMetaDeltaScanPrefix`) that `rawZSetPhysTypeAt` already probed.
- **Medium — TTL probe on hot collection path**: `zsetStorageHintAt` now passes `isNonStringCollectionType(physType)` to `hasExpired`, skipping the `!redis|str|` probe when the physType is a known non-string collection. Saves a `GetAt` on the hot path.

Ready for re-review.

PR #570, opened 2026-04-21
perf(redis-lua): stream ZRANGEBYSCORE from the score index

bootjp:
## Summary
Phase B of the ZSet fast-path plan. Production pprof after PRs #565 / #567 / #568 showed `cmdZRangeByScore` + `cmdZRangeByScoreAsc` holding **40 active goroutines** — the top residual Lua-path signal — driven by BullMQ delay-queue polls. The legacy path loaded ALL zset members via `zsetState` → `ensureZSetLoaded` before filtering by score bounds: O(N) `GetAt`s per `ZRANGEBYSCORE`, so an N=1000 delay queue cost several seconds per call.

Replace load-then-filter with a **bounded scan over the pre-sorted score index** (`!zs|scr|<user>|<sortable-score>|<member>`).

### Changes

**`adapter/redis_compat_commands.go`** — new `zsetRangeByScoreFast` helper plus small utilities:
- Legacy-blob zsets, string-priority corruption, and empty-result-but-zset-missing cases return `hit=false` so the slow path's WRONGTYPE / disambiguation semantics remain intact.
- `zsetRangeEmptyFastResult` checks `ZSetMetaKey` to separate "empty range on a live zset" from "no zset here".
- Split into `zsetFastPathEligible` / `zsetFastScanLimit` / `zsetScoreScan` / `decodeZSetScoreRange` / `finalizeZSetFastRange` to stay under the cyclomatic-complexity cap.

**`adapter/redis_lua_context.go`** — `cmdZRangeByScore` gated on `luaZSetAlreadyLoaded` + `cachedType` (mirrors PR #567 / #568 safety guards):
- In-script ZADD / ZREM / DEL / SET / type-change still go through the legacy full-load path (`cmdZRangeByScoreSlow`).
- `zsetScoreScanBounds` maps `zScoreBound` (-Inf / value-inc / value-exc / +Inf) to a byte range; exclusive edges are handled post-scan by `scoreInRange`.

### Per-call effect on BullMQ-style delay-queue polls

| Scenario | Before | After (fast-path hit) |
|---|---|---|
| `ZRANGEBYSCORE key 0 now LIMIT 0 10` over N=1000 zset | ~1000 member `GetAt`s | 1 legacy-blob `ExistsAt` + 1 bounded `ScanAt` (≤10 rows) + 1 priority-guard `ExistsAt` + 1 TTL probe |
| Empty range | ~1000 `GetAt`s | 1 legacy-blob + 1 `ZSetMetaKey` + 1 TTL |
| In-script `ZADD` then `ZRANGEBYSCORE` | unchanged | unchanged (slow path via `luaZSetAlreadyLoaded`) |

## Test plan
- [x] `go test -race -short ./adapter/...` passes
- [x] 10 new cases in `adapter/redis_lua_collection_fastpath_test.go`:
  - fast-path hit (ascending)
  - WITHSCORES
  - LIMIT + offset
  - exclusive bounds `(min (max`
  - empty range on live zset
  - missing key
  - WRONGTYPE (key holds a string)
  - in-script `ZADD` visible to subsequent `ZRANGEBYSCORE`
  - `ZREVRANGEBYSCORE` (reverse iteration)
  - TTL-expired zset
- [ ] Production: deploy and verify `cmdZRangeByScore` active goroutines drop from ~20+20 back toward the keyTypeAt-style residual. Expect p50 redis.call latency inside Lua to fall from the current 3s toward ms-scale as delay-queue polls stop pulling the whole zset.

## Follow-ups (out of scope)
- `ZRANGE` (by rank) streaming — needs the same pattern with positional iteration.
- `ZPOPMIN` — writes; needs the streaming scan plus the delete-member transaction plumbing.
- `HGETALL` / `HMGET` — separate wide-column prefix scans.

## Summary by CodeRabbit

* **New Features**
  * Optimized sorted-set range-by-score fast path for faster, paginated queries with correct inclusive/exclusive bound handling.

* **Bug Fixes**
  * Safer fallback to full-path when results may be incomplete; fixed empty-range, truncation, type-mismatch and TTL/expiration behaviors; negative LIMIT offsets now rejected and negative limits treated as “no limit.”

* **Tests**
  * Added extensive Lua-scripted tests covering range reads, WITHSCORES, LIMIT semantics, exclusive bounds, in-script mutations, TTL, and regressions.

### DIFF
--- a/adapter/redis_compat_commands.go
+++ b/adapter/redis_compat_commands.go
@@ -1,6 +1,7 @@
 package adapter
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


bootjp:
gemini のコメント対応 push 済みです (`5e124b97`).

- `luaScriptContext` に `ctx` フィールドを追加、`newLuaScriptContext` で受け取った ctx を保存するようにしました
- `scriptCtx()` ヘルパー経由で新規 fast-path (`zrangeByScoreFastPath`) は `c.ctx` を使うようにしました (BullMQ hot path で Eval のデッドライン / キャンセルが store 層まで届きます)
- 既存の 15 箇所以上の `context.Background()` 呼び出しは gemini の提案通り別 PR で一括移行します（このスタックの scope を保つため）。新規コードには正しいパターンが入ったので、以降の cmd* ハンドラは `scriptCtx()` 経由で ctx を透過できます

bootjp:
CodeRabbit (Major x2) と gemini (High x2) の指摘を対応しました (`a38b8172`)。

**修正内容:**

1. **排他最小バウンド (CodeRabbit Major)** — `ZRANGEBYSCORE key (2 +inf LIMIT 0 1` で score=2 が多数あると、scan が `offset+limit` 予算を score=2 の行で消費して score=3 以上のマッチを取り逃していました。`zsetScoreScanBounds` で排他最小を `PrefixScanEnd(scorePrefix)` に移動、scan 時点で score=min 全行をスキップ。

2. **スキャン打ち切り検出 (gemini High)** — `finalizeZSetFastRange` が `scanLimit` を受け取り、raw スキャンが上限に達し、かつ request 未満足なら `hit=false` を返して slow path へフォールバック。サイレント切り詰め防止。`zsetFastPathTruncated` ヘルパーに集約。

3. **空結果 tail が delta-only zset を見落とし (CodeRabbit Major)** — `ExistsAt(ZSetMetaKey)` ではなく `resolveZSetMeta` を使用（base meta + delta 両方を認識）。また priority guard も empty tail で走らせるようにして、corrupted redisStrKey + zset meta で empty ではなく WRONGTYPE が返るように。

4. **offset/limit 防御チェック (gemini High)** — `cmdZRangeByScore` で negative offset / 非 sentinel negative limit を `ERR syntax error` で早期拒否。hostile script の OOM / panic 防止。

regression test 4 本追加 (exclusive-min with many at bound / truncation / fresh zset empty range / negative offset)。既存テストも全通過。

bootjp:
Addressed the latest review round (97bc74f4):

- **gemini HIGH / CodeRabbit Minor — `zsetFastScanLimit` signed-integer overflow**: compare `limit > maxWideScanLimit - offset` before adding, so `math.MaxInt + math.MaxInt` no longer wraps to a negative `scanLimit` and misroutes live zsets into the empty-result tail.
- **gemini Medium — `decodeZSetScoreRange` over-allocation**: extracted `zsetDecodeAllocSize(kvLen, offset, limit)` that returns `(kvLen - offset)` capped by `limit`, saving up to `maxWideScanLimit` wasted slice capacity on small windows at large offsets.
- **gemini Medium — negative LIMIT parity with upstream Redis**: `cmdZRangeByScore` now coerces `options.limit < 0` → `-1` instead of returning `ERR syntax error`, so `ZRANGEBYSCORE k -inf +inf LIMIT 0 -1` in Lua behaves exactly like the real server.

New regression test `TestLua_ZRANGEBYSCORE_NegativeLimitReturnsAll` pins the negative-limit semantics.

/gemini review

bootjp:
Addressed the latest review round (d8838182):

- **gemini Medium — `zsetFastScanLimit` limit==0 short-circuit**: return 0 unconditionally when `limit == 0`; the caller's `scanLimit <= 0` branch already routes to `zsetRangeEmptyFastResult`, which still runs `resolveZSetMeta` for correct WRONGTYPE / existence disambiguation. No wasted score-index scan for `LIMIT 0 0`.
- **gemini Medium — `decodeZSetScoreRange` loop ordering**: hoisted the `len(entries) >= limit` check above the offset-skip branch, so a small `limit` combined with a large `offset` exits the decode loop immediately instead of burning `offset` iterations on the skip path.

Correctness unchanged in both cases.

/gemini review

bootjp:
Addressed the latest review finding (bb6ba4f1):

- **gemini Medium — `zsetScoreScanBounds` silently drops `+/-Inf` exact matches**: `-Inf` as `maxBound` and `+Inf` as `minBound` were both mapping to the empty-range endpoints (`full` / `fullEnd`), so `ZRANGEBYSCORE key -inf -inf` / `+inf +inf` returned empty instead of the legitimate members Redis would return (`ZADD key +inf m` is valid).
  - `maxBound = -Inf` → `PrefixScanEnd(ZSetScoreRangeScanPrefix(key, -Inf))`
  - `minBound = +Inf` → `ZSetScoreRangeScanPrefix(key, +Inf)`

Added regression tests `TestLua_ZRANGEBYSCORE_NegInfExactMatch` and `TestLua_ZRANGEBYSCORE_PosInfExactMatch` pinning the corrected behaviour.

/gemini review

bootjp:
Addressed the latest review finding (9ef65d0f):

- **gemini Medium — large-offset wasted scan**: when `offset >= maxWideScanLimit`, the fast path was scanning `maxWideScanLimit` rows and skipping all of them, then hitting the truncation guard and falling back anyway. Added an explicit `hit=false` short-circuit at the top of `zsetRangeByScoreFast` so the slow path answers directly without the redundant score-index scan.

Added `TestLua_ZRANGEBYSCORE_LargeOffsetShortCircuit` pinning the behaviour (`LIMIT 200000 10` on a 3-member zset returns empty cleanly).

/gemini review

bootjp:
Addressed the latest review findings (b718dc93):

- **gemini Medium — WITHSCORES reply under-allocation**: `zsetRangeReply` sized `out` to `len(entries)` but each entry emits 2 replies when `withScores=true`. Pre-allocate `2 * len(entries)` in that case. Bounded by `maxWideScanLimit` at the fast-path layer so allocation remains capped.
- **gemini Medium — dead expired branch**: `zsetRangeEmptyFastResult` had `if expired { return nil,true,nil }` followed by `return nil,true,nil` — identical returns. Collapsed to a single tail return; kept the `hasExpired` call for its error-surfacing side effect.

/gemini review

bootjp:
CodeRabbit review acknowledged — thanks for the LGTM.

The test-helper consolidation nit is a fair style point but I'm deferring it intentionally: each test body is narrow and the preamble being visible inline keeps the regression-pinning story readable next to the assertion. Happy to revisit in a follow-up refactor PR if the file grows further.

No other actionable findings. Marking the review cycle complete on my side.

PR #571, opened 2026-04-21
obs(redis-lua): per-cmd fast-path outcome counter

bootjp:
## Summary
- Add `elastickv_lua_cmd_fastpath_total{cmd,outcome}` counter to measure how often Lua-side fast paths actually take the fast path vs fall back (hit / skip_loaded / skip_cached_type / fallback).
- Wire `LuaFastPathObserver` through Registry → RedisServer option → `luaScriptContext.observeFastPathOutcome`.
- Instrument `cmdZRangeByScore` as the first caller.

## Motivation
Post-deploy CPU profile on a BullMQ-style workload shows `cmdZRangeByScoreSlow` consuming ~14x the CPU of `zsetRangeByScoreFast` (860ms vs 60ms in a 30 s profile), despite the fast path being eligible on paper. Without per-outcome instrumentation it is not possible to tell whether scripts are hitting `luaZSetAlreadyLoaded` (a prior ZCARD/ZCOUNT on the same key), `cachedType` (SET/DEL earlier in the script), or the server-side `hit=false` fallback (legacy-blob zset, truncation, large-offset short-circuit).

Shipping this counter first so we can decide whether the follow-up should loosen the `luaZSetAlreadyLoaded` guard, or target a different cause.

## Test plan
- [x] `go test ./adapter/... ./monitoring/...` green
- [x] New `TestLuaFastPathObserver*` in monitoring/hotpath_test.go
- [x] Zero-value observer safe (already used by tests that do not wire a registry)
- [ ] Deploy to prod, watch `rate(elastickv_lua_cmd_fastpath_total{cmd=zrangebyscore}[5m])` by outcome for ~5 minutes

## Summary by CodeRabbit

* **New Features**
  * Added Prometheus metrics to monitor Lua script fast-path execution performance, tracking per-command outcomes including successful hits, skipped operations from cached data, and fallback scenarios.

* **Tests**
  * Added tests validating Lua fast-path metrics accuracy and zero-value safety.

### DIFF
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -266,6 +266,11 @@ type RedisServer struct {
 	requestObserver     monitoring.RedisRequestObserver
 	luaObserver         monitoring.LuaScriptObserver
 	luaFastPathObserver monitoring.LuaFastPathObserver
+	// luaFastPathZRange is the pre-resolved counter bundle for the
+	// ZRANGEBYSCORE / ZREVRANGEBYSCORE Lua fast path. Resolved once in
+	// WithLuaFastPathObserver so the hot path does not pay for
+	// CounterVec.WithLabelValues on every redis.call().
+	luaFastPathZRange monitoring.LuaFastPathCmd
 	// baseCtx is the parent context for per-request handlers.
 	// NewRedisServer creates a cancelable context here; Stop() cancels
 	// it so in-flight handlers abort promptly instead of running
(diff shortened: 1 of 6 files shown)


bootjp:
Addressed gemini's review (1cd79de4):

- **Medium — `WithLabelValues` mutex on hot path**: the observer API now pre-resolves counter handles at server construction via `LuaFastPathObserver.ForCommand(cmd)`, which returns a `LuaFastPathCmd` bundle with four resolved `prometheus.Counter` handles. `cmdZRangeByScore` calls `.ObserveHit()` / `.ObserveSkipAlreadyLoaded()` / `.ObserveSkipCachedType()` / `.ObserveFallback()` directly — each is a single atomic `Inc`, no map lookups on the redis.call() path.

Zero-value `LuaFastPathCmd` remains safe (each `ObserveXxx` nil-checks the handle), so tests that do not wire a registry continue to work without change.

/gemini review

bootjp:
Addressed gemini's review round 2 (e75442ec):

- Fixed the doc comment on `LuaFastPathObserver` that referenced a non-existent `ObserveLuaFastPath-on-handle` method. Now describes the actual `Observe*` call shape on `LuaFastPathCmd` and notes the counters are non-blocking atomic `Inc` (client_golang's default Counter uses sync/atomic internally).
- Renamed the field `skipAlreadyLoad` / constant `LuaFastPathOutcomeSkipAlreadyLoad` / method `ObserveSkipAlreadyLoaded` → `skipLoaded` / `LuaFastPathOutcomeSkipLoaded` / `ObserveSkipLoaded` so every layer agrees with the metric label value `skip_loaded`.

/gemini review

PR #572, opened 2026-04-21
obs(redis-lua): subdivide ZRANGEBYSCORE fast-path fallback by reason

bootjp:
## Summary

Follow-up to PR #571. The counter it added showed (on n1 / leader, 15 min of BullMQ traffic):

```
fallback         1463  (96.4%)
hit                54  (3.6%)
skip_loaded         0
skip_cached_type    0
```

The Lua-layer guards are not the issue; the fast path is being rejected server-side. The current single `fallback` bucket is too coarse to tell *why*: legacy-blob eligibility, large-offset short-circuit, truncation, empty-key missing-zset, or wrong-type guard.

This PR subdivides the `fallback` outcome label so the real cause is visible in 5 minutes of prod scrapes.

## Changes

- New outcome labels (replacing single `fallback`):
  - `fallback_ineligible` — `zsetFastPathEligible` rejected (legacy-blob zset)
  - `fallback_large_offset` — `offset >= maxWideScanLimit` short-circuit
  - `fallback_truncated` — scan hit `scanLimit` with request unsatisfied
  - `fallback_missing_key` — `zsetRangeEmptyFastResult` saw `zsetExists=false`
  - `fallback_wrong_type` — `hasHigherPriorityStringEncoding=true`
  - `fallback_other` — error / unclassified
- `zsetRangeByScoreFast` grows a `reason string` return (empty when `hit=true`); all `hit=false` sites populate a specific reason.
- `zsetRangeEmptyFastResult` now returns `(hit, reason, err)`; the old `[]redisZSetEntry` return was always `nil`.
- `LuaFastPathCmd` pre-resolves 6 new counters (same hot-path shape as #571: one atomic `Inc` per redis.call()).
- `ObserveFallback(reason)` dispatches via a small switch; unknown reasons land on `fallback_other`.

## Motivation

Pre-deploy hypothesis: most fallbacks are `fallback_missing_key` (BullMQ polling empty delayed queues). Ship this first, measure in prod, then write the actual fix accordingly.

## Test plan

- [x] `go test -race -short ./adapter/...` (53s) green
- [x] `go test -race ./monitoring/...` green
- [x] `TestLuaFastPathObserverCountsByCmdAndOutcome` updated for new labels + unknown-reason dispatch
- [ ] Deploy to prod; watch `sum by (outcome) (rate(elastickv_lua_cmd_fastpath_total{cmd="zrangebyscore"}[5m]))`

## Summary by CodeRabbit

* **New Features**
  * Improved query performance monitoring with categorized fallback reasons for sorted set operations, providing enhanced visibility into query optimization effectiveness and performance diagnostics.

### DIFF
--- a/adapter/redis_compat_commands.go
+++ b/adapter/redis_compat_commands.go
@@ -1808,9 +1808,9 @@ func (r *RedisServer) zsetRangeByScoreFast(
 	offset, limit int,
 	scoreFilter func(float64) bool,
 	readTS uint64,
-) ([]redisZSetEntry, bool, string, error) {
+) ([]redisZSetEntry, bool, monitoring.LuaFastPathFallbackReason, error) {
 	if eligible, err := r.zsetFastPathEligible(ctx, key, readTS); err != nil || !eligible {
-		return nil, false, monitoring.LuaFastPathOutcomeFallbackIneligible, err
+		return nil, false, monitoring.LuaFastPathFallbackIneligible, err
 	}
 	// Large-offset short-circuit: once offset >= maxWideScanLimit,
 	// the fast path can only scan maxWideScanLimit rows then skip all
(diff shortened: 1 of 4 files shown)


bootjp:
Addressed gemini's review (aa0c0dec):

- **Medium — bare `string` for fallback reason**: introduced `monitoring.LuaFastPathFallbackReason` (named string type) and propagated it through `zsetRangeByScoreFast` → `finalizeZSetFastRange` → `zsetRangeEmptyFastResult` → `zrangeByScoreFastPath` → `ObserveFallback`. Typos in call sites now fail at compile time instead of silently routing to `fallback_other`.

The corresponding constants are renamed `LuaFastPathOutcomeFallback*` → `LuaFastPathFallback*` (on the new type); `LuaFastPathOutcomeHit/SkipLoaded/SkipCachedType` stay as bare-string constants since they're non-fallback outcomes. Label values on the wire are unchanged — `string(reason)` preserves dashboards.

/gemini review

PR #576, opened 2026-04-21
docs(redis-lua): document Lua commit batching architecture

bootjp:
## Summary

Investigated whether `redis.call(...)` writes inside a Lua script each submit their own Raft proposal (the suspected cause of the 5s -> 10s Raft Commit p99 regression after a recent deploy, with BullMQ's `moveJobFromPrioritizedToActive` issuing ~5-7 `redis.call`s per script).

**Finding: the Lua adapter already batches all in-script writes into a single Raft proposal.** Every `cmd*` handler in `luaScriptContext` mutates only in-memory state maps (`strings`, `lists`, `hashes`, `sets`, `zsets`, `streams`, `ttls`); no handler crosses into Raft. After `state.PCall` returns, `runLuaScript` calls `scriptCtx.commit()` exactly once, which allocates one `commitTS`, walks the sorted `touched` set, assembles every key's Put/Del elements into one `kv.OperationGroup{IsTxn: true, ...}`, and issues a single `coordinator.Dispatch`. A grep over `adapter/redis_lua_context.go` confirms exactly one `Dispatch` call site, in `commit()`.

Therefore there is nothing to ship on the "coalesce in-script mutations" front. This PR adds `docs/lua-commit-batching.md` capturing the architecture so the next investigator does not redo this work, and so any future regression that re-introduces per-call proposals can be caught in review.

The p99 regression must be caused by something else (proposal size / raft tick config / snapshot churn / storage latency) - not by per-`redis.call` Raft fan-out.

## Test plan

- [ ] No code changes; existing `go test -race -short ./adapter/...` still applies and is unaffected.
- [ ] Doc-only: verify `docs/lua-commit-batching.md` renders on GitHub.
- [ ] Spot-check references: `runLuaScript` / `luaScriptContext.exec` / `luaScriptContext.commit` / `coordinator.Dispatch` call sites match the file paths and function names cited in the doc.

## Summary by CodeRabbit

* **Documentation**
  * Added documentation describing Lua script execution and commit batching, including script processing flow, commit semantics, retry handling, and accumulated write operation behavior.

### DIFF
--- a/docs/lua-commit-batching.md
+++ b/docs/lua-commit-batching.md
@@ -1,0 +1,115 @@
+# Lua Commit Batching Architecture
+
+## Summary
+
+**One Redis EVAL / EVALSHA = one Raft proposal** for all writes performed by
+`redis.call(...)` inside the script.
+
+This document captures the current behaviour of the Redis-Lua adapter's commit
+path, verified in April 2026 while investigating whether BullMQ-style scripts
+(e.g. `moveJobFromPrioritizedToActive`, which performs 5-7 `redis.call`s
+including `ZREM` + `LPUSH` + `HSET` + `HINCRBY`) were inflating Raft commit
+traffic proportionally to the number of in-script mutations. They were not:
+the batching is already in place.
+
+The investigation was triggered by a production observation that Raft Commit
+p99 increased from ~5s to ~10s after a deploy. The batching layer described
+here is *not* the regression; look elsewhere (e.g. proposal size, raft tick
+tuning, snapshot churn, storage latency).
+
+## Call flow
+
+For `EVAL`/`EVALSHA` on the adapter:
+
+1. `(*RedisServer).eval` / `.evalsha` in `adapter/redis_lua.go` dispatch to
+   `runLuaScript`.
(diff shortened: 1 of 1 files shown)


bootjp:
Thanks for the review! Addressed all three comments in 340d38e0:

- Line 43: replaced hardcoded line number with a function-name reference ("in the `commit()` method of `adapter/redis_lua_context.go`").
- Line 48: fixed the invalid Go struct-literal shorthand — now uses the actual `StartTS: startTS, CommitTS: commitTS, Elems: elems` form so the snippet compiles.
- Line 100: replaced the hardcoded `~L3084` marker with "the single coordinator.Dispatch site".

/gemini review

bootjp:
Re-checked gemini's latest review (2026-04-21T20:51:25Z): it reports "I have no feedback to provide" with no new inline comments. The three inline suggestions from the earlier review (2026-04-21T19:12:03Z) — avoiding hardcoded line numbers in two spots and fixing the OperationGroup struct-literal shorthand — were all addressed in 340d38e0. No further doc changes needed.

/gemini review

PR #578, opened 2026-04-21
obs(pebble): expose LSM internals as prometheus metrics

bootjp:
## Summary

- Adds an `elastickv_pebble_*` Prometheus metrics family that periodically polls each raft group's `*pebble.DB.Metrics()` snapshot so operators can graph and alert on write-stall precursors.
- Mirrors the existing `DispatchCollector` shape: a `PebbleMetricsSource` interface (satisfied by the `pebbleStore` via a new `Metrics()` accessor guarded by `dbMu`), a `PebbleCollector` that polls on the same 5 s cadence from `main.go`.
- Covered by unit tests using a `fakePebbleSource` returning canned `*pebble.Metrics`; asserts gauge-overwrite semantics and reset-safe counter deltas.

## Exposed metrics (per `group` label)

Gauges (overwritten each tick):
- `elastickv_pebble_l0_sublevels` - canonical write-stall precursor
- `elastickv_pebble_l0_num_files`
- `elastickv_pebble_compact_estimated_debt_bytes`
- `elastickv_pebble_compact_in_progress`
- `elastickv_pebble_memtable_count` / `_size_bytes` / `_zombie_count`
- `elastickv_pebble_block_cache_size_bytes`

Counters (positive-delta only, reset-safe across Restore/swap):
- `elastickv_pebble_compact_count_total`
- `elastickv_pebble_block_cache_hits_total` / `_misses_total`

## Deviations from the task description

- **No `WriteStallCount` / `WriteStallDuration`**: pebble v2.1.4 does not expose those on the `Metrics` struct (they are only delivered via `EventListener.WriteStallBegin` / `End` callbacks). Wiring an event listener is doable but it would require modifying `defaultPebbleOptions()` to install one - that is a larger surface than this observability pass. L0 sublevels and compaction debt are the more actionable write-stall proxies and are included. Can follow up with an `EventListener`-based stall counter in a separate PR if desired.
- **Counter semantics**: the first observation emits the absolute cumulative value (e.g. 100 hits -> `_total` = 100); subsequent observations emit only the positive delta. A source reset (`Restore` reopens the DB with fresh counters) causes a smaller observed value - that delta is silently dropped so the Prometheus counter stays monotonic, mirroring `DispatchCollector` behaviour.

## Test plan

- [x] `go build ./...`
- [x] `go test -race -count=1 -short ./monitoring/... ./store/...`
- [x] `make lint` (0 issues)

### DIFF
--- a/main.go
+++ b/main.go
@@ -187,6 +187,9 @@ func run() error {
 	if collector := metricsRegistry.DispatchCollector(); collector != nil {
 		collector.Start(runCtx, dispatchMonitorSources(runtimes), raftMetricsObserveInterval)
 	}
+	if collector := metricsRegistry.PebbleCollector(); collector != nil {
+		collector.Start(runCtx, pebbleMonitorSources(runtimes), raftMetricsObserveInterval)
+	}
 	compactor := kv.NewFSMCompactor(
 		fsmCompactionRuntimes(runtimes),
 		kv.WithFSMCompactorActiveTimestampTracker(readTracker),
@@ -442,6 +445,30 @@ func raftMonitorRuntimes(runtimes []*raftGroupRuntime) []monitoring.RaftRuntime 
 			GroupID:      runtime.spec.id,
 			StatusReader: runtime.engine,
 			ConfigReader: runtime.engine,
+		})
+	}
+	return out
+}
+
+// pebbleMonitorSources extracts the MVCC stores that expose
+// *pebble.DB.Metrics() so monitoring can poll LSM internals (L0
+// sublevels, compaction debt, memtable, block cache) for the
+// elastickv_pebble_* metrics family. Stores that do not satisfy the
+// interface (non-Pebble backends, if any are added later) are skipped
+// silently.
+func pebbleMonitorSources(runtimes []*raftGroupRuntime) []monitoring.PebbleSource {
+	out := make([]monitoring.PebbleSource, 0, len(runtimes))
+	for _, runtime := range runtimes {
+		if runtime == nil || runtime.store == nil {
+			continue
+		}
+		src, ok := runtime.store.(monitoring.PebbleMetricsSource)
+		if !ok {
+			continue
+		}
+		out = append(out, monitoring.PebbleSource{
+			GroupID: runtime.spec.id,
+			Source:  src,
 		})
 	}
 	return out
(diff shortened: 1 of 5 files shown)


bootjp:
Addressed the typo in `monitoring/pebble.go`: renamed `blockCacheMisse` -> `blockCacheMisses` on lines 182, 263, 271, 272 (commit b8a8b67e). Tests pass (`go test -race -count=1 -short ./monitoring/... ./store/...`) and `golangci-lint run ./monitoring/...` is clean.

/gemini review

bootjp:
Addressed gemini's latest medium-priority review (2026-04-21T20:53:10Z) in 2af6f78f.

**Change**: precompute `PebbleSource.GroupIDStr` once at construction (in `pebbleMonitorSources`) instead of calling `strconv.FormatUint` inside `observeOnce` on every 5s tick per group. Over the process lifetime this removes one small-string allocation per group per tick.

- `monitoring/pebble.go`: add `GroupIDStr string` field to `PebbleSource`; use `src.GroupIDStr` as the `group` label in `observeOnce`; drop the now-unused `strconv` import.
- `main.go`: populate `GroupIDStr: strconv.FormatUint(runtime.spec.id, 10)` in `pebbleMonitorSources`.
- `monitoring/pebble_test.go`: populate `GroupIDStr` in the three `PebbleSource` literals used by the collector tests.

**Verification**: `go build ./...` clean; `go test -race -count=1 -short ./monitoring/... ./store/...` passes (monitoring 2.085s, store 7.011s); `make lint` 0 issues.

/gemini review

bootjp:
/gemini review

Addressed Copilot's 3 review items in be3c19b4:

1. `store/lsm_store.go:1803` — Updated `Metrics()` docstring to accurately describe nil semantics: nil only before the first Open or after a failed Open; Restore callers block on `dbMu` (held exclusively by Restore) rather than observing nil, and `Close()` does not clear `s.db`. Behavior unchanged (out of scope for this PR).
2. `monitoring/pebble.go:19` — Removed "Count" from the header comment; only `BlockCache.Size` is exported (Size is more actionable than Count).
3. `monitoring/pebble.go:53` — Renamed `blockCacheMissTotal` to `blockCacheMissesTotal` for consistency with `blockCacheHitsTotal`, the adjacent snapshot field `blockCacheMisses`, and the exported metric name `..._misses_total`. All usages propagated.

Verification: `go build ./...` clean, `go test -race -count=1 -short ./monitoring/... ./store/...` passes, `make lint` 0 issues.

PR #579, opened 2026-04-21
perf(redis-lua): cache keyTypeAt=None results in Eval scope

bootjp:
## Summary
- Extend `luaScriptContext.cachedType` with a per-Eval `negativeType` map populated when `server.keyTypeAt()` returns `redisTypeNone`. MVCC at the pinned `startTS` guarantees the absence result is stable for the script's duration.
- Short-circuit `cmdZScore` / `cmdZRangeByScore` / `cmdHGet` / `cmdHExists` / `cmdSIsMember` to their empty-reply shape on a cached-None type instead of punting to the slow path, eliminating the remaining `rawKeyTypeAt` probes that `zsetState` / `hashState` / `setState` would otherwise issue for a missing key.
- No explicit invalidation needed: in-script writes populate `cachedLoadedTypes[...].exists=true`, which `cachedType()` consults first and thus shadows the negative entry automatically.

Motivation: post-#573/575 pprof still showed `rawKeyTypeAt` net write 14% CPU cum on the BullMQ "poll a missing delayed zset" loop because the negative result was not cached; every subsequent `redis.call` re-paid the ~8-seek type probe.

## Test plan
- [x] `go test -race -short ./adapter/...` passes (63.58s)
- [x] `TestLua_ZSCORE_DelThenZScoreReturnsNil` still green (DEL-then-ZSCORE path unchanged)
- [x] `TestLua_HGET_SetThenHGetReturnsWrongType` / `TestLua_ZSCORE_SetThenZScoreReturnsWrongType` still green (SET-over-hash/zset invariant preserved)
- [x] New `TestLuaNegativeTypeCache_SingleProbePerKey` pins "repeated `keyType()` on missing key = exactly 1 server probe" via in-process `keyTypeProbeCount` counter
- [x] New `TestLua_HGET_MissingThenSetReturnsWrongType` / `TestLua_ZSCORE_MissingThenZAddReturnsScore` pin the correctness invariants called out in the task spec
- [x] New `TestLua_ZSCORE_MissingThenMissingStillReturnsNil` / `TestLua_ZRANGEBYSCORE_MissingReturnsEmptyArray` cover the hot-case wire reply shape

### DIFF
--- a/adapter/redis_lua_context.go
+++ b/adapter/redis_lua_context.go
@@ -43,6 +43,11 @@ type luaScriptContext struct {
 	// populate the concrete-type state (c.strings/c.zsets/... with
 	// exists=true), which cachedType() returns first, so the negative
 	// cache is shadowed without needing explicit invalidation.
+	//
+	// Bounded by maxNegativeTypeCacheEntries to prevent a malicious or
+	// buggy script from probing unbounded unique non-existent keys and
+	// blowing up memory. Once full, further misses fall back to the
+	// server-side probe (still correct, just not cached).
 	negativeType map[string]bool
 
 	// keyTypeProbeCount counts how many times the server-side keyTypeAt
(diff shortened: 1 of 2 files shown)


bootjp:
Thanks for the review. Both items addressed in 441cd20a:

1. **Bounded `negativeType` map** — added `const maxNegativeTypeCacheEntries = 10_000` (matches `kv.maxReadKeys` magnitude). The insert site in `keyType()` now guards with `len(c.negativeType) < maxNegativeTypeCacheEntries` before caching; once full, misses fall through to the server probe without caching. Correctness invariant "if cached, it's correct" is preserved — dropping caching just means re-probing, never stale reads. A new regression test `TestLuaNegativeTypeCache_BoundedSize` probes cap+50 unique missing keys and asserts the map stays at the cap, the first `cap` keys remain cached (no re-probe), and post-cap keys fall back to the server probe every time.

2. **Context propagation** — swapped `context.Background()` → `c.scriptCtx()` at the `keyTypeAt` call, so the Eval's deadline and cancellation now flow into the storage probe.

`go test -race -count=1 -short ./adapter/...` and `make lint` both green.

/gemini review

PR #581, opened 2026-04-22
fix(kv): make txn abort idempotent when rollback marker exists

bootjp:
## Summary

Fix a 2PC abort-retry race that surfaces in prod as `"secondary write failed" ... "write conflict"` log spam on `!txn|rb|` rollback-marker keys, and close the reciprocal commit-after-abort gap flagged in review.

```
"msg":"secondary write failed","cmd":"EVALSHA",
"err":"<string>:118: key: !txn|rb|!redis|ttl|misskey.bootjp.me:queue:...:stalled-check\^A\ufffd-\^C9\ufffd\^@\^@: write conflict"
```

## Root cause

`handleAbortRequest` is not idempotent. Once a `(primaryKey, startTS)` pair has been aborted, the rollback marker `!txn|rb|<primaryKey>+<startTS>` sits at `commitTS = abortTS`. A second abort of the same pair — from a concurrent lock resolver, a retry, or a dualwrite async replay — rebuilds:

- `Delete` on the already-tombstoned lock / intent keys
- `Put` on the already-present rollback marker

Every mutation has `latestCommitTS = abortTS > startTS`, so MVCC `checkConflicts` rejects all three as `ErrWriteConflict`. `TestFSMAbort_SecondAbortSameTimestampConflicts` was literally pinning this bug's current behaviour.

## Fix (per-key idempotency, NOT marker short-circuit)

The ABORT request is **not** short-circuited on rollback-marker presence. The marker only proves that *some* prior abort for `(primaryKey, startTS)` ran; it does not prove that cleanup ran for the specific keys in *this* request. In particular `ShardStore.tryAbortExpiredPrimary` issues an ABORT whose mutation list contains only the primary key, so a later lock-resolver abort for a secondary key (same `primaryKey`, same `startTS`) would see the marker already present and must still clean up that secondary's lock / intent.

Idempotency is enforced **per key**:

- `shouldClearAbortKey`: if the lock is already gone, skip cleanup for that key (no stale `Delete` on a tombstoned lock/intent).
- `appendRollbackRecord`: if the rollback marker is already present for `(primaryKey, startTS)`, skip the `Put` (no byte-identical re-Put that MVCC would reject as a write conflict).

Both cost one extra `GetAt` on the hot abort path; the common case (fresh abort) is unaffected.

## Symmetric commit / rollback invariant

Desired invariant: for any `(primaryKey, startTS)`, at most one of `{rollback marker, commit record}` is ever present. The invariant is enforced by symmetric guards rather than assumed:

- `appendRollbackRecord`: verifies the commit record is absent on **both** the first-time and idempotent paths (not just the first-time path), returning `ErrTxnAlreadyCommitted` if a commit record already exists.
- `commitApplyStartTS`: verifies the rollback marker is absent before writing a commit record, returning a new `ErrTxnAlreadyAborted` if one exists. This catches out-of-order apply (COMMIT after ABORT), buggy clients, replay races — the case addressed directly by the review.

## Test plan

- [x] `go test -race -count=1 ./kv/...` green
- [x] `make lint` clean
- [x] `TestFSMAbort_SecondAbortSameKeysIsIdempotent` — same-abortTS and later-abortTS retry (HLC-monotonic lock-resolver race path)
- [x] `TestFSMAbort_SecondAbortDifferentKeysCleansRemainder` — primary-only first abort followed by secondary-key abort must still clean the secondary (regression for a would-be over-broad marker short-circuit)
- [x] `TestFSMAbort_RejectsAlreadyCommittedTxn` — abort-after-commit rejected with `ErrTxnAlreadyCommitted`
- [x] `TestFSMAbort_CommitAfterAbortIsRejected` — commit-after-abort rejected with `ErrTxnAlreadyAborted` (new symmetric guard)
- [ ] Deploy; `"secondary write failed"` log rate on `!txn|rb|` should drop to zero

Relates to the BullMQ `stalled-check` traffic class across the `relationship`, `deliver`, `objectStorage`, and webhook queues.

## Summary by CodeRabbit

* **Bug Fixes**
  * Commits for transactions that were previously aborted are now rejected.
  * Abort operations are idempotent — retries (even at later timestamps) succeed without error.
  * Distributed lock cleanup across primary/secondary keys is more reliable; no orphaned locks remain.
  * No unnecessary lock deletions occur when locks are already absent.

* **Tests**
  * Added coverage validating abort idempotency, lock-resolver races, and commit-after-abort rejection.

### DIFF
--- a/kv/fsm.go
+++ b/kv/fsm.go
@@ -457,12 +457,26 @@ func (f *kvFSM) handleCommitRequest(ctx context.Context, r *pb.Request) error {
 // commitApplyStartTS resolves the startTS to use for MVCC conflict detection
 // during a COMMIT. If a commit record already exists for the primary key it
 // returns commitTS (making the apply idempotent); otherwise it returns startTS.
+//
+// It is also the symmetric guard to appendRollbackRecord: if a rollback marker
+// already exists for (primaryKey, startTS) the commit is rejected with
+// ErrTxnAlreadyAborted. Together with the commit-record check in
+// appendRollbackRecord, this enforces the invariant that at most one of
+// {rollback marker, commit record} is present for any (primaryKey, startTS).
 func (f *kvFSM) commitApplyStartTS(ctx context.Context, primaryKey []byte, startTS, commitTS uint64) (uint64, error) {
 	recordedCommitTS, committed, err := f.txnCommitTS(ctx, primaryKey, startTS)
 	if err != nil {
 		return 0, err
 	}
 	if !committed {
+		// No commit record yet: reject if a rollback marker is present.
+		// This catches out-of-order apply (COMMIT after ABORT), buggy
+		// clients, and replay races.
+		if _, rerr := f.store.GetAt(ctx, txnRollbackKey(primaryKey, startTS), ^uint64(0)); rerr == nil {
+			return 0, errors.WithStack(ErrTxnAlreadyAborted)
+		} else if !errors.Is(rerr, store.ErrKeyNotFound) {
+			return 0, errors.WithStack(rerr)
+		}
 		return startTS, nil
 	}
 	if recordedCommitTS != commitTS {
@@ -634,28 +648,39 @@ func (f *kvFSM) buildAbortCleanupStoreMutations(ctx context.Context, muts []*pb.
 }
 
 func (f *kvFSM) appendRollbackRecord(ctx context.Context, primaryKey []byte, startTS uint64, storeMuts *[]*store.KVPairMutation) error {
+	// Desired invariant: for any (primaryKey, startTS) pair, at most
+	// one of {rollback marker, commit record} is present. The invariant
+	// holds when aborts/commits flow through the symmetric guards in
+	// this function and handleCommitRequest, but we cannot *assume* it
+	// on entry (a buggy client, replay, or race may violate it), so we
+	// verify it in-line below on both the first-time and idempotent
+	// paths.
+	//
 	// Idempotent rollback: if the marker already exists for this
 	// (primaryKey, startTS), skip the Put. Rollback markers are
 	// deterministic ({txnRollbackVersion}) and a second Put against
 	// the already-tombstoned key would otherwise be rejected by the
 	// MVCC store as a write conflict (latestCommitTS > startTS).
-	//
-	// Safety: if the rollback marker is present, the commit record
-	// must be absent for this (primaryKey, startTS) because the
-	// commit-wins-over-rollback check below ran when the marker was
-	// first written, so we skip the commit check on the idempotent
-	// path.
+	markerPresent := false
 	if _, err := f.store.GetAt(ctx, txnRollbackKey(primaryKey, startTS), ^uint64(0)); err == nil {
-		return nil
+		markerPresent = true
 	} else if !errors.Is(err, store.ErrKeyNotFound) {
 		return errors.WithStack(err)
 	}
 
-	// Don't allow rollback to win after commit record exists.
+	// Verify the invariant regardless of marker presence: if a commit
+	// record is present for this (primaryKey, startTS), refuse to
+	// write (or confirm) a rollback marker. This catches out-of-order
+	// apply where a COMMIT somehow landed after a prior ABORT, as
+	// well as the normal "commit wins over rollback" race.
 	if _, err := f.store.GetAt(ctx, txnCommitKey(primaryKey, startTS), ^uint64(0)); err == nil {
 		return errors.WithStack(ErrTxnAlreadyCommitted)
 	} else if err != nil && !errors.Is(err, store.ErrKeyNotFound) {
 		return errors.WithStack(err)
+	}
+
+	if markerPresent {
+		return nil
 	}
 
 	*storeMuts = append(*storeMuts, &store.KVPairMutation{
--- a/kv/fsm.go
+++ b/kv/fsm.go
@@ -525,24 +525,17 @@ func (f *kvFSM) handleAbortRequest(ctx context.Context, r *pb.Request, abortTS u
 		return errors.WithStack(ErrTxnCommitTSRequired)
 	}
 
-	// Idempotency short-circuit: if the rollback marker for this
-	// (primaryKey, startTS) already exists, a previous abort already
-	// completed the whole cleanup atomically (ApplyMutations writes the
-	// rollback marker together with the lock/intent deletes in one
-	// batch, so the marker's presence proves cleanup ran). Without this
-	// guard a retry or a concurrent second lock-resolver would re-emit
-	// Delete mutations on already-tombstoned lock/intent keys and a
-	// duplicate rollback-marker Put — all three would be rejected by
-	// the MVCC store as write conflicts (latestCommitTS > startTS) and
-	// surface in prod as "secondary write failed" log spam without
-	// changing any state. Rollback markers are deterministic
-	// ({txnRollbackVersion}) so second-writer-wins would be equivalent
-	// anyway; skipping the work is simpler and cheaper.
-	if _, err := f.store.GetAt(ctx, txnRollbackKey(meta.PrimaryKey, startTS), ^uint64(0)); err == nil {
-		return nil
-	} else if !errors.Is(err, store.ErrKeyNotFound) {
-		return errors.WithStack(err)
-	}
+	// NOTE: do NOT short-circuit the whole request on rollback-marker
+	// presence. The marker only proves that SOME prior abort for this
+	// (primaryKey, startTS) ran; it does not prove cleanup ran for the
+	// specific keys in *this* request. In particular
+	// ShardStore.tryAbortExpiredPrimary issues an ABORT whose mutation
+	// list contains only the primary key, so a later lock-resolver
+	// abort for a secondary key (same primaryKey, same startTS) would
+	// see the marker already present and must still clean up that
+	// secondary's lock/intent. Idempotency is enforced per-key in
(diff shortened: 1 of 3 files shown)


bootjp:
-

bootjp:
Addressed gemini **HIGH** + 3 Copilot items. Pushed a1d8588f.

**gemini HIGH — commit/rollback consistency gap**

1. `appendRollbackRecord` now verifies the commit record is absent on **both** the first-time and idempotent paths (previously only on first-time). If the invariant is ever violated (e.g. buggy reordering wrote a commit after abort), abort returns `ErrTxnAlreadyCommitted`.
2. `commitApplyStartTS` now probes `txnRollbackKey(primaryKey, startTS)` before writing a commit record. If the marker is present, commit is rejected with a new `ErrTxnAlreadyAborted`. This is the reciprocal guard that was missing — the case where a COMMIT arrives after ABORT (replay, buggy client, race) no longer silently creates the inconsistent (marker + commit both present) state.

Together these close the gap: the invariant "at most one of {rollback marker, commit record} present" is now *enforced* on both sides, not assumed.

**Copilot #3 — PR body** rewritten to accurately describe per-key idempotency (`shouldClearAbortKey` + `appendRollbackRecord` marker-skip) rather than the incorrect "probe marker and return early" description.

**Copilot #4 — comment** in `appendRollbackRecord` now states the invariant is *desired* and *verified in-line*, not assumed.

**Copilot #5 — test comment** at `fsm_abort_test.go:303` reworded: "without performing additional writes or store mutations (reads are allowed — the idempotent path still probes for the rollback marker and commit record via GetAt)".

**New test**: `TestFSMAbort_CommitAfterAbortIsRejected` pins the symmetric guard. The flip direction is already covered by `TestFSMAbort_RejectsAlreadyCommittedTxn`.

**Error choice**: `ErrTxnAlreadyAborted = errors.New("txn already aborted")`, following the naming convention of the existing `ErrTxnAlreadyCommitted`.

Verification:
- `go test -race -count=1 ./kv/...` — PASS (3.846s)
- `make lint` — 0 issues

/gemini review

bootjp:
/gemini review

Addressed all 3 medium items: replaced `GetAt` + `errors.Is(err, ErrKeyNotFound)` presence checks with `store.ExistsAt` in `kv/fsm.go`:

1. `commitApplyStartTS` (~L475): rollback-marker presence check
2. `appendRollbackRecord` idempotent path (~L664): rollback-marker presence check
3. `appendRollbackRecord` commit-check (~L676): commit-record presence check

Semantics identical. `go test -race -count=1 ./kv/...` green; `make lint` clean.

Commit: 97993995

bootjp:
Thanks for the review, but I believe this one misreads what `appendRollbackRecord` is doing. The two `ExistsAt` calls here are **not** store-layer OCC conflict checks — they are application-layer invariant enforcement, and removing them would reintroduce exactly the correctness bug this PR was opened to fix.

The invariant (documented in the comment at `kv/fsm.go:652-665` and the symmetric comment on `commitApplyStartTS` at `kv/fsm.go:457-465`) is:

> For any `(primaryKey, startTS)` pair, at most one of `{txnRollbackKey, txnCommitKey}` must exist.

Each `ExistsAt` call has a specific, test-pinned purpose:

1. **`ExistsAt(txnCommitKey(pk, startTS), MaxTS)` → `ErrTxnAlreadyCommitted`** — rejects an abort that arrives for a txn whose COMMIT record already landed (commit-wins-over-abort). Pinned by `TestFSMAbort_RejectsAlreadyCommittedTxn` (`kv/fsm_abort_test.go:112-143`). Remove this and that test fails and the invariant breaks.

2. **`ExistsAt(txnRollbackKey(pk, startTS), MaxTS)` → skip Put** — idempotency for the marker. Rollback markers are deterministic (written at `txnRollbackVersion`), and a second `Put` against the already-tombstoned key would be rejected by the MVCC store as a write conflict (`latestCommitTS > startTS`). This is what makes lock-resolver races safe. Pinned by `TestFSMAbort_LockResolverRaceLeavesNoOrphan` (`kv/fsm_abort_test.go:433-482`), which fires `tryAbortExpiredPrimary` plus N per-secondary lock-resolver aborts plus a redundant replay for the same `(pk, startTS)` and asserts no orphan lock/intent remains.

3. The symmetric guard in `commitApplyStartTS` (`kv/fsm.go:466-482`) rejects COMMIT when a rollback marker is present (`ErrTxnAlreadyAborted`). Pinned by `TestFSMAbort_CommitAfterAbortIsRejected` (`kv/fsm_abort_test.go:514-554`). This is the other half of the same invariant and is exactly the reciprocal check the earlier Copilot `[high]` review asked for.

The OCC rule from `AGENTS.md` (\"timestamp issuance should originate from the Raft leader to prevent skewed OCC decisions\") is about **mutation-path write-conflict detection** — i.e. whether a user write at `commitTS` collides with a newer committed version. That is not what these probes do. These probes read txn-control records at `MaxTS` to verify a coexistence invariant before synthesizing store mutations; they do not influence MVCC conflict resolution and do not depend on any timestamp the store layer issues locally.

Keeping the checks. No code change.

/gemini review

bootjp:
Responding to the HIGH review on `kv/fsm.go:682` ("two `ExistsAt` calls need a lock / shared snapshot timestamp / startTS as readTS").

I think this review misreads the concurrency model. No code change is warranted.

**Serialization argument**: `appendRollbackRecord` is only reachable via `kvFSM.Apply` -> `applyRequestErr` -> `handleRequest` -> `handleTxnRequest` -> `handleAbortRequest`. `hashicorp/raft` drives `FSM.Apply` from a single goroutine (`runFSM` in `raft@v1.7.3/fsm.go:86`, consuming `fsmMutateCh`). The `raft.FSM` interface docs state explicitly: *"Apply and Snapshot are always called from the same thread"* (`raft@v1.7.3/fsm.go:34`). Therefore, between the `ExistsAt` on `txnRollbackKey` (`fsm.go:666`) and the `ExistsAt` on `txnCommitKey` (`fsm.go:676`), no other `Apply` invocation can run against this FSM, and thus no other writer can insert/remove either record. An additional user-space lock adds nothing; there is no state variable to protect beyond what the MVCC store already commits durably.

The only concurrent raft-side callback is `FSMSnapshot.Persist`, which is read-only against a captured snapshot handle (see `kv/snapshot.go`) and cannot mutate `txnRollbackKey`/`txnCommitKey`.

**Read timestamp choice**: Both reads deliberately use `MaxTS` (`^uint64(0)`) because we want to observe the latest committed state of the store at apply time, not a snapshot at the transaction's `startTS`. Using `startTS` as the read timestamp would be a bug: the commit record for this txn is written with `commitTS > startTS` (asserted in `TestFSMAbort_AbortTSMustBeGreaterThanStartTS`), so a read at `startTS` would never see it, defeating the whole "commit wins over rollback" invariant check at `fsm.go:680`. Since `Apply` is serialized, the two `MaxTS` reads are on a logically identical store snapshot - no read-skew is possible.

**Tests that would catch a regression**: `TestFSMAbort_CommitAfterAbortIsRejected`, `TestFSMAbort_RejectsAlreadyCommittedTxn`, `TestFSMAbort_SecondAbortSameKeysIsIdempotent`, `TestFSMAbort_RollbackRecord` in `kv/fsm_abort_test.go`.

If I've missed a concurrency path (a non-`Apply` writer of these keys, a second FSM instance sharing the store, a background rewriter, etc.), please cite the specific call site and I will reopen.

/gemini review

bootjp:
@gemini-code-assist Pushing back on the `appendRollbackRecord` medium (kv/fsm.go:686) — keeping the `commitExists` check on the idempotent path is deliberate. Brief history:

1. That check was ADDED SPECIFICALLY in response to an earlier gemini HIGH review in this same PR (commit `a1d8588f`, "fix(kv): reject commit-after-abort and tighten rollback idempotency"). The earlier review argued: *"the safety claim relies on `handleCommitRequest` performing a reciprocal check for the rollback marker... if a COMMIT request is applied after an ABORT has already written the rollback marker... `handleCommitRequest` would miss it."* Reordering/short-circuiting now reverts that fix and loses defense-in-depth.

2. Cost: it's a single `ExistsAt` (presence-only, no value clone) on what is already the uncommon path (lock-resolver race, retry after a previous abort). It is NOT the hot path — the prod log spam that motivated this PR came from `appendRollbackRecord`'s *first-time* path tripping on a write conflict, and the idempotent fast-return still short-circuits the marker Put before any mutation is appended. Reordering saves nothing on first-time traffic and saves one `ExistsAt` on an already-cold retry path.

3. The symmetric check in `commitApplyStartTS` (kv/fsm.go:466-482) is the primary defense against a commit-after-abort apply ordering, but the in-line verification here in `appendRollbackRecord` catches buggy-reorder / replay / dual-write bugs that could bypass `handleCommitRequest` entirely. `TestFSMAbort_CommitAfterAbortIsRejected` pins exactly this behavior — a commit landing after an abort must be rejected no matter which FSM entry point sees it first.

4. Removing it would require trusting the invariant check in `commitApplyStartTS` to be the only line of defense — which is exactly what Copilot warned against in the earlier round: *"adjust the safety comment to avoid asserting an invariant [relying on checks elsewhere]"*. The current code follows that guidance by verifying in-line on both the first-time and idempotent paths.

Given 1-4, I'd like you to explicitly acknowledge whether giving up defense-in-depth (and reverting the earlier HIGH fix from `a1d8588f`) is the intended tradeoff before I consider the reorder. If the answer is "yes, the commitApplyStartTS check is sufficient and the symmetric in-line check is redundant", I'll weigh that against `TestFSMAbort_CommitAfterAbortIsRejected`'s contract; if not, the current structure stays.

/gemini review

PR #582, opened 2026-04-22
test(redis): wait for follower apply catch-up in MultiExec list tests

bootjp:
## Summary
- Root-cause `TestRedis_MultiExec_DelThenRPushRecreatesList` flake: the test reads `readTS()` directly on `nodes[1]`, which returns the node-local `store.LastCommitTS()`. That value can trail the raft-committed EXEC on a follower (or a leader that just responded), while the client-facing LRANGE path waits for local apply via LeaseRead/LinearizableRead. The direct `ScanAt(readTS)` therefore observes the pre-EXEC state.
- Fix by introducing `waitForListState` helper that polls `resolveListMeta` + raw item scan at the node-local `readTS` via `require.Eventually` (250ms / 5s) until both reflect the expected post-EXEC state. Real synchronization on raft apply progress, not a sleep.
- Swap the ad-hoc `readTS/resolveListMeta/ScanAt` block in `TestRedis_MultiExec_DelThenRPushRecreatesList` for the helper.

## Test plan
- [x] `go test -race -count=10 -run TestRedis_MultiExec ./adapter/` -> 10/10 pass (20.5s)
- [x] `go test -race -count=1 -short ./adapter/...` passes (~62s)
- [x] `make lint` clean (0 issues)

### DIFF
--- a/adapter/redis_multi_test.go
+++ b/adapter/redis_multi_test.go
@@ -32,6 +32,10 @@ import (
 // gap deterministically without a timing-based sleep.
 func waitForListState(t *testing.T, n Node, key []byte, expectedLen int, expectedValues []string) {
 	t.Helper()
+	if expectedValues != nil && len(expectedValues) != expectedLen {
+		t.Fatalf("waitForListState: expectedValues length %d does not match expectedLen %d",
+			len(expectedValues), expectedLen)
+	}
 	require.Eventually(t, func() bool {
 		return listStateMatches(n, key, expectedLen, expectedValues)
 	}, 5*time.Second, 250*time.Millisecond,
--- a/adapter/redis_multi_test.go
+++ b/adapter/redis_multi_test.go
@@ -36,8 +36,23 @@ func waitForListState(t *testing.T, n Node, key []byte, expectedLen int, expecte
 	require.Eventually(t, func() bool {
 		readTS := n.redisServer.readTS()
 		meta, exists, err := n.redisServer.resolveListMeta(ctx, key, readTS)
-		if err != nil || !exists || meta.Len != int64(expectedLen) {
+		if err != nil {
 			return false
+		}
+		// Redis represents an empty / deleted list as the absence of
+		// the meta key. When the caller is verifying deletion
+		// (expectedLen == 0), !exists is the success signal;
+		// requiring exists==true would make this helper unusable for
+		// post-DEL waits. For non-empty expectations a missing meta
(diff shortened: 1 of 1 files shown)


bootjp:
Addressed gemini's medium (d63f92bc):

- `waitForListState` now accepts `!exists` as success when `expectedLen == 0`. In Redis an empty / deleted list is represented by the absence of its meta key; the previous check was failing post-DEL waits.
- For `expectedLen > 0`, `!exists` still means "not yet applied" and keeps polling.

Verified: `go test -race -count=10 -run TestRedis_MultiExec ./adapter/` 10/10 green.

/gemini review

bootjp:
Refactored `waitForListState` to satisfy `cyclop` (was 12, limit 10).

- Split the `require.Eventually` body into four helpers:
  - `listStateMatches` — single-shot predicate
  - `listMetaMatches` — meta + `expectedLen==0` empty/absent semantics
  - `scanListItems` — `ScanAt` + length check
  - `listValuesMatch` — optional value equality
- No behavior change; polling/timeouts and the `expectedLen==0` branch preserved.
- `go test -race -count=10 -run TestRedis_MultiExec ./adapter/` → pass (20.7s).
- `make lint` → 0 issues.

Commit: 403367eb

/gemini review

bootjp:
Addressed the latest two review threads in 9e0258d7:

- **waitForListState misuse guard (review 14:04 UTC) + listValuesMatch panic (review 14:08 UTC, line 99)**: Added a single fail-fast check at the top of `waitForListState`: if `expectedValues != nil && len(expectedValues) != expectedLen`, the helper `t.Fatalf`s immediately. This centralizes the invariant at the public entry point, so `scanListItems`'s existing `len(kvs) == expectedLen` guarantee plus this length check means `listValuesMatch` cannot panic. Keeping the check in one place instead of duplicating into both leaf helpers preserves readability.

- **Rejected: empty-list / deleted-list check (review 13:46 UTC)**: This comment describes the pre-fix code. The current `listMetaMatches` (lines 60-69) already handles `expectedLen == 0` with `!exists || meta.Len == 0`, landed in commit d63f92bc before the review snapshot was taken. No change needed.

Test results on this commit:
- `go test -race -count=10 -run TestRedis_MultiExec ./adapter/` -> 10/10 pass (20.9s)
- `go test -race -count=1 -short ./adapter/...` -> pass (64.9s)
- `make lint` -> 0 issues

/gemini review

PR #585, opened 2026-04-22
obs(proxy): classify secondary write failures by reason

bootjp:
## Motivation

After #581 we still see occasional secondary-write failures in production with payloads like `!txn|rb|!redis|ttl|...` (OCC write conflicts racing against rollback tombstones). Today the only signal is the `secondary write failed` log line from `recordSecondaryWriteFailure` — the unlabelled `proxy_secondary_write_errors_total` counter tells us something is wrong but gives no hint whether the failure is an OCC conflict, a retry-loop giving up, a lost leader, or a timeout. Ops have been grepping logs to triage.

## Change

- Adds `proxy_secondary_write_errors_by_reason_total{cmd,reason}` (CounterVec) alongside the existing unlabelled counter (kept for dashboard backwards-compat).
- New helper `classifySecondaryWriteError(err) string` maps error strings to one of:
  - `retry_limit` — `"retry limit exceeded"` (checked first, since the retry-limit message embeds `"write conflict"`)
  - `write_conflict` — raw OCC conflict, e.g. `!txn|rb|` races
  - `deadline_exceeded` — `errors.Is(err, context.DeadlineExceeded)` or substring
  - `not_leader` — leadership lost mid-dispatch
  - `txn_already_finalized` — `"txn already committed"` / `"txn already aborted"`
  - `other` — everything else
- `recordSecondaryWriteFailure` increments both counters.
- Tests: table test for the classifier + a test asserting both counters tick.

## Suggested Grafana query

```promql
sum by (cmd, reason) (rate(proxy_secondary_write_errors_by_reason_total[5m]))
```

A per-reason stacked graph makes the `!txn|rb|` conflict pattern (the motivating incident) show up as a spike in `reason="write_conflict"` without needing to touch logs.

## Test plan

- [x] `go test -race -count=1 ./proxy/...`
- [x] `make lint`

## Summary by CodeRabbit

* **Chores**
  * Secondary write error monitoring now automatically classifies failures by specific error type to improve diagnostics and observability.
  * New detailed metrics track write failure patterns across different error categories, enabling better visibility and root cause analysis.

* **Tests**
  * Added comprehensive unit tests validating error classification logic and metrics recording functionality across various failure scenarios.

PR #587, opened 2026-04-22
obs(store): track MVCC write conflicts by key-prefix class

bootjp:
## Summary

- Adds a protocol-independent OCC write-conflict metric at the MVCC store layer. Both `checkConflicts` (write-write) and `checkReadConflicts` (read-write) in `store/lsm_store.go` / `store/mvcc_store.go` now increment a bounded per-(kind, key_prefix) counter immediately before returning `NewWriteConflictError`; a new `monitoring.WriteConflictCollector` polls the stores every 5 s (matching the existing `DispatchCollector` / `PebbleCollector` cadence) and mirrors deltas into `elastickv_store_write_conflict_total{group, kind, key_prefix}`.
- `kind` splits `read` (RW conflict via read set) from `write` (WW conflict via mutation set) since ops implications differ. `key_prefix` uses a bounded classification aligned with existing prefix constants (`!txn|lock|`, `!txn|rb|`, `!redis|str|`, `!hs|`, `!zs|`, `!lst|`, `!ddb|`, ...) with an `other` fallback so user-supplied keys cannot grow cardinality.

## Motivation

`#585` adds the proxy-side view (per client request). This PR adds the underlying store-side view (per Raft-applied proposal), so the signal is visible regardless of protocol adapter (Redis, DynamoDB, raw KV). In the `!txn|rb|` production incident the proxy counter spiked per user request but a single request fanned out to many Raft proposals; the store-side metric would have surfaced the real pressure (txn_rollback bucket) directly.

## Sample Grafana query

```
sum by (key_prefix) (rate(elastickv_store_write_conflict_total[5m]))
```

Split by conflict kind:

```
sum by (kind, key_prefix) (rate(elastickv_store_write_conflict_total[5m]))
```

## Test plan

- [x] `go test -race -count=1 ./store/... ./monitoring/... ./adapter/... ./kv/...`
- [x] `make lint`
- [x] New unit tests:
  - `store/write_conflict_counter_test.go`: table test for `classifyWriteConflictKey` covering every bucket; counter record / snapshot / nil-safety.
  - `store/write_conflict_store_test.go`: integration tests for both pebble-backed and in-memory MVCC stores exercising the WW and RW conflict paths and asserting the right bucket increments.
  - `monitoring/write_conflict_test.go`: collector delta behaviour, source-reset (counter-down) rebasing, malformed-label skip, nil-safety.

### DIFF
--- a/kv/leader_routed_store.go
+++ b/kv/leader_routed_store.go
@@ -279,6 +279,16 @@ func (s *LeaderRoutedStore) LastCommitTS() uint64 {
 	return s.local.LastCommitTS()
 }
 
+// WriteConflictCountsByPrefix delegates to the local MVCC store. The
+// leader-routed wrapper does not add cross-group conflict detection of
+// its own, so the node-local view IS the authoritative view.
+func (s *LeaderRoutedStore) WriteConflictCountsByPrefix() map[string]uint64 {
+	if s == nil || s.local == nil {
+		return map[string]uint64{}
+	}
+	return s.local.WriteConflictCountsByPrefix()
+}
+
 const globalLastCommitTSTimeout = 200 * time.Millisecond
 
 // GlobalLastCommitTS returns the most recently committed HLC timestamp from
--- a/kv/shard_store.go
+++ b/kv/shard_store.go
@@ -1174,6 +1174,23 @@ func (s *ShardStore) LastCommitTS() uint64 {
 	return max
 }
 
+// WriteConflictCountsByPrefix aggregates OCC conflict counts across
+// every shard group owned by this ShardStore. Per-shard counts share
+// the same "<kind>|<key_prefix>" label schema, so a simple sum gives
+// the node-wide view. The result is always non-nil.
+func (s *ShardStore) WriteConflictCountsByPrefix() map[string]uint64 {
+	out := map[string]uint64{}
+	for _, g := range s.groups {
+		if g == nil || g.Store == nil {
+			continue
+		}
+		for label, count := range g.Store.WriteConflictCountsByPrefix() {
+			out[label] += count
+		}
+	}
+	return out
+}
+
 func (s *ShardStore) Compact(ctx context.Context, minTS uint64) error {
 	for _, g := range s.groups {
 		if g == nil || g.Store == nil {
--- a/main.go
+++ b/main.go
@@ -184,13 +184,7 @@ func run() error {
 		adapter.WithDistributionCoordinator(coordinate),
 		adapter.WithDistributionActiveTimestampTracker(readTracker),
 	)
-	metricsRegistry.RaftObserver().Start(runCtx, raftMonitorRuntimes(runtimes), raftMetricsObserveInterval)
-	if collector := metricsRegistry.DispatchCollector(); collector != nil {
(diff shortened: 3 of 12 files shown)


bootjp:
@gemini-code-assist Both medium-priority comments (store/lsm_store.go:872 and store/mvcc_store.go:502) appear to misread the diff. Clarifying with evidence so you can re-evaluate.

## This PR does not add `checkConflicts` / `checkReadConflicts`

Both functions are pre-existing. The PR only adds a counter increment alongside the existing `NewWriteConflictError` return. From `git show 9a13db4c -- store/lsm_store.go store/mvcc_store.go`:

```diff
 		if exists && ts > startTS {
+			s.writeConflicts.record(WriteConflictKindWrite, classifyWriteConflictKey(mut.Key))
 			return NewWriteConflictError(mut.Key)
 		}
```

```diff
 		if latestVer, ok := s.latestVersionLocked(mut.Key); ok && latestVer.TS > startTS {
+			s.writeConflicts.record(WriteConflictKindWrite, classifyWriteConflictKey(mut.Key))
 			return NewWriteConflictError(mut.Key)
 		}
```

No new OCC check is introduced. This is an observability PR: it surfaces the conflict count per `(kind, key_prefix)` for Prometheus. Removing the OCC check was never on the table here, and doing so would be a correctness regression.

## The AGENTS.md rule is about timestamp issuance, not disabling OCC

AGENTS.md:35 (the only rule matching your wording):

> Avoid leader-local timestamps in persistence; timestamp issuance should originate from the Raft leader to prevent skewed OCC decisions.

Plain reading: the leader must issue timestamps (HLC/TSO) so OCC decisions use a consistent clock. It does not say "OCC checks at the store layer are redundant." In fact it presupposes that OCC decisions happen — the rule exists to make them correct. HLC and OCC are orthogonal: HLC makes `startTS`/`commitTS` globally monotonic; OCC enforces `commitTS > last write` per key and `>= all reads`. Two writers sharing a `startTS` can still collide on a key even with perfect HLC; OCC rejects the loser.

`docs/collection_metadata_delta_design.md:90` even spells it out: Raft leader uniqueness guarantees single-source `commitTS`, and OCC on field/member keys is what provides mutual exclusion.

## Load-bearing tests that would break if OCC were removed

- `store/lsm_store_txn_test.go:101` `TestPebbleStore_ApplyMutations_WriteConflict` — direct write-write conflict.
- `store/lsm_store_txn_test.go:472` `TestPebbleStore_ApplyMutations_ReadConflict` — direct read-write conflict via `checkReadConflicts`.
- `kv/fsm_occ_test.go:75` `TestOnePhaseTxnDetectsWriteConflict` and `kv/fsm_occ_test.go:110` `TestOnePhaseTxnDetectsReadWriteConflictViaReadKeys` — FSM-level OCC.
- `kv/fsm_abort_test.go:306` `TestFSMAbort_SecondAbortSameKeysIsIdempotent` — abort-path semantics that depend on the store-level conflict check being honored.

## Offer

If the underlying concern is coupling between the hot OCC path and the observability counter, I am happy to move the `record(...)` call into a small wrapper near the `NewWriteConflictError` return site so OCC logic stays visually isolated. Happy to do that as a follow-up if you confirm.

/gemini review

PR #588, opened 2026-04-22
perf(pebble): set explicit 256 MiB block cache with env override

bootjp:
## Summary

- Production Pebble block-cache hit rate was 0.003% (1.8B misses vs 58k hits, `block_cache_size_bytes=0`) because `defaultPebbleOptions` never set `Cache` and `EnsureDefaults` supplied pebble's built-in 8 MiB baseline, which churned immediately against the working set.
- Populate `Cache: pebble.NewCache(pebbleCacheBytes)` with a 256 MiB per-store default. Override contract: `ELASTICKV_PEBBLE_CACHE_MB` (integer MiB, clamped to `[8, 65536]`; empty / malformed / out-of-range values fall back to the default, including `"0"` which is below the floor).
- Track the cache ref on `pebbleStore` and `Unref` it in `Close`, `reopenFreshDB`, and `swapInTempDB`; transient tmpDB opens (snapshot restore) Unref their cache on the way out so a restore cycle no longer leaks a 256 MiB cache.
- Add `elastickv_pebble_block_cache_capacity_bytes` gauge in `monitoring/pebble.go`, sourced from a new optional `PebbleCacheCapacitySource` interface (`*pebbleStore.BlockCacheCapacityBytes()` returns `Cache.MaxSize()`). Operators can now see configured capacity alongside current usage.

A process-wide shared cache plumbed through `NewPebbleStore` is deferred to a follow-up PR (marked with a `TODO(perf/pebble)` comment) because it requires changing the constructor signature and all call sites.

## Test plan

- [x] `go test -race -count=1 ./store/... ./monitoring/...`
- [x] `make lint`
- [x] `TestPebbleCacheEnvOverride` covers empty / garbage / 0 / below floor / at both bounds / above ceiling / negative
- [x] `TestDefaultPebbleOptionsCarriesCache` verifies the cache is wired at the configured size
- [x] `TestPebbleCollectorEmitsBlockCacheCapacity` and `TestPebbleCollectorSkipsCapacityWhenUnsupported` cover the new capacity gauge
- [ ] Observe production `elastickv_pebble_block_cache_capacity_bytes == 268435456` and hit rate climbing after rollout

### DIFF
--- a/monitoring/pebble.go
+++ b/monitoring/pebble.go
@@ -47,9 +47,10 @@ type PebbleMetrics struct {
 	memtableZombieCount *prometheus.GaugeVec
 
 	// Block cache.
-	blockCacheSizeBytes   *prometheus.GaugeVec
-	blockCacheHitsTotal   *prometheus.CounterVec
-	blockCacheMissesTotal *prometheus.CounterVec
+	blockCacheSizeBytes     *prometheus.GaugeVec
+	blockCacheCapacityBytes *prometheus.GaugeVec
+	blockCacheHitsTotal     *prometheus.CounterVec
+	blockCacheMissesTotal   *prometheus.CounterVec
 }
 
 func newPebbleMetrics(registerer prometheus.Registerer) *PebbleMetrics {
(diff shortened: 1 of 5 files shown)


bootjp:
@codex code review

bootjp:
@codex code review

PR #589, opened 2026-04-22
fix(raft): retention GC for fsm snapshots + tunable snapshot trigger

bootjp:
## Findings

- **Snapshot trigger** — `defaultSnapshotEvery = 10_000` entries past the last snapshot. Hard-coded; no env override. With fat Lua-script proposals, 10k entries easily produces the multi-GiB gap we observed.
- **fsm-snap/ retention** — healthy. `purgeOldSnapshotFiles` (keep `defaultMaxSnapFiles = 3`) is called after every `persist.SaveSnap` in both snapshot paths (`persistCreatedSnapshot`, `persistLocalSnapshotPayload`). The 2.3 GB bloat is three genuinely-large snapshots, not retention leakage.
- **WAL retention — BROKEN.** The engine calls `persist.Release(snap)` correctly, which invokes `wal.ReleaseLockTo(snap.Metadata.Index)`. But that method only CLOSES the file locks on obsolete segments; it does not delete them. etcd-server normally runs a separate `fileutil.PurgeFile` goroutine to unlink old `.wal` segments. elastickv was missing that entirely — grep confirms zero `PurgeFile` or equivalent call in the tree. Segments are 64 MiB by default, so 6.4 GB ≈ 100 segments, i.e. every segment since cluster bootstrap.

## Change

- New `purgeOldWALFiles(walDir, keep)` in `internal/raftengine/etcd/wal_purge.go`. Sorts `.wal` names lex (== chronological for etcds `%016x-%016x.wal` format), and for each deletion candidate attempts `fileutil.TryLockFile` first — a still-held flock means the wal package has not released that segment, so we skip it and pick it up on the next snapshot. Safe against deleting the active tail.
- Wired into both snapshot persistence paths, right after `purgeOldSnapshotFiles`. Default keep = 5 segments (~320 MiB cap).
- New env vars (both fall back to current defaults on invalid/unset input):
  - `ELASTICKV_RAFT_SNAPSHOT_COUNT` — override the 10_000 snapshot-trigger threshold.
  - `ELASTICKV_RAFT_MAX_WAL_FILES` — override the 5-segment WAL retention cap.
- No changes to fsm-snap retention — it already worked.

## How retention is now bounded

Per node, after the first snapshot following deploy:
- `wal/` ≤ `ELASTICKV_RAFT_MAX_WAL_FILES` × 64 MiB (default 320 MiB).
- `fsm-snap/` ≤ `defaultMaxSnapFiles` × snapshot size (unchanged: 3 × size).
- Snapshot cadence: once per `ELASTICKV_RAFT_SNAPSHOT_COUNT` applied entries (default 10_000; consider dropping to 1_000–2_000 on fat-proposal workloads).

## Deploy runbook

After merge + rollout, the stale WAL and fsm-snap files on disk will NOT self-clean until the next snapshot runs (retention is prospective, not historical). Recommend: on each node, after confirming the new binary is running and the first new snapshot has been observed in Grafana, manually `rm` the stale `fsm-snap/*.fsm` files older than the 2 newest and the stale `wal/*.wal` segments older than the 5 newest.

## Test plan
- [x] `go test -race -count=1 ./internal/raftengine/etcd/...` — passes (8.9 s).
- [x] `make lint` — clean.
- [x] New tests cover: keep-N invariant, never-deletes-tail, missing dir is no-op, clamps keep=0 to 1, ignores non-.wal entries, env-var parsing (default/override/invalid/clamp).
- [ ] Stage deploy on one node; verify wal/ stays bounded across multiple snapshot cycles.

## Release Notes

* **New Features**
  * Added environment variable configuration for Raft snapshot frequency and WAL file retention limits, allowing operators to optimize storage usage and durability characteristics for their deployment.
  * WAL files are now automatically purged according to configured retention policies, preventing unbounded disk usage.

* **Tests**
  * Added comprehensive test coverage for WAL retention, purging logic, and configuration parsing with edge case validation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/tidwall/redcon v1.6.2
 	github.com/vmihailenco/msgpack/v5 v5.4.1
 	github.com/yuin/gopher-lua v1.1.2
+	go.etcd.io/etcd/client/pkg/v3 v3.6.10
 	go.etcd.io/etcd/server/v3 v3.6.10
 	go.etcd.io/raft/v3 v3.6.0
 	go.uber.org/zap v1.27.1
@@ -100,7 +101,6 @@ require (
 	github.com/vmihailenco/tagparser/v2 v2.0.0 // indirect
 	go.etcd.io/bbolt v1.4.3 // indirect
 	go.etcd.io/etcd/api/v3 v3.6.10 // indirect
-	go.etcd.io/etcd/client/pkg/v3 v3.6.10 // indirect
 	go.etcd.io/etcd/client/v3 v3.6.10 // indirect
 	go.etcd.io/etcd/pkg/v3 v3.6.10 // indirect
 	go.opentelemetry.io/auto/sdk v1.2.1 // indirect
--- a/internal/raftengine/etcd/engine.go
+++ b/internal/raftengine/etcd/engine.go
@@ -181,15 +181,26 @@ type Engine struct {
 	// once at Open from ELASTICKV_RAFT_DISPATCHER_LANES so the run-time code
 	// path is branch-free per message and does not need to re-read env vars.
 	dispatcherLanesEnabled bool
-	dispatchStopCh         chan struct{}
-	dispatchCtx            context.Context
-	dispatchCancel         context.CancelFunc
-	snapshotReqCh          chan snapshotRequest
-	snapshotResCh          chan snapshotResult
-	snapshotStopCh         chan struct{}
-	closeCh                chan struct{}
-	doneCh                 chan struct{}
-	startedCh              chan struct{}
+	// snapshotEvery is the FSM-snapshot trigger threshold captured once at
+	// Open from ELASTICKV_RAFT_SNAPSHOT_COUNT. maybePersistLocalSnapshot
+	// runs on every Ready-drain pass, so re-parsing the env var (and
+	// potentially emitting a warning on malformed input) on every call
+	// would flood logs and burn CPU on the raft hot path. Cache it once.
+	snapshotEvery uint64
+	// maxWALFiles is the WAL retention cap captured once at Open from
+	// ELASTICKV_RAFT_MAX_WAL_FILES. Purges are relatively rare (only after
+	// snapshot release) but caching avoids a second warning-per-invalid
+	// call site and keeps the knob consistent across the engine lifetime.
+	maxWALFiles    int
+	dispatchStopCh chan struct{}
+	dispatchCtx    context.Context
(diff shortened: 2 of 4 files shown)


bootjp:
Addressed three review items in this push (3b341e18..888cf1e6):

**1. codex P2 — non-lock WAL purge errors were silently swallowed** (`wal_purge.go:100`)
`purgeOldWALFiles` previously treated EVERY `fileutil.TryLockFile` failure as "segment is still active." `TryLockFile` also returns open/permission/I/O errors, so under a filesystem fault WAL retention was effectively disabled with no warning at default log levels. Fix: only `errors.Is(err, fileutil.ErrLocked)` skips silently; other errors are now logged at warn level and aggregated into the returned error. The lock safety invariant is preserved and strengthened — spurious lock failures no longer hide real I/O faults. Commit `5f452e2f`.

**2. coderabbit Major — persistCreatedSnapshot bypassed the Open-time cache** (`engine.go:2098`)
`persistCreatedSnapshot` still called `maxWALFilesFromEnv()` directly while the sibling `persistLocalSnapshotPayload` at line 3335 already went through `e.walRetention()` (the cache added in 3b341e18). That re-parsed the env var on every snapshot persist (re-emitting the "invalid ELASTICKV_RAFT_MAX_WAL_FILES" warning) and meant the two paths could disagree if the env var was mutated in-process. Both paths now share `e.walRetention()`. Commit `df2b41ff`.

**3. coderabbit Minor — flock-skip path was never exercised** (`wal_purge_test.go:88`)
The prior `TestPurgeOldWALFiles_NeverDeletesActiveTail` claimed in its docstring to hold a flock on the newest segment, but no lock was ever taken. With `keep=1` the newest file was excluded by the lexicographic cutoff alone, so the test was only a duplicate of `TestPurgeOldWALFiles_KeepsMostRecentN`. Renamed it to `TestPurgeOldWALFiles_NewestSurvivesKeepCutoff` with a comment that matches its actual behaviour, and ADDED `TestPurgeOldWALFiles_SkipsLockedMidListSegment` which grabs a real `fileutil.TryLockFile` on a mid-list segment (index `len(names)/2`) and asserts: newest survives (keep cutoff), locked mid-list segment survives (flock path), all older unlocked segments are purged. Skipped on Windows where fileutil flock semantics differ. Commit `888cf1e6`.

**Verification**
- `go test -race -count=1 -timeout 300s ./internal/raftengine/etcd/...` — PASS (8.83s)
- `TestPurgeOldWALFiles_SkipsLockedMidListSegment` — PASS (confirms the new flock path is now covered)
- `make lint` — 0 issues

/gemini review

@codex review

PR #592, opened 2026-04-23
perf(store): add ELASTICKV_FSM_SYNC_MODE for FSM apply fsync opt-out

bootjp:
## Root cause

Every committed raft entry triggered a `pebble.Sync` inside `store.ApplyMutations` / `store.DeletePrefixAt` (see `store/lsm_store.go:1056, 1108`), on top of the raft WAL fsync that `etcd/raft` already performs per `Ready` batch in `persistReadyToWAL` (`internal/raftengine/etcd/wal_store.go:376`).

The raft Ready loop in `drainReady` (`internal/raftengine/etcd/engine.go:1389`) already batches multiple entries per `Ready`, so the raft WAL fsync is fine. The hot fsync is the FSM-side `b.Commit(pebble.Sync)` that the apply loop (`applyCommitted`) hits once per entry. A prior cleanup (`docs/review_todo.md` section 3.4) intentionally kept `ApplyMutations` on `pebble.Sync`; this CL makes that choice tunable.

Microbenchmark (Apple M1 Max, APFS tempdir, `BenchmarkApplyMutations_SyncMode`):

| mode   | ns/op    | allocs/op |
|--------|---------:|----------:|
| sync   | 16292899 | 7         |
| nosync |    16293 | 8         |

~1000x on this platform. Real hardware fsync latency varies, but the sync/nosync ratio is consistently large on any WAL that enforces platform durability.

## Durability argument

Pebble's FSM-commit fsync is redundant with the raft WAL under this codebase's crash-recovery model:

1. Raft WAL (`etcd/raft`) fsyncs every committed entry via `persist.Save` before `Advance`.
2. On restart, `newMemoryStorage` (`internal/raftengine/etcd/persistence.go:352`) reloads the snapshot + all WAL entries. `newRawNode` does not set `Config.Applied`, so `etcdraft` defaults it to `snapshot.Metadata.Index`.
3. The engine sets `e.applied = maxAppliedIndex(LocalSnap)` and every committed entry past the snapshot is re-delivered through `CommittedEntries` on the first `Ready`.
4. `kv/fsm.applyCommitWithIdempotencyFallback` treats an already-committed key (`LatestCommitTS >= commitTS`) as an idempotent retry, so replaying an entry whose effect survived the crash is safe.
5. FSM snapshots are fsynced (`writeFSMSnapshotFile` then `f.Sync()` in `fsm_snapshot_file.go`).

Therefore a crash that loses the unfsynced tail of Pebble's own WAL is recoverable: raft replays from the last fsynced FSM snapshot onwards, and the idempotent apply path re-materialises the lost state. Pebble on the FSM commit path effectively becomes a volatile cache of applied state whose durability boundary is the raft WAL.

Other `pebble.Sync` call sites (snapshot-batch commit, metadata-restore writes, compaction `commitSnapshotBatch`) are untouched: those are orthogonal durability boundaries (e.g. restore-directory swap) and are not per-proposal cost.

## Env var + default

* `ELASTICKV_FSM_SYNC_MODE=sync` (default) - current behaviour.
* `ELASTICKV_FSM_SYNC_MODE=nosync` - `b.Commit(pebble.NoSync)` on the FSM hot path. Raft WAL remains the durability boundary.

Unknown values fall back to `sync` (fail-safe toward durability). Parsing is case-insensitive and whitespace-tolerant.

A Prometheus gauge `elastickv_fsm_apply_sync_mode{mode="sync"|"nosync"}` is set at `NewRegistry` time via `store.FSMApplySyncModeLabel()`, so dashboards can alert if a rolling deploy accidentally flips the durability posture.

## Test plan

- [x] `go test ./store/... ./monitoring/... ./kv/... ./internal/raftengine/... -count=1`
- [x] Env var parsing: sync/nosync/mixed-case/whitespace/unknown then sync default
- [x] Functional equivalence of sync vs nosync on a Pebble store
- [x] Clean-shutdown reopen visibility (NoSync + Close + reopen preserves writes)
- [x] Prometheus gauge mutual exclusivity across successive SetFSMApplySyncMode calls
- [ ] (Follow-up) Jepsen-style OS-level crash test for unfsynced-tail recovery - tracked in `JEPSEN_TODO.md`

## Benchmark

```
go test ./store -run='^$' -bench='BenchmarkApplyMutations_SyncMode' -benchtime=2s -benchmem
BenchmarkApplyMutations_SyncMode/sync-10      141   16292899 ns/op   368 B/op   7 allocs/op
BenchmarkApplyMutations_SyncMode/nosync-10    129262     16293 ns/op   284 B/op   8 allocs/op
```

## Related

- Previous lever documented in `docs/review_todo.md` section 3.4 (ApplyMutations retained `pebble.Sync`).
- Pebble block cache default 256 MiB (#588)
- WAL retention / purge (#589)
- Lease read via AppliedIndex (#575)
- Raft dispatcher lanes (#577)

## Summary by CodeRabbit

* **New Features**
  * Configuration option for FSM commit durability mode with adjustable sync and no-sync settings
  * New Prometheus metric exposing the current durability mode on the `/metrics` endpoint

* **Tests**
  * Unit tests validating FSM durability mode functionality
  * Integration tests ensuring functional correctness across different durability modes
  * Performance benchmarks evaluating FSM commit operation characteristics

### DIFF
--- a/main.go
+++ b/main.go
@@ -117,6 +117,10 @@ func run() error {
 	var lc net.ListenConfig
 
 	metricsRegistry := monitoring.NewRegistry(*raftId, *myAddr)
+	// Record the active FSM apply sync mode so operators can see on the
+	// /metrics endpoint which durability posture this node is running in.
+	// See store.FSMApplySyncModeLabel for the underlying env resolution.
+	metricsRegistry.SetFSMApplySyncMode(store.FSMApplySyncModeLabel())
 
 	// Create the shared HLC before building shard groups so every FSM can update
 	// physicalCeiling when HLC lease entries are applied to the Raft log.
--- a/main.go
+++ b/main.go
@@ -117,10 +117,6 @@ func run() error {
 	var lc net.ListenConfig
 
 	metricsRegistry := monitoring.NewRegistry(*raftId, *myAddr)
-	// Record the active FSM apply sync mode so operators can see on the
-	// /metrics endpoint which durability posture this node is running in.
-	// See store.FSMApplySyncModeLabel for the underlying env resolution.
-	metricsRegistry.SetFSMApplySyncMode(store.FSMApplySyncModeLabel())
 
 	// Create the shared HLC before building shard groups so every FSM can update
 	// physicalCeiling when HLC lease entries are applied to the Raft log.
@@ -141,6 +137,15 @@ func run() error {
 	)
 	if err != nil {
 		return err
+	}
+
+	// Record the active FSM apply sync mode so operators can see on the
+	// /metrics endpoint which durability posture this node is running in.
+	// The label is resolved per-pebbleStore from ELASTICKV_FSM_SYNC_MODE
+	// in NewPebbleStore; read it off the first constructed store (all
+	// shards share the same env and therefore the same label).
+	if label := fsmApplySyncModeLabelFromRuntimes(runtimes); label != "" {
+		metricsRegistry.SetFSMApplySyncMode(label)
 	}
 
 	cleanup := internalutil.CleanupStack{}
@@ -432,6 +437,35 @@ func raftMonitorRuntimes(runtimes []*raftGroupRuntime) []monitoring.RaftRuntime 
 		})
 	}
 	return out
+}
+
+// fsmApplySyncModeLabeler narrows an MVCCStore to those implementations
+// that can report the resolved ELASTICKV_FSM_SYNC_MODE label. The
+// pebble-backed store satisfies this today; alternate backends (none
+// yet) would either implement it or be skipped.
+type fsmApplySyncModeLabeler interface {
+	FSMApplySyncModeLabel() string
+}
+
+// fsmApplySyncModeLabelFromRuntimes returns the FSM apply sync-mode
+// label resolved by the first shard store that exposes it. All shards
+// on a node read the same ELASTICKV_FSM_SYNC_MODE env var at
+// construction time so the label is uniform across the runtimes;
+// returning the first one suffices. Returns "" when no runtime
+// exposes the accessor, in which case the caller skips emitting the
+// gauge to avoid publishing a misleading default.
+func fsmApplySyncModeLabelFromRuntimes(runtimes []*raftGroupRuntime) string {
+	for _, runtime := range runtimes {
+		if runtime == nil || runtime.store == nil {
+			continue
+		}
+		src, ok := runtime.store.(fsmApplySyncModeLabeler)
+		if !ok {
+			continue
+		}
+		return src.FSMApplySyncModeLabel()
+	}
+	return ""
 }
 
 // pebbleMonitorSources extracts the MVCC stores that expose
--- a/monitoring/pebble.go
+++ b/monitoring/pebble.go
@@ -51,6 +51,13 @@ type PebbleMetrics struct {
 	blockCacheCapacityBytes *prometheus.GaugeVec
 	blockCacheHitsTotal     *prometheus.CounterVec
 	blockCacheMissesTotal   *prometheus.CounterVec
+
+	// FSM apply sync mode. Resolved once from ELASTICKV_FSM_SYNC_MODE at
+	// process start (see store/lsm_store.go). The label-scoped gauge is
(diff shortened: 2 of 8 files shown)


bootjp:
Thanks for the reviews. Pushed 56af1db0 and 1236dae7 addressing the actionable items:

### 1. gosec G115 in `lsm_store_sync_mode_benchmark_test.go` (coderabbit / reviewdog)
Added an explicit `if i < 0` guard before the `uint64(i) * 2` conversion in `BenchmarkApplyMutations_SyncMode`. The counter is non-negative by construction, but the explicit guard is what the linter needs to clear G115.

### 2. `SetFSMApplySyncMode` unknown-label handling (coderabbit minor + gemini medium)
Previously, an unrecognised `activeLabel` was accepted verbatim: it stayed pinned at 1 across subsequent calls (only `"sync"` / `"nosync"` were explicitly zeroed), and the gauge could disagree with the actual `pebble.WriteOptions` the store uses, since `store.resolveFSMApplyWriteOpts` maps unknown `ELASTICKV_FSM_SYNC_MODE` values to `sync`.

Fixed by coercing any label outside `{"sync","nosync"}` to `"sync"` before setting — same fallback the store resolver uses, so the two paths agree. Docstring updated; new test `TestSetFSMApplySyncMode_UnknownLabelCoercesToSync` primes the gauge with `nosync`, calls `SetFSMApplySyncMode("batch")`, and asserts `sync=1 / nosync=0`.

### 3. Global-state test helper `setFSMApplyWriteOptsForTest` (gemini medium) — deferred
Leaving this as-is for this PR. The reviewer's own "References" section on [that comment](https://github.com/bootjp/elastickv/pull/592#discussion_r3130834197) cites the repo guideline: *"During large refactorings, limit the scope of a single pull request. Defer related but wide-reaching API changes to dedicated follow-up pull requests to manage complexity and avoid scope creep."* Changing `NewPebbleStore` to accept `WriteOptions` as a dependency is exactly that kind of wide-reaching API change — it ripples through every store construction site (main, engine bootstrap, tests, benchmarks) and is orthogonal to the durability opt-out this PR introduces. Happy to do it in a follow-up; not in this PR.

Local verification:
- `go test -race ./store/... ./monitoring/...` — pass
- `golangci-lint run ./store/... ./monitoring/...` — 0 issues

/gemini review
@codex review

bootjp:
Re-raising the same item. This PR intentionally defers the `NewPebbleStore(WriteOptions)` dependency-injection refactor to a follow-up.

Justification (from the previous round): gemini's own reference note says *"During large refactorings, limit the scope of a single pull request. Defer related but wide-reaching API changes to dedicated follow-up pull requests to manage complexity and avoid scope creep."* This PR's scope is the `ELASTICKV_FSM_SYNC_MODE` knob + the mutual-exclusivity gauge shape; changing `NewPebbleStore`'s signature would touch every call site across `store/`, `kv/`, `adapter/`, `main.go`, plus every test harness that constructs a store. That diff would dominate the diff, obscure the perf change, and invite merge conflicts with other in-flight work.

Concretely for test safety: `setFSMApplyWriteOptsForTest` already uses `t.Cleanup` to restore the prior value, and no test in this package uses `t.Parallel()`. The race window is zero under the current harness. If future parallelism lands, the helper can be swapped for struct-injected options in a dedicated refactor PR without changing the production semantics of this one.

Leaving item as-is. No new commits.

/gemini review
@codex review

bootjp:
-

bootjp:
Follow-up on the codex P2 correctness concern raised during this PR's review (`ELASTICKV_FSM_SYNC_MODE=nosync` bleeding into non-raft callers).

Addressed in #600: splits `ApplyMutations` / `DeletePrefixAt` (now unconditionally `pebble.Sync`) from new `ApplyMutationsRaft` / `DeletePrefixAtRaft` entry points that observe `s.fsmApplyWriteOpts`. `kv/fsm.go` is now the only caller of the `*Raft` variants, so the `nosync` knob cannot weaken durability for the `CatalogStore.Save` / `EnsureCatalogSnapshot` path (or any other direct caller) that has no raft-log replay backstop.

Tests + lint green. See #600 for the before/after table and new coverage.

PR #593, opened 2026-04-23
perf(raft): raise MaxInflightMsgs=1024, MaxSizePerMsg=4MB defaults

bootjp:
## Summary

- Expose etcd/raft's `MaxInflightMsgs` and `MaxSizePerMsg` as runtime-tunable env vars (`ELASTICKV_RAFT_MAX_INFLIGHT_MSGS`, `ELASTICKV_RAFT_MAX_SIZE_PER_MSG`), following the same pattern as `ELASTICKV_RAFT_SNAPSHOT_COUNT` (#589) and `ELASTICKV_RAFT_MAX_WAL_FILES` (#589).
- Raise production defaults: `MaxInflightMsgs` 256 -> **1024**, `MaxSizePerMsg` 1 MiB -> **4 MiB**.
- Align `main.go`'s `FactoryConfig` with the engine's internal `defaultMaxInflightMsg` (which was already 1024 for `stepCh` sizing per PR #560 but was being overridden by the caller-supplied 256).

## Why

**MaxSizePerMsg bump (the main win).** Under small-entry KV workloads (Redis-style, median entry ~500 B), a 1 MiB cap per MsgApp forces one MsgApp per ~2000 entries. At 4 MiB, each MsgApp amortises ~4x more entries. Concrete effects:

- Leader: fewer dispatcher goroutine wake-ups per committed byte -> lower CPU use per applied entry.
- Follower: fewer recv syscalls; less contention between apply loop and read path.
- Wire: no change in bytes-on-wire; only framing count.

**MaxInflightMsgs bump (insurance).** In-rack RTT is ~0.2 ms, so the replication-rate cap `MaxInflight * MaxSize / RTT` is already well above the NIC at the old defaults. The bump is for burst write phases where entries arrive faster than round-trip completion, and to align with the engine-internal 1024 already used for `stepCh` / `dispatchReportCh` sizing.

**Worst-case memory bound.** Per-peer buffered memory is `MaxInflight * MaxSize` = 1024 * 4 MiB = 4 GiB worst-case (was 256 * 1 MiB = 256 MiB). In practice MsgApp payloads are far smaller than the cap, so steady-state is a small fraction of this.

## Env vars

| Name | Type | Default | Min | Semantics |
|------|------|---------|-----|-----------|
| `ELASTICKV_RAFT_MAX_INFLIGHT_MSGS` | int | 1024 | 1 | Per-peer in-flight MsgApp cap. |
| `ELASTICKV_RAFT_MAX_SIZE_PER_MSG` | int (bytes) | 4194304 | 1024 | Per-MsgApp byte cap. |

Env var wins over caller-supplied `OpenConfig` value so operators can retune a running deployment without a rebuild. Invalid values (non-numeric, zero/negative, sub-1 KiB size) log a warning and fall back to the compiled-in default. Chosen values are logged at info level at Open.

## Before / after

| Knob | Before | After |
|------|--------|-------|
| `raft.Config.MaxInflightMsgs` | 256 (from main.go) | 1024 |
| `raft.Config.MaxSizePerMsg` | 1 MiB | 4 MiB |
| `raft.Config.MaxCommittedSizePerReady` | 1 MiB | 4 MiB |

Qualitative: fewer dispatch wake-ups per committed byte under small-entry workloads; deeper pipelining during write bursts. No expected change on wire bandwidth or correctness.

## Test plan

- [x] `go test ./internal/raftengine/etcd/... -count=1` passes
- [x] `go test ./... -count=1` passes across all packages
- [x] `golangci-lint run ./internal/raftengine/...` clean
- [x] Unit tests: env-unset -> defaults 1024 / 4 MiB propagate; valid override propagates to `raft.Config`; invalid values fall back to defaults with a warning
- [x] Existing `TestSnapshotEveryFromEnv_*` / `TestMaxWALFilesFromEnv_*` still pass (pattern mirrored)

## Surprising findings

- `defaultMaxInflightMsg` inside `internal/raftengine/etcd/engine.go` had already been bumped to 1024 (for `stepCh` sizing) and documented as such, but `main.go` was still passing 256 at the `FactoryConfig` boundary. The normalizer only fills defaults when the caller passes 0/<=0, so 256 leaked straight through to `raft.Config.MaxInflightMsgs`. This PR fixes that drift.
- `MaxInflightMsgs` / `MaxSizePerMsg` were not previously overridden anywhere else in the tree.

## Summary by CodeRabbit

* **New Features**
  * Runtime tuning via ELASTICKV_RAFT_MAX_INFLIGHT_MSGS and ELASTICKV_RAFT_MAX_SIZE_PER_MSG with clear precedence and safety clamping; invalid values log warnings and fall back to safe defaults.

* **Chores**
  * Increased default Raft limits (inflight msgs: 256→512; per-message size: 1MiB→2MiB) and added a minimum inbound channel capacity to improve stability.

* **Tests**
  * Added tests covering env override parsing, precedence, clamping, and channel sizing behavior.


bootjp:
Addressed the two review findings from gemini and codex on this PR.

### Commits
- bb6ff01f `fix(raft): apply MaxInflightMsg default on invalid env override` — addresses gemini medium x2. `maxInflightMsgFromEnv` / `maxSizePerMsgFromEnv` now return `(defaultMaxInflightMsg, true)` and `(defaultMaxSizePerMsg, true)` on invalid input instead of `(0, false)`, so the "using default" warning log is no longer a lie when the caller supplied a non-default value.
- 0546484d `fix(raft): size stepCh from resolved MaxInflightMsg, not compiled default` — addresses codex P1. `Open()` previously hard-coded `stepCh` and `dispatchReportCh` capacity at `defaultMaxInflightMsg` (1024). With `ELASTICKV_RAFT_MAX_INFLIGHT_MSGS=2048`, the raft layer honoured 2048 but the inbound channels stayed at 1024 and tripped `errStepQueueFull`, silently defeating the knob. Channels are now sized from `prepared.cfg.MaxInflightMsg` (post-`normalizeLimitConfig`) via a new `inboundChannelCap()` helper that clamps to `minInboundChannelCap=256` as a survivable floor.

### Channel capacity before/after
| Env var | Before (stepCh / dispatchReportCh cap) | After |
|---|---|---|
| unset | 1024 | 1024 |
| `ELASTICKV_RAFT_MAX_INFLIGHT_MSGS=2048` | 1024 (bug) | 2048 |
| `ELASTICKV_RAFT_MAX_INFLIGHT_MSGS=512` | 1024 | 512 |
| `ELASTICKV_RAFT_MAX_INFLIGHT_MSGS=128` | 1024 | 256 (floor) |

### Tests added / updated
- `TestMaxInflightMsgFromEnv_FallsBackOnInvalid` / `TestMaxSizePerMsgFromEnv_FallsBackOnInvalid` — updated to pin the new `(default, true)` contract.
- `TestNormalizeLimitConfig_InvalidEnvOverridesCaller` — new; invalid env + caller-supplied value resolves to the compiled-in default.
- `TestInboundChannelCap` — new; floor / passthrough unit test.
- `TestOpen_InboundChannelsHonourMaxInflightEnv` — new; opens a real engine with env=2048 and asserts `cap(engine.stepCh) == 2048` and `cap(engine.dispatchReportCh) == 2048`.
- `TestOpen_InboundChannelsDefaultCap` — new; env unset -> cap = `defaultMaxInflightMsg`.

`go test -race ./internal/raftengine/...` green. `golangci-lint run ./internal/raftengine/...` reports 0 issues.

/gemini review
@codex review

bootjp:
Addressed both gemini medium items.

### Item 1 — memory-footprint concern (option B)

Commit `a248131a` — lowered `defaultMaxInflightMsg` from 1024 back to 512, kept `defaultMaxSizePerMsg` at 4 MiB. Worst-case per-peer buffered product drops from 4 GiB to 2 GiB.

**Why B over A:** the existing code comment explicitly frames the 4 MiB bump as a batching win (amortise more entries per MsgApp, cut dispatcher wake-ups / recv syscalls on small-entry KV). Dropping `MaxSizePerMsg` to 2 MiB would undo that narrative. The 1024 bump was framed as a 4x belt-and-suspenders margin over the 256-slot bottleneck seen during the pebble-seek-storm incident, which PRs #560/#562/#563/#565 have since largely removed. 512 still gives a 2x safety margin over the pre-#529 default of 256 while preserving the MsgApp batching improvement. Operators who want deeper pipelining on wide-bandwidth LAN can still raise via `ELASTICKV_RAFT_MAX_INFLIGHT_MSGS` without a rebuild.

Also updated `main.go`'s `etcdMaxInflightMsg` caller-side constant so the engine default is actually exercised, and pinned the test expectation from 1024 to 512.

### Item 2 — clamping doc fix

Commit `ca6a4151` — `minMaxSizePerMsg` comment now says "fall back to the default rather than rejecting" instead of "clamp to this floor", matching the implementation in `maxSizePerMsgFromEnv` which returns `defaultMaxSizePerMsg` on sub-floor / malformed input.

### Verification

- `go test -race ./internal/raftengine/...` clean (17.4s)
- `golangci-lint run ./internal/raftengine/... .` clean

/gemini review

@codex review

bootjp:
Lowered `defaultMaxSizePerMsg` from 4 MiB to 2 MiB (commit 4edd02d5) to address the OOM concern raised on this PR.

**Worst-case memory math**

With `MaxInflightMsgs=512` (from the prior commit):

| | per-peer | 5-node leader (4 followers) |
|---|---|---|
| old (4 MiB) | 512 × 4 MiB = **2 GiB** | **8 GiB** |
| new (2 MiB) | 512 × 2 MiB = **1 GiB** | **4 GiB** |

This fits comfortably inside the 4–16 GiB RAM envelope of typical elastickv nodes, while still doubling etcd/raft's 1 MiB upstream default.

**Batching impact is marginal**

On the small-entry KV workload this PR targets (median ~500 B), 2 MiB / 500 B ≈ 4000 entries per MsgApp — well past the point where per-RPC amortisation saturates. 4 MiB bought essentially nothing on top of that.

**Changes**

- `internal/raftengine/etcd/engine.go`: `defaultMaxSizePerMsg = 2 << 20`, rationale block updated to reflect 512 × 2 MiB = 1 GiB worst case.
- `main.go`: `etcdMaxSizePerMsg = 2 << 20` aligned with the engine default, block comment updated.
- `internal/raftengine/etcd/engine_test.go`: `TestNormalizeLimitConfig_DefaultsWhenUnset` pinned to 2 MiB.

Note that `MaxCommittedSizePerReady` tracks `MaxSizePerMsg` (they share the same field in the `raft.Config` wiring at `engine.go` L613-614), so no separate change is needed.

**Verification**

- `go test -race ./internal/raftengine/... .` — pass
- `golangci-lint run ./internal/raftengine/...` — 0 issues

Per the user's explicit confirmation that `MaxInflightMsgs=512, MaxSizePerMsg=2 MiB` is the intended operating point.

/gemini review
@codex review

bootjp:
Addressed codex P3: `OpenConfig` doc comments in `internal/raftengine/etcd/engine.go` now reflect the real defaults (`MaxSizePerMsg`=2 MiB, `MaxInflightMsg`=512) instead of the stale 4 MiB / 1024 values.

Files corrected:
- `internal/raftengine/etcd/engine.go` (OpenConfig `MaxSizePerMsg` and `MaxInflightMsg` field comments)

/gemini review
@codex review

bootjp:
Addressed review feedback in d60cdcff:

**Codex P2 (x2) — env-override safety:**
- maxInflightMsgFromEnv now clamps values above maxMaxInflightMsg (8192) back to the compiled-in default with a warning. Prevents a fat-fingered ELASTICKV_RAFT_MAX_INFLIGHT_MSGS=1e8 from allocating multi-GB of channel memory at Open() and crashing startup.
- maxSizePerMsgFromEnv now clamps values above maxMaxSizePerMsg (64 MiB = internal.GRPCMaxMessageBytes) back to the default. Prevents Raft from emitting MsgApp payloads the gRPC transport physically cannot carry. TestMaxMaxSizePerMsg_MatchesTransportBudget pins the cap to the transport budget so a transport-budget change without a paired Raft-cap update fails CI.

**CodeRabbit minors (x2):**
- TestOpen_InboundChannels* now explicitly clears ELASTICKV_RAFT_MAX_SIZE_PER_MSG alongside the inflight var so an ambient shell env cannot influence the channel-cap assertions.
- Replaced the stale "silently cap at 1024" comment on the inboundChannelCap call with the current value (512 — today defaultMaxInflightMsg).

Boundary tests (at cap and cap+1) added for both overrides to catch off-by-one regressions.

/gemini review
@codex review

bootjp:
Addressed Codex P2 follow-up in 639ef081:

**Reserve gRPC envelope headroom on MaxSizePerMsg.** Right — etcd/raft MaxSizePerMsg caps the *entries-data* size per MsgApp, not the full raftpb.Message envelope. The envelope adds Term/Index/From/To/LogTerm/Commit plus per-entry framing, so a batch that exactly hits MaxSizePerMsg serializes to a frame slightly larger than MaxSizePerMsg. With the prior cap (maxMaxSizePerMsg == GRPCMaxMessageBytes exactly), a full-sized batch could tip the serialized frame past the 64 MiB transport limit and fail replication with ResourceExhausted even though env validation had accepted the value.

Now clamped at GRPCMaxMessageBytes - raftMessageEnvelopeHeadroom (1 MiB), so accepted overrides are strictly smaller than the transport budget. 1 MiB is an order of magnitude over the observed wire overhead for 8192-entry batches (tens of KiB), keeping the transport budget as the single source of truth while making the env override a safe bound under real batching.

The pinning test (renamed TestMaxMaxSizePerMsg_ReservesEnvelopeHeadroom) now asserts strict-less-than GRPCMaxMessageBytes so a future change that collapses the headroom trips CI.

/gemini review
@codex review

bootjp:
Round 3 in 8f72769e:

**Codex P2 — clamp caller-supplied OpenConfig values.** Right, the env-side guards in maxInflightMsgFromEnv / maxSizePerMsgFromEnv only catch values coming through the env vars; a programmatic caller passing OpenConfig{MaxInflightMsg: 100_000_000} skipped the clamp entirely and would have triggered multi-GB channel allocations at Open(). normalizeLimitConfig now applies maxMaxInflightMsg / maxMaxSizePerMsg to the resolved values too: exceeding either upper bound falls back to the compiled-in default with a warning. Env overrides still run first, so legitimate operator tuning is unaffected. Boundary tests (caller value at cap and cap+1) pin the new behaviour.

**Codex P2 — declined silently: "derive cap from worst-case protobuf overhead".** Re-checked against the etcd-raft source. limitSize already enforces MaxSizePerMsg using ent.Size() — the *encoded* protobuf size of each entry, including Type / Term / Index / Data-length framing — not just payload bytes. The MsgApp envelope above MaxSizePerMsg only adds the outer Message fields (Type/To/From/Term/LogTerm/Index/Commit/Reject/RejectHint, plus protobuf wrapping), which is bounded by hundreds of bytes — well within the 1 MiB headroom this PR already reserves. The "tiny entries add many MiB of framing" claim double-counts the per-entry overhead etcd-raft already includes in its limit, so no further adjustment is warranted.

/gemini review
@codex review

bootjp:
Round 4 in 11cb09bf — Codex was right, my prior justification was wrong.

**Corrected: halve MaxSizePerMsg cap to absorb tiny-entry framing.** Re-checked the etcd-raft source carefully: limitSize counts only Entry.Size() per entry, but raftpb.Message.Size() also adds per-entry framing inside the Entries repeated field — `1 byte (field tag) + sovRaft(len) + Entry payload`. For a minimal Entry (Term + Index + Type tags+varints, no Data, ~4 B), the framing is 2 B. So tiny-entry workloads encode at up to (4+2)/4 = 1.5x the entries-data budget on the wire. With the prior 1 MiB fixed headroom, a config of MaxSizePerMsg = GRPCMaxMessageBytes - 1 MiB on a tiny-entry workload could still serialize past the 64 MiB transport limit and fail with ResourceExhausted — exactly the failure mode the previous Codex P2 flagged.

Replaced the fixed-subtraction headroom with a divisor: maxMaxSizePerMsg = GRPCMaxMessageBytes / 2 = 32 MiB. This admits the worst-case 1.5x growth (32 MiB entries → 48 MiB on wire, well below 64 MiB transport) and is the minimum cap that keeps the env override safe under any entry-size distribution.

TestMaxMaxSizePerMsg_AbsorbsTinyEntryFraming added as a quantitative pin against the worst-case ratio so a future change that raises the divisor below 2 fails CI. Apologies for the prior decline-with-justification — the 1 MiB headroom analysis was incomplete because I had double-counted Entry.Size() with the per-Entry Message-level framing.

/gemini review
@codex review

bootjp:
@codex review

Verified the current HEAD already addresses the outstanding raft message limit findings:
- inbound stepCh and dispatchReportCh are sized from the resolved MaxInflightMsg via inboundChannelCap(prepared.cfg.MaxInflightMsg).
- invalid env overrides return the compiled-in defaults with ok=true, matching the warnings.
- MaxInflight and MaxSizePerMsg env/direct config values are capped before allocations / transport use, with gRPC headroom reserved.
- Open() tests clear both limit env vars and cover the overridden inbound channel capacity.

No additional code change was needed on top of the current branch head.

bootjp:
@codex review

PR #594, opened 2026-04-23
obs(redis): expose unsupported-command names via bounded metric

bootjp:
## Summary

- Add a new Prometheus counter `elastickv_redis_unsupported_commands_total{command}` that records the **actual** name of Redis commands the adapter rejects at the route-miss path (`adapter/redis.go`). Today those are all collapsed into `command="unknown"`, which leaves a large unnamed bar in Grafana's "Errors by Command".
- The new metric is **additive**: the existing `elastickv_redis_requests_total{command="unknown",outcome="error"}` and `elastickv_redis_errors_total{command="unknown"}` counters are unchanged, so existing dashboards and alerts keep working.

## Cardinality guarantee

- Hard cap of **32 distinct** command-name label values (`maxUnsupportedCommandLabels = 32`), guarded by a `sync.Mutex` on an in-memory seen-set in `RedisMetrics`.
- Once the cap is reached, any further novel name folds into `command="other"`. Already-seen names still record with their real label even after the cap is full.
- Label values are uppercased, trimmed, and **truncated to 64 chars** (`maxUnsupportedCommandLabelLen = 64`) before use, so a pathological client cannot pollute label values via protocol abuse (e.g. sending a huge first argument).

## Wiring

- `monitoring.RedisRequestReport` gains an `Unsupported bool` field. The adapter sets it to `true` only at the route-miss path via a new helper `observeRedisUnsupported`; all other call sites are unchanged.
- `RedisMetrics.ObserveRedisRequest` still runs the existing "unknown"-bucket path for unsupported commands, then *additionally* increments the new bounded counter when `Unsupported` is set.

## Test plan

- [x] `go test -race ./monitoring/...`
- [x] `go test -race ./adapter/...`
- [x] `golangci-lint run ./monitoring/... ./adapter/...`
- [x] Unsupported command records its real (uppercased) name.
- [x] Two different unsupported commands produce two label values.
- [x] The 33rd distinct name folds into `command="other"` and further overflow increments accumulate there.
- [x] An already-seen name (after cap reached) still records with its real label, not `"other"`.
- [x] Extreme-long (200-char) name is truncated to 64 chars.
- [x] Concurrent observers do not race (checked with `-race`).
- [x] Existing `NOTACOMMAND` test now asserts the new counter carries `NOTACOMMAND` alongside the unchanged `unknown` bucket.

## Summary by CodeRabbit

* **New Features**
  * Added dedicated monitoring for unsupported Redis commands with a new metric tracking command names and frequency, enabling clearer visibility into command rejections.

* **Tests**
  * Expanded test coverage for unsupported command tracking, including label cardinality management and concurrent operation validation.

### DIFF
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -606,7 +606,12 @@ func (r *RedisServer) Run() error {
 			if !ok {
 				r.traceCommandError(conn, name, cmd.Args[1:], "unsupported")
 				conn.WriteError("ERR unsupported command '" + string(cmd.Args[0]) + "'")
-				r.observeRedisUnsupported(name, time.Since(start))
+				// Pass the RAW command bytes (not the already-uppercased `name`)
+				// so that the unsupported-command observer can detect invalid
+				// UTF-8 before strings.ToUpper silently rewrites the bytes to
+				// the U+FFFD replacement character. See observeUnsupportedCommand
+				// in monitoring/redis.go.
+				r.observeRedisUnsupported(string(cmd.Args[0]), time.Since(start))
 				return
 			}
 
@@ -871,6 +876,11 @@ func (r *RedisServer) observeRedisError(command string, dur time.Duration) {
 // counters (which bucket the name into "unknown"), this flags the
 // report so the monitoring layer can record the real command name in
 // its bounded-cardinality unsupported-commands counter.
+//
+// IMPORTANT: `command` must be the RAW bytes the client sent (not an
+// already-uppercased value). The monitoring layer relies on seeing the
+// raw bytes to detect invalid UTF-8 before strings.ToUpper silently
+// replaces invalid bytes with the Unicode replacement character.
 func (r *RedisServer) observeRedisUnsupported(command string, dur time.Duration) {
 	if r.requestObserver == nil {
 		return
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -606,7 +606,7 @@ func (r *RedisServer) Run() error {
 			if !ok {
 				r.traceCommandError(conn, name, cmd.Args[1:], "unsupported")
 				conn.WriteError("ERR unsupported command '" + string(cmd.Args[0]) + "'")
-				r.observeRedisError(name, time.Since(start))
+				r.observeRedisUnsupported(name, time.Since(start))
 				return
 			}
 
@@ -863,6 +863,23 @@ func (r *RedisServer) observeRedisError(command string, dur time.Duration) {
 		Command:  command,
 		IsError:  true,
 		Duration: dur,
+	})
+}
+
+// observeRedisUnsupported records a command that was rejected because
+// the adapter has no route for it. In addition to the usual error
+// counters (which bucket the name into "unknown"), this flags the
+// report so the monitoring layer can record the real command name in
+// its bounded-cardinality unsupported-commands counter.
+func (r *RedisServer) observeRedisUnsupported(command string, dur time.Duration) {
+	if r.requestObserver == nil {
+		return
+	}
+	r.requestObserver.ObserveRedisRequest(monitoring.RedisRequestReport{
+		Command:     command,
+		IsError:     true,
+		Duration:    dur,
+		Unsupported: true,
 	})
 }
 
--- a/adapter/redis_metrics_test.go
+++ b/adapter/redis_metrics_test.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"testing"
 	"time"
+	"unicode/utf8"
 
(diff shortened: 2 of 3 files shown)


bootjp:
Addressed the CodeRabbit + codex P1 in 85398572.

### Bug (attack vector)
`RedisMetrics.observeUnsupportedCommand` truncated to 64 bytes with `name = name[:maxUnsupportedCommandLabelLen]`, a raw byte slice. If the name contained multibyte UTF-8 runes (e.g. 63 ASCII bytes + `é`), the cut could land mid-rune and produce invalid UTF-8. `prometheus.(*CounterVec).WithLabelValues` then panics — not errors — so a Redis client could crash the handling goroutine by sending an unsupported command name >64 bytes containing any non-ASCII byte.

### Fix
- Truncate on UTF-8 rune boundaries, preserving the 64-byte cap (semantics unchanged for ASCII, which is the expected path).
- Add an ingress guard: if the raw command bytes are not valid UTF-8, fold to a fixed `"invalid_utf8"` sentinel. The guard runs **before** `strings.ToUpper`, because `ToUpper` silently rewrites invalid bytes to U+FFFD and would mask the problem.

### Behaviour
- Before: 63 ASCII + `é...` crashed the goroutine via prometheus panic. Raw binary blob as a command name did the same.
- After: no panic in either case. Valid-UTF-8 inputs are truncated on a rune boundary with byte length <= 64; invalid-UTF-8 inputs are folded to `invalid_utf8`.

### Tests (`adapter/redis_metrics_test.go`)
- `TestRedisMetricsUnsupportedCommandTruncatesOnRuneBoundary` — 63 'A' + many 'é', asserts `NotPanics`, `len <= 64`, `utf8.ValidString`.
- `TestRedisMetricsUnsupportedCommandRejectsInvalidUTF8` — raw `0xff 0xfe 0xfd 0xc3 0x28`, asserts sentinel label emitted.
- `TestRedisMetricsUnsupportedCommandTruncatesLongName` — now also asserts `utf8.ValidString` on the label.

`go test -race ./monitoring/... ./adapter/...` passes for the redis-metrics tests (pre-existing `Test_grpc_transaction` flake on this branch is unrelated — same failure occurs with our change reverted). `golangci-lint run ./monitoring/... ./adapter/...` reports 0 issues.

/gemini review
@codex review

bootjp:
Addressed the five review items from the prior round.

**db4fa305 — fix(obs): preserve raw command bytes to reach invalid_utf8 sentinel**
Codex P2. The adapter's `Run()` loop was handing the already-`strings.ToUpper`'d `name` to `observeRedisUnsupported`. `strings.ToUpper` silently rewrites invalid UTF-8 bytes to U+FFFD, so by the time `observeUnsupportedCommand` ran `utf8.ValidString` the value was synthetically "valid" and the `invalid_utf8` sentinel path was unreachable — a hostile client sending binary garbage could burn through the 32 distinct-name slots with replacement-char strings. Now the adapter passes raw `string(cmd.Args[0])`; `ToUpper`/`TrimSpace` happen inside the observer, AFTER the UTF-8 check. Contract documented on both sides. Added `TestRedisMetricsUnsupportedCommandRawInvalidUTF8Bytes`.

**cc416338 — perf(obs): cap raw input, simplify rune truncation, add RWMutex fast path**
Three items bundled because they all touch `observeUnsupportedCommand`:

- **Gemini HIGH (security):** early `if len(raw) > 256 { raw = raw[:256] }` before any O(n) string op. Cutting mid-rune is fine; the downstream `utf8.ValidString` or rune-aware truncation handles it. Guards against multi-megabyte first-argument CPU abuse.
- **Gemini Medium (simplify):** rune-truncation loop now uses the range-loop byte index `i` and `utf8.RuneLen(r)` directly; the manual accumulator and negative-size guard are gone since the upstream UTF-8 validity check makes them dead code.
- **Gemini Medium (lock):** `unsupportedMu` is now `sync.RWMutex`. Already-admitted names take the read lock and return; only novel names acquire the write lock with a double-check to avoid the RUnlock→Lock race.

`unicode/utf8` was already imported (Item 5 no-op).

New tests: `TestRedisMetricsUnsupportedCommandRejectsPathologicalLength` (1 MiB input, no panic, label still 64-byte-capped valid UTF-8) and `TestRedisMetricsUnsupportedCommandRWMutexFastPath` (32 workers x 500 iters, mixed fast/slow paths, runs clean under `-race`).

Local verification:
- `go test -race -timeout 60s -run TestRedisMetrics ./adapter/...` — pass.
- `go test -race ./monitoring/...` — pass.
- `golangci-lint run ./monitoring/... ./adapter/...` — 0 issues.

/gemini review
@codex review

PR #596, opened 2026-04-23
test(redis): fix flaky TestRedis_LuaRPopLPushBullMQLikeLists

bootjp:
## Summary

- `TestRedis_LuaRPopLPushBullMQLikeLists` flaked on CI with `etcd raft engine is not leader` on the first `RPush` right after `createNode(t, 3)` returned. The freshly-elected raft leader briefly stepped down ("stepped down to follower since quorum is not active") before the test's first write reached it, so the write raced a re-election.
- Root cause is benign startup leader churn on slow GitHub Actions runners under `-race` — not a product bug. `waitForRaftReadiness` only confirms that a leader exists at one instant, which is not enough to guarantee stability for the very next write.
- Add small `rpushEventually` / `lpushEventually` helpers (backed by a shared `doEventually` that retries only on `strings.Contains(err, "not leader")` for 5s at 50ms intervals) and use them for the first two writes in the flaky test. `waitForRaftReadiness` is intentionally left untouched because other tests rely on its exact semantics.

Failing CI run: https://github.com/bootjp/elastickv/actions/runs/24842581631/job/72720509753

## Test plan

- [x] `go test -race -run TestRedis_LuaRPopLPushBullMQLikeLists ./adapter/ -count=5` — 5/5 pass locally.
- [x] `go test -race ./adapter/... -timeout 900s` — only unrelated pre-existing flake `Test_grpc_transaction` fires (nil-pointer panic in teardown); explicitly out of scope per the fix request.
- [x] `golangci-lint run ./adapter/...` — 0 issues.
- [ ] Re-run CI job linked above to confirm the flake is gone.

### DIFF
--- a/adapter/redis_lua_compat_test.go
+++ b/adapter/redis_lua_compat_test.go
@@ -121,8 +121,12 @@ func TestRedis_LuaRPopLPushBullMQLikeLists(t *testing.T) {
 	rdb := redis.NewClient(&redis.Options{Addr: nodes[0].redisAddress})
 	defer func() { _ = rdb.Close() }()
 
-	require.NoError(t, rdb.RPush(ctx, "bull:test:wait", "job-1", "job-2", "job-3").Err())
-	require.NoError(t, rdb.LPush(ctx, "bull:test:active", "job-0").Err())
+	// Raft leadership can briefly churn right after createNode returns — the
+	// freshly-elected leader can step down if the initial heartbeats miss
+	// quorum on a slow CI runner (see waitForRaftReadiness). Retry the first
+	// writes so they ride out that re-election instead of failing the test.
+	rpushEventually(t, ctx, rdb, "bull:test:wait", "job-1", "job-2", "job-3")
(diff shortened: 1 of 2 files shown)


bootjp:
Addressed two Gemini medium items in 4a81141b:

1. **`isTransientNotLeaderErr` case-sensitivity** — now lowercases `err.Error()` before matching so variants like `ERR Not Leader` from Redis protocol error bodies / upper layers are caught.
2. **`doEventually` swallowed error** — switched to `assert.Eventually` with a captured `lastErr`, then `require.NoError(lastErr)` after the loop. Timeouts now surface the real underlying error instead of "condition never met in 5s". Non-transient errors still fail fast because the closure returns `true` on them, exiting the loop immediately.

Verification:
- `go test -race -run TestRedis_LuaRPopLPushBullMQLikeLists ./adapter/ -count=5` — 5/5 pass.
- `golangci-lint run ./adapter/...` — 0 issues.

/gemini review
@codex review

PR #599, opened 2026-04-23
test(helper): event-driven leader-readiness with 300ms stability window

bootjp:
## Summary

Replace the polling-based `waitForRaftReadiness` in `adapter/test_util.go` with an event-driven helper that also enforces a 300ms leader-stability window.

### Why

PR #596's CI log showed this race:

```
15:04:52 became leader at term 2
15:04:53 stepped down to follower since quorum is not active
15:04:54 (first test write) -> "not leader" error
```

The previous `assert.Eventually(... n.raft.State() == raft.Leader ...)` helper returned between the `became leader` line and the subsequent `stepped down` line, so the test's first write raced into the brief non-leader window and failed.

### What changes

- Subscribe to `raft.LeaderCh()` (hashicorp/raft's buffered-1 bool channel that emits `true` on leader, `false` on step-down) instead of polling.
- Drain any pre-existing notifications before we start so we only observe this run's transitions.
- After we see the `true` event, require the leader to hold `raft.Leader` (and all followers to hold `raft.Follower` pointing at the leader's address) for a **300ms stability window**, sampled every ~25ms. If anything flips during the window, go back to waiting for the next `true` event.
- Bump the overall readiness timeout from 5s to 10s at the call site in `createNode` to give headroom for the stability window plus real election delay.
- `nodes[0]` is the bootstrap leader (it's the only `Voter` at bootstrap and is the one that calls `BootstrapCluster`), so observing only `nodes[0].raft.LeaderCh()` is correct.

Complementary to PR #596's mid-test `rpushEventually`/`lpushEventually` retry helper: this patch fixes the **setup-side** race so most tests don't need the retry at all. The in-test retry still covers genuine mid-test churn (e.g. leader transfer triggered by the workload), so both should stay.

## Test plan

- [x] `go test -race -run '^TestRedis_LuaRPopLPushBullMQLikeLists$' ./adapter/ -count=10` — 0 failures, 20.8s
- [x] `go test -race -run '^TestRedis_' ./adapter/ -count=1` — all pass, 23.8s
- [x] `go test -race ./adapter/... -timeout 900s` — all pass, 65.5s (no pre-existing flakes observed in this run)
- [x] `golangci-lint run ./adapter/...` — 0 issues

### DIFF
--- a/.github/workflows/jepsen-test-scheduled.yml
+++ b/.github/workflows/jepsen-test-scheduled.yml
@@ -101,6 +101,16 @@ jobs:
             --max-txn-length ${{ inputs.max-txn-length || '4' }} \
             --ports 63791,63792,63793 \
             --host 127.0.0.1
+      - name: Run Redis ZSet safety Jepsen workload against elastickv
+        working-directory: jepsen
+        timeout-minutes: 10
+        run: |
+          timeout 480 ~/lein run -m elastickv.redis-zset-safety-workload \
+            --time-limit ${{ inputs.time-limit || '150' }} \
+            --rate ${{ inputs.rate || '10' }} \
+            --concurrency ${{ inputs.concurrency || '8' }} \
+            --ports 63791,63792,63793 \
+            --host 127.0.0.1
       - name: Run DynamoDB Jepsen workload against elastickv
         working-directory: jepsen
         timeout-minutes: 10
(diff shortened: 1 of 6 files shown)


bootjp:
Closing: this branch was written against the hashicorp/raft stack (`LeaderCh()`, `raft.ServerAddress`) which has been removed from main as part of the etcd/raft migration. A literal merge is no longer possible. The stability-window concept is being re-implemented against the etcd/raft engine API in PR #602. Linking to this for history.

PR #601, opened 2026-04-23
feat(redis): implement HELLO command

bootjp:
## Summary

Client libraries (go-redis, ioredis, jedis, redis-py >= 4.x) routinely issue HELLO at connect time for RESP3 negotiation / capability probing. Before this change elastickv rejected HELLO as an unsupported command, which (a) inflated the elastickv_redis_unsupported_commands_total / unknown-bucket metrics every single time a client opened a connection, and (b) forced clients into awkward error fallback paths at handshake time.

This PR implements HELLO in the Redis adapter and wires per-connection state so client-name / id round-trip correctly.

## What is supported / rejected

| Subcommand | Behaviour |
| --- | --- |
| HELLO (no args) | Success. Returns flat RESP2 array: server=redis, version=7.0.0, proto=2, id=<conn-id>, mode=standalone, role=master/slave, modules=(empty array) |
| HELLO 2 | Success, same payload. |
| HELLO 3 | Rejected with NOPROTO unsupported protocol version. redcon is RESP2-only, so there is no path to RESP3. |
| HELLO foo / non-numeric protover | Rejected with NOPROTO. |
| HELLO 2 AUTH user pass | Rejected with NOPERM HELLO AUTH is not supported. elastickv has no Redis-layer auth; silently accepting credentials would be a security footgun. |
| HELLO 2 SETNAME myclient | Success; name persisted in connState.clientName. Subsequent CLIENT GETNAME on the same connection returns it. |

role flips between master / slave via coordinator.IsLeader(). version is hardcoded to 7.0.0 for max client compatibility (the code comment explains why elastickv's own version is not exposed here).

## Connection-id / client-name machinery

- Added connIDSeq atomic.Uint64 on RedisServer, allocated lazily via ensureConnID on first HELLO / CLIENT ID / CLIENT INFO.
- Added connID + clientName fields on connState.
- Extended CLIENT SETNAME / CLIENT GETNAME / CLIENT ID / CLIENT INFO to use the same slots, so HELLO SETNAME and CLIENT GETNAME are interoperable.
- Added HELLO to monitoring.redisCommandSet so observations record against the HELLO label instead of the unknown bucket.

## Test plan

- [x] go test -race ./adapter/... -run TestHello passes (11 new tests in adapter/redis_hello_test.go).
- [x] go test -race ./monitoring/... passes.
- [x] golangci-lint run ./adapter/... ./monitoring/... clean.
- [x] Verified Test_consistency_satisfy_write_after_read_sequence and Test_grpc_transaction flakes are pre-existing on main (unrelated to this change).
- [x] Existing TestRedisServer_Info_*, TestRedisMetrics_*, TestRedisMisskeyCompat, and TestRedisMetrics_MissingKeyNotCountedAsError still pass.

New test coverage:
- HELLO with no args returns expected map.
- HELLO 2 returns expected map.
- HELLO 3 returns NOPROTO error.
- HELLO foo (non-numeric) returns NOPROTO error.
- HELLO 2 AUTH user pass returns NOPERM error.
- HELLO 2 SETNAME myclient is visible via subsequent CLIENT GETNAME.
- role flips master/slave with coordinator.IsLeader().
- HELLO 2 SETNAME with missing value returns syntax error (no panic).
- connection id is stable within a connection and distinct across connections.
- Metrics side-effect: HELLO is recorded under the HELLO label, not unknown.

## Release Notes

* **New Features**
  * Added support for Redis `HELLO` command for protocol negotiation and server information retrieval.
  * Added `CLIENT ID` command returning unique, stable connection identifiers.
  * Added `CLIENT SETNAME` and `CLIENT GETNAME` commands for managing connection names.

* **Tests**
  * Added comprehensive test coverage for HELLO command functionality and option handling.


bootjp:
Addressed codex P2 on 525e993f in f3a1903d.

**Root cause**: `parseHelloArgs` mutated `state.clientName` immediately when it saw `SETNAME`, but subsequent tokens could still fail (an `AUTH` clause triggering `NOPERM`, or an unknown option triggering a syntax error). On error the connection name stayed overwritten, so a later `CLIENT GETNAME` observed a side effect of a command that returned an error reply. Real Redis commands are all-or-nothing; we now match that.

**Fix**: buffer the `SETNAME` value in `parseHelloArgs`, commit to `state.clientName` only after the full arg list validates successfully.

```go
var (
    pendingName    string
    pendingNameSet bool
)
for i := 1; i < len(args); {
    // ... may return error ...
    if opt.hasName {
        pendingName = opt.name
        pendingNameSet = true
    }
}
if pendingNameSet {
    state.clientName = pendingName
}
```

Also extracted `parseHelloOption` as a pure helper so `parseHelloArgs`'s cyclomatic complexity stays under the linter budget after the buffering change, and hoisted the arity literals `2` / `3` into `helloSetNameOptionArity` / `helloAuthOptionArity` constants to satisfy `mnd`.

**Regression tests added** (both previously would have let the mutation leak):
- `TestHello_SetNameNotPersistedOnLaterAuthError` — `HELLO 2 SETNAME new AUTH u p` → `NOPERM`, `clientName` stays `"prev"`.
- `TestHello_SetNameNotPersistedOnLaterUnknownOption` — `HELLO 2 SETNAME new BOGUS` → syntax error, `clientName` stays `"prev"`.

`go test -race -run '^TestHello_' ./adapter/` passes. `golangci-lint run ./adapter/...` clean.

/gemini review
@codex review

bootjp:
Addressed both coderabbit items on f3a1903d in 2b24bcf6.

**MAJOR — CLIENT subcommand arity.** `argsLen` only guarantees one subcommand token, so malformed calls silently succeeded:

| input | old behaviour | new behaviour |
|---|---|---|
| `CLIENT SETNAME` (no name) | `OK`, no mutation | `ERR wrong number of arguments for 'client\|setname' command` |
| `CLIENT GETNAME extra` | returns the name as if no operand | error |
| `CLIENT ID junk` | returns ID | error |
| `CLIENT INFO extra` | returns INFO | error |

Each subcommand now validates its exact arg count via a shared `checkClientArity` helper. Split each case into its own method (`clientSetName`, `clientGetName`, `clientID`, `clientInfo`) to keep the dispatch switch under the cyclomatic-complexity budget.

Regression tests added:
- `TestClient_SetName_RejectsWrongArity` — both missing name (too short) and extra args (too long) → error, `clientName` preserved
- `TestClient_GetName_RejectsWrongArity`
- `TestClient_ID_RejectsWrongArity`
- `TestClient_Info_RejectsWrongArity`

**MINOR — test-stub kind for `WriteString`.** `helloRecordingConn.WriteString` recorded its call as `kind="bulk"`, so a future regression that swapped `WriteBulkString` for `WriteString` in the HELLO reply path would pass silently. Changed to `kind="string"`; no existing assertions broke since every HELLO key/value in the current implementation uses `WriteBulk` / `WriteBulkString`.

`go test -race -run '^TestHello_|^TestClient_' ./adapter/` green. `golangci-lint run ./adapter/...` clean.

/gemini review
@codex review

Issue #551, opened 2026-04-19
lease-read: revisit CLOCK_MONOTONIC_RAW (deferred from #549)


bootjp:
### Background
Lease-read PR #549 uses Go's `time.Now()` for the lease wall-clock. This carries Linux `CLOCK_MONOTONIC` (rate-adjusted by NTP slew at up to 500 ppm = 0.05%). Worst-case slew error at the 700 ms lease window is ~0.35 ms — three orders of magnitude below the 300 ms `leaseSafetyMargin`.

### Decision (deferred)
TiKV uses `CLOCK_MONOTONIC_RAW` (NTP-slew-immune). Adopting the same in elastickv would require dropping the `time.Time` API on the lease path and calling `golang.org/x/sys/unix.ClockGettime(unix.CLOCK_MONOTONIC_RAW, &ts)`. Not worth the API churn at the current margin.

### Re-open trigger
- Safety margin tightened below ~5 ms.
- Jepsen-style clock-strobe testing reveals lease violations attributable to NTP rate adjustment.

References: PR #549, `docs/lease_read_design.md`.

bootjp:
PR #549 のレビュー検討で結論済み: NTP slew の最大誤差 (500 ppm × 700 ms = 0.35 ms) は margin (300 ms) に対して 1000 倍以上小さく、CLOCK_MONOTONIC_RAW 化は不要。

bootjp:
NTP的に十分なマージンであることは、安全性を保証することを意味しないので reopen

PR #604, opened 2026-04-23
kv/lease: adopt CLOCK_MONOTONIC_RAW for lease-read path (#551)

bootjp:
Go's time.Now() is backed by CLOCK_MONOTONIC, which POSIX lets NTP slew at up to 500 ppm. At the current 300 ms leaseSafetyMargin the worst-case slew error over a 700 ms lease window is ~0.35 ms — three orders of magnitude inside the margin — but lease safety should not rest on NTP being well-behaved. TiKV uses CLOCK_MONOTONIC_RAW for the same reason; this change adopts it in elastickv.

New internal/monoclock package wraps clock_gettime(CLOCK_MONOTONIC_RAW) via x/sys/unix on Linux/Darwin/FreeBSD, with a runtime-monotonic fallback on other platforms. leaseState, quorumAckTracker, the etcd engine's single-node ack, LeaseProvider.LastQuorumAck, and all coordinator LeaseRead / Dispatch / refreshLeaseAfterDispatch sites sample monoclock.Now() instead of time.Now(). Tests updated to match; quorum-ack tracker pins the raw-monotonic frame with a regression test. docs/lease_read_design.md §3.1-3.2 expanded with the rationale.

## Summary by CodeRabbit

* **Documentation**
  * Updated lease design documentation to reflect timing mechanism updates.

* **Internal Improvements**
  * Transitioned lease timing to use monotonic clock measurements instead of wall-clock time for improved reliability.
  * Introduced monotonic clock abstraction for intra-process timing operations.

* **Testing**
  * Added comprehensive test coverage for monotonic clock semantics and lease timing behavior.


bootjp:
@codex review

bootjp:
@codex review

bootjp:
@codex review

bootjp:
## Codex review items — addressed

- **Item 1 (P1, FreeBSD build break)** — commit `54a6f98b`
  Option B: dropped `freebsd` from `monoclock_unix.go`'s build tag (`//go:build linux || darwin`), FreeBSD now falls into the existing `monoclock_fallback.go` using Go's runtime monotonic clock. `GOOS=freebsd GOARCH=amd64 go build ./...` succeeds.

- **Item 2 (P1, clock failure must fail-closed)** — commit `03edaa16`
  Kept the `(ns int64)` return (no signature change) and extended the fail-closed contract to every consumer: `leaseState.valid` and `engineLeaseAckValid` both reject a zero `monoclock.Instant` explicitly, so callers sampling `monoclock.Zero` (clock_gettime denied under seccomp) fall through to `LinearizableRead`. Regression tests in `kv/lease_state_test.go` and `kv/lease_read_test.go` pin the invariant.

- **Item 3 (P3, lease rollback CAS needs unique token)** — commit `587717eb`
  Replaced the `atomic.Int64` + `atomic.Uint64` pair with `atomic.Pointer[leaseSlot]` where each `leaseSlot` wraps `{expiryNanos, gen}`. Each successful extend installs a freshly-allocated slot, so rollback is **pointer-identity CAS** on the extender's own slot — a concurrent extender that captured a post-invalidate generation and computed the same expiry (clock-granularity tie) cannot be clobbered because its slot is a distinct allocation. New `TestLeaseState_RollbackCASUsesPointerIdentity` deterministically simulates the clock-tie race.

### Checks

- `GOOS=freebsd GOARCH=amd64 go build ./...` — clean
- `go test -race ./kv/... ./internal/monoclock/... ./internal/raftengine/...` — all green
- `go vet ./...` — clean
- `golangci-lint run ./kv/... ./internal/monoclock/...` — 0 issues

/gemini review
@codex review

PR #608, opened 2026-04-23
fix(redis): validate CLIENT subcommand arity + distinguish string/bulk in test stub

bootjp:
Addresses coderabbit review on f3a1903d.

MAJOR: CLIENT subcommand arity (adapter/redis_compat_commands.go)

`argsLen` only guarantees CLIENT has one subcommand, so malformed requests silently succeeded:

  CLIENT SETNAME          (missing name)  -> OK, no mutation
  CLIENT GETNAME extra    (extra operand) -> name as if no operand
  CLIENT ID junk                          -> ID returned
  CLIENT INFO extra                       -> INFO returned

Each subcommand now validates its exact arg count via a shared checkClientArity helper, returning
`ERR wrong number of arguments for 'client|<sub>' command` on mismatch. Each case was also extracted into its own method so the switch body stays under the cyclomatic budget.

MINOR: distinguish string from bulk in helloRecordingConn test stub (adapter/redis_hello_test.go)

WriteString was recorded as kind="bulk", which meant a future regression swapping WriteBulkString for WriteString in the HELLO reply path would not fail any test. Now recorded as kind="string"; no existing assertions broke since all HELLO key/value writes use WriteBulk / WriteBulkString.

Regression tests added: CLIENT SETNAME / GETNAME / ID / INFO arity checks (+4 tests). go test ./adapter -race -run TestClient_|TestHello_ passes; golangci-lint clean.

### DIFF
--- a/adapter/redis_compat_commands.go
+++ b/adapter/redis_compat_commands.go
@@ -228,7 +228,7 @@ func checkClientArity(conn redcon.Conn, cmd redcon.Command, sub string, want int
 // CLIENT GETNAME gets the right answer without having to re-issue
 // CLIENT SETNAME.
 func clientSetName(conn redcon.Conn, cmd redcon.Command, state *connState) {
-	if checkClientArity(conn, cmd, "SETNAME", clientSetNameMinArgs) {
+	if checkClientArity(conn, cmd, "SETNAME", clientSetNameArgCount) {
 		return
 	}
 	state.clientName = string(cmd.Args[2])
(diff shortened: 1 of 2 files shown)


bootjp:
Addressed Copilot review (all 3 comments) in 7d283f62:

**CLIENT SETINFO arity enforcement.** SETINFO was still returning OK for any arity (including missing operands), the last remaining silent-success CLIENT subcommand — exactly what this PR is supposed to eliminate. Now routed through checkClientArity with clientSetInfoArgCount = 4 (CLIENT + SETINFO + attr + value), mirroring SETNAME. New TestClient_SetInfo_RejectsWrongArity pins all four arity shapes (2 / 3 / 4 / 5 args).

**Constant name clarity.** Renamed clientSetNameMinArgs to clientSetNameArgCount. The helper compares len(cmd.Args) == want, so the "Min" suffix misrepresented the semantics and invited a future refactor that could swap the helper for a >= check and silently re-introduce the wrong-arity bug. Added clientSetInfoArgCount as a separate constant for the same reason.

**Test consistency.** Replaced hardcoded []byte("CLIENT") in the arity tests with []byte(cmdClient) so the command name stays in one place.

/gemini review
@codex review

PR #597, opened 2026-04-23
obs(grafana): add Pebble internals + unsupported-commands panel; refactor Redis/Cluster dashboards

bootjp:
## Summary

- Add **Elastickv Pebble Internals** dashboard covering block cache hit rate/capacity, L0 pressure, compactions, memtables, FSM apply sync mode, and store write conflicts.
- Add **Unsupported Commands by Name** panel to the Redis dashboard (driven by PR #594 `elastickv_redis_unsupported_commands_total`) and annotate the legacy **Errors by Command** panel so operators know the `unknown` bucket is broken down by real name in the new panel.
- Fold the standalone Redis hot-path (PR #560) dashboard into a collapsed row at the bottom of the Redis summary dashboard, delete the standalone file, and retitle the summary to **Elastickv Redis** so there is a single Redis-facing dashboard.
- Rename the misnamed `elastickv-cluster-summary.json` (which was already a pure DynamoDB dashboard) to `elastickv-dynamodb.json` / **Elastickv DynamoDB** with `uid=elastickv-dynamodb`, and retitle `elastickv-cluster-overview.json` to **Elastickv Cluster** now that it is the sole cluster landing page.

## Before / After file list

| State | File |
| --- | --- |
| Added | `monitoring/grafana/dashboards/elastickv-pebble-internals.json` |
| Renamed | `elastickv-cluster-summary.json` to `elastickv-dynamodb.json` |
| Retitled | `elastickv-cluster-overview.json` (title now `Elastickv Cluster`) |
| Modified | `elastickv-redis-summary.json` (title `Elastickv Redis`; added Unsupported Commands panel and a collapsed Hot Path row) |
| Deleted | `monitoring/grafana/dashboards/elastickv-redis-hotpath.json` |
| Unchanged | `monitoring/grafana/dashboards/elastickv-raft-status.json` |

## What each dashboard now covers

- **Elastickv Cluster** - leader identity, cluster-wide latency/error posture, per-node Raft health, and an embedded Redis row.
- **Elastickv DynamoDB** - per-operation throughput/latency and per-table error/volume breakdown for the DynamoDB-compatible API.
- **Elastickv Raft Status** - unchanged: membership, leader changes, failed proposals, node state, index drift, backlog, leader contact.
- **Elastickv Redis** - per-command throughput/latency/errors, unsupported-commands breakdown, secondary-write and Lua diagnostics, and a collapsible `Hot Path (legacy PR #560)` row with the 9 GET-fast-path panels preserved verbatim.
[...]

PR #609, opened 2026-04-23
obs(grafana): pair stat + timeseries on Pebble internals dashboard

bootjp:
## Summary

Follow-up to #597 that landed before this improvement could ride along. The Pebble Internals dashboard rendered as flat stat-only rows (5 node values per metric), which is a bad fit for LSM signals that matter *as trends* (L0 build-up, compaction debt, memtable count approaching write-stall).

This PR restructures every gauge panel into a **stat (current value, colour-coded) + timeseries (per-node history with threshold line)** pair.

## Changes

Single file: `monitoring/grafana/dashboards/elastickv-pebble-internals.json`.

Final shape: 26 panels — 5 collapsible rows + 21 data panels (10 stat + 11 timeseries).

| Metric | Paired | Threshold line |
|---|---|---|
| `elastickv_pebble_l0_sublevels` | ✓ | red @ 20 (write-stall start) |
| `elastickv_pebble_l0_num_files` | ✓ | red @ 1000 |
| `elastickv_pebble_memtable_count` | ✓ | red @ 5 |
| `elastickv_pebble_memtable_size_bytes` | ✓ | — |
| `elastickv_pebble_memtable_zombie_count` | ✓ | — |
| `elastickv_pebble_compact_estimated_debt_bytes` | ✓ | — |
| `elastickv_pebble_compact_in_progress` | ✓ | — |
| `elastickv_pebble_block_cache_hit_rate` | ✓ (pre-existing) | — |

Rate counters (`compact_count_total`, `block_cache_hits_total`, `block_cache_misses_total`) left as timeseries-only — an instantaneous rate stat is meaningless.

`elastickv_store_write_conflict_total` is labelled by `key_prefix` not `node_id`, so per-node stat reduction doesn't apply. Single timeseries.

`elastickv_pebble_fsm_apply_sync_mode` stays a single stat (boolean mode indicator).

Layout: two 12-wide × 8-high panels per row, grouped under Block Cache / LSM Level State (collapsed) / Compaction / FSM Apply / Store Write Conflicts rows. Queries standardised on `max by (node_id)`.

## Test plan

- [x] JSON valid: `python3 -c "import json; json.load(open('monitoring/grafana/dashboards/elastickv-pebble-internals.json'))"`
- [ ] Import into Grafana, panels render without errors (operator verification)

/gemini review
@codex review


bootjp:
Addressed Codex P2 in db6115fe:

**Aggregate memtable bytes across groups per node.** elastickv_pebble_memtable_size_bytes is emitted per raft group, so max by (node_id) was picking only the largest group memtable on multi-group nodes and dropping the rest — two 200 MiB groups still displayed 200 MiB instead of 400 MiB, under-reporting real per-node memory pressure.

Both the "Memtable Size (current)" stat and the "Memtable Size Over Time" timeseries now use sum by (node_id). Panel descriptions updated to spell out the aggregation choice so a future reader can tell it is deliberate.

**Note on the other panels.** max by (node_id) is still correct for the L0 sublevels / num_files panels — sublevels is a per-group stall-risk signal where the worst group is what matters, and swapping to sum there would mask backpressure on the group actually being throttled. The memtable-bytes case is different because the metric is a bytes counter, so total node memory really is the sum across groups. Leaving the other panels untouched keeps this PR scoped to the flagged issue.

/gemini review
@codex review

bootjp:
Addressed Codex P2 (x2) follow-up in b16137e8:

**Compaction debt and compactions-in-progress now sum across groups per node.** Same per-group aggregation bug as the memtable panels fixed in db6115fe: both metrics carry a group label, and max by (node_id) dropped all but the busiest group. On multi-group nodes that meant two groups each holding 400 MiB of debt rendered as 400 MiB instead of 800 MiB, and two groups each running two compactions rendered as 2 instead of 4 — hiding per-node compaction pressure the panels were specifically meant to surface.

Switched both stat + timeseries panels for compact_estimated_debt_bytes and compact_in_progress to sum by (node_id). Descriptions updated to spell out the aggregation and the (per-group max × group-count) ceiling an operator should expect.

Still leaving max by (node_id) on the L0 sublevels / num_files panels — those are per-group stall-risk signals where the worst group is what matters, and summing there would dilute exactly the backpressure signal that drives write-stall investigation.

/gemini review
@codex review

PR #612, opened 2026-04-24
feat(memwatch): graceful shutdown on memory-pressure threshold

bootjp:
## Summary
- Add `internal/memwatch`: opt-in watchdog that polls `runtime.MemStats.HeapInuse` and fires the root-ctx cancel when a configurable threshold is crossed, causing orderly shutdown with exit code 2 instead of kernel OOM-SIGKILL.
- Wire into `main.go` via two env vars (default OFF): `ELASTICKV_MEMORY_SHUTDOWN_THRESHOLD_MB`, `ELASTICKV_MEMORY_SHUTDOWN_POLL_INTERVAL`.
- Motivation: 2026-04-24 incident — all 4 live nodes SIGKILLed 22-169 times in 24h under traffic spike. Each kill risked WAL-tail truncation and reset lease/raft state, producing election storms + p99 6-8s.

## Design notes
- `HeapInuse` over RSS (no syscall per poll) and over `Sys` / `Alloc` / `NumGC` (include freed-but-not-returned memory or are monotonic counters). Rationale in package doc.
- Single-shot (CAS-guarded): `OnExceed` fires exactly once even if allocation stays high across polls.
- No duplicate shutdown path — hooks into the existing root-ctx `cancel()`; no direct raft/adapter manipulation.
- Zero-cost when disabled: no goroutine, no ticker if threshold == 0.
- Exit code 2 distinguishes memory-pressure exit from crash (1) and normal shutdown (0).

## Operational use
Set on each node alongside `GOMEMLIMIT`:

```
-e GOMEMLIMIT=1800MiB
-e ELASTICKV_MEMORY_SHUTDOWN_THRESHOLD_MB=1900
```

`GOMEMLIMIT` makes Go GC aggressively; memwatch catches the case where even GC can't keep up (e.g., non-heap growth like mmap'd files or goroutine stacks) and exits cleanly before the kernel kills the process.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/memwatch/...` — 4 tests: fires-once, does-not-fire, context-cancel-stops, zero-threshold-disabled
- [ ] Manual: on a test node, set `ELASTICKV_MEMORY_SHUTDOWN_THRESHOLD_MB=200`, load keys until heap crosses 200MB, observe log line + exit code 2
- [ ] Operational: combine with `GOMEMLIMIT` on all production nodes; verify that a traffic spike triggers orderly shutdown (no WAL corruption) rather than SIGKILL

## Follow-ups (tracked, not in this PR)
- Prometheus metric for memwatch-fired counter (per `docs/design/2026_04_24_proposed_resilience_roadmap.md`)
- Compose with admission control (item 6 of resilience roadmap) via a shared "draining" flag so new work is rejected before shutdown

## Release Notes

* **New Features**
  * Added an optional memory watchdog that monitors heap usage and initiates graceful shutdown when a configured threshold is exceeded. Configurable via environment variables and disabled by default.

### DIFF
--- a/internal/memwatch/memwatch.go
+++ b/internal/memwatch/memwatch.go
@@ -1,0 +1,195 @@
+// Package memwatch provides a user-space memory watchdog that triggers
+// an orderly shutdown before the kernel OOM-killer sends SIGKILL.
+//
+// Motivation
+//
+// elastickv runs in memory-constrained containers (e.g. 3GB RAM VMs). Go's
+// runtime is unaware of the container/host memory limit, and even with
+// GOMEMLIMIT set the process can still lose the race against the kernel
+// OOM-killer under sustained memtable/goroutine growth. A SIGKILL leaves
+// the Raft WAL potentially truncated mid-operation; a cooperative SIGTERM
+// path lets the node sync the WAL and stop raft cleanly, avoiding the
+// election storms and lease loss that follow crash-restarts.
+//
+// The watcher polls runtime.ReadMemStats at a fixed cadence. When
+// HeapInuse crosses the configured threshold it invokes OnExceed once
+// and exits. The watcher never calls os.Exit or sends signals itself;
+// callers wire OnExceed to the existing shutdown path (typically a
+// root context.CancelFunc).
+//
+// Wiring in elastickv (see main.go):
+//
+//	ctx, cancel := context.WithCancel(context.Background())
+//	// ... build runtimes, servers, errgroup ...
+//	w := memwatch.New(memwatch.Config{
+//	    ThresholdBytes: threshold,
+//	    PollInterval:   pollInterval,
+//	    OnExceed: func() {
+//	        memoryPressureExit.Store(true) // flips exit code to 2
+//	        cancel()                       // fires the same shutdown path
+//	    }, //                                  SIGTERM would use.
+//	})
+//	eg.Go(func() error { w.Start(runCtx); return nil })
+//
+// Metric choice
+//
+// We read runtime.MemStats.HeapInuse. It is the closest Go-runtime-visible
+// proxy for "how close are we to OOM" without a syscall per poll. RSS from
+// /proc/self/status is more accurate but requires a read syscall on every
+// poll; at the 1s cadence this watchdog runs that accuracy isn't worth the
+// cost. We deliberately do NOT use MemStats.Sys, NumGC or Alloc: Sys and
+// NumGC include memory the runtime has already released back to the OS (or
+// are monotonic counters) and Alloc counts only currently-live heap objects,
+// missing the span-level overhead that the OOM-killer actually sees.
+package memwatch
+
+import (
+	"context"
+	"log/slog"
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// DefaultPollInterval is the polling cadence used when Config.PollInterval
+// is zero. One second is frequent enough to catch fast-growing memtables
+// before the kernel kills the process, but infrequent enough that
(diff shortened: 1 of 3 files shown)


bootjp:
Addressed review feedback in ab4bdab4:
- main.go: moved `memoryPressureExit` check before `log.Fatalf` (Gemini HIGH / Codex P2)
- main.go: extracted `bytesPerMiB` constant (golangci mnd)
- memwatch.go: replaced package-level mutex with per-Watcher `sync.Once` (Gemini MEDIUM)
- memwatch.go: enforced Start single-shot via `started.CompareAndSwap` (Gemini MEDIUM)

runtime/metrics migration deferred — rationale in commit message.

/gemini review
@codex review

bootjp:
Addressed round-2 feedback in 9034431b:
- golangci gci (main.go:18)
- CodeRabbit Major overflow (main.go:164): guard mb*bytesPerMiB wrap; disable watcher on overflow
- CodeRabbit Minor effective interval (main.go:611): log the resolved PollInterval, not the zero pre-default
- Gemini Medium swallowed error (main.go:136): log residual run() error before os.Exit(2)

Deferred: CodeRabbit Minor `t.Fatalf` from goroutine (follow-up cleanup), Gemini Medium runtime/metrics migration (rationale in prior commit).

/gemini review
@codex review

bootjp:
Round 3 in 807520dd:
- Gemini Medium: enforced 10ms floor on PollInterval (`MinPollInterval`) so misconfig can't turn into STW-storm.
- Gemini Medium: filter `context.Canceled` from the post-shutdown-error log — it's the expected cancel propagation, not a failure.

Deferred: runtime/metrics migration — rationale unchanged across rounds.

/gemini review
@codex review

bootjp:
Round 4 — no code change.

The sole new comment is Gemini Medium on `internal/memwatch/memwatch.go:176` asking to migrate from `runtime.ReadMemStats` to `runtime/metrics`. Declining again; see prior rounds (ab4bdab4, 9034431b, 807520dd). Rationale is unchanged: migration is out of scope for this PR, tracked as a follow-up.

No other reviewer feedback this round.

/gemini review
@codex review

bootjp:
Migrated in f203b1a8.

You were right to keep flagging it. Dropped the prior "STW is negligible at 1s" rationale — the MinPollInterval floor added in round 3 (10ms) is where the pause would actually register, and runtime/metrics has been a clean alternative since Go 1.16.

Details:
- Threshold now compared against `/memory/classes/heap/objects:bytes` + `/memory/classes/heap/unused:bytes` — the runtime/metrics equivalent of `MemStats.HeapInuse` (allocated heap spans including reusable free slots; excludes OS-released memory).
- `[]metrics.Sample` allocated once in `New`, reused every poll — no per-poll allocation, no name resolution on the hot path.
- Warning log still structured: `heap_inuse_bytes`, `heap_objects_bytes`, `heap_released_bytes`, `next_gc_bytes`. Dropped `heap_sys_bytes` since Sys includes released memory and isn't useful for the "how close to OOM" question.
- `sampleUint64` defensively returns 0 for `KindBad` so a future runtime retiring a metric name fails closed (no false-positive trigger) rather than panicking.

Tests unchanged and still pass; lint clean.

/gemini review
@codex review

bootjp:
Addressed in 82bc1a54:

**Codex P1 (immediate threshold check):** `Start` now samples once before the first tick. If the process is already above the threshold at Start — crash-loop restart, large startup allocations — the watcher fires immediately rather than waiting up to `PollInterval` (1 s default) for the first tick. Closes the window where the kernel OOM-killer could act before the watcher had ever observed pressure.

/gemini review
@codex review

PR #613, opened 2026-04-24
fix(raft): auto-repair truncated WAL tail on startup

bootjp:
## Summary
When the kernel OOM-killer SIGKILLs the app mid-WAL-write, the last preallocated WAL segment is left with a torn trailing record. `wal.ReadAll` returns `io.ErrUnexpectedEOF` and our wrapper propagated it directly — the process failed to start and entered a crash-loop until an operator manually quarantined the bad segment.

This PR invokes `wal.Repair` on startup when `ReadAll` returns `io.ErrUnexpectedEOF`, matching the recovery path that etcd's own server uses. The partial tail record is truncated, the node resumes from the last fully-committed index, and catches up from the leader via normal raft replication.

## Motivation
2026-04-24 incident: a traffic spike caused kernel OOM-SIGKILL on all 4 live nodes 22-169 times in 24h. One kill landed inside a WAL record, making the node un-startable for 5+ hours until manual quarantine. With this fix the same state would self-repair on the next restart with no operator involvement.

## Scope of repair
- **Triggers only on `io.ErrUnexpectedEOF`** — the torn-trailing-record signature.
- **CRC mismatches and other errors propagate unchanged.** Those are genuine corruption, not in-flight-write artifacts; auto-truncating them would silently discard valid state.
- **One repair attempt only.** If `wal.Repair` returns false, or the second `ReadAll` still fails, the error surfaces wrapped as "WAL unrepairable".
- **One log line emitted** on repair: `"WAL tail truncated, repairing"` with dir + original error.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/raftengine/etcd/...` — full suite passes (7.3s)
- [x] `TestLoadWalStateRepairsTruncatedTail` — seeds a WAL, chops the tail inside framing, asserts re-open succeeds and pre-truncation entries survive
- [x] `TestLoadWalStateUnrepairableCRCMismatchReturnsError` — flips a byte inside a record, asserts the error propagates (repair cannot mask real corruption)
- [x] `TestOpenAndReadWALSucceedsWithoutRepair` — happy-path sanity check
- [ ] Manual: reproduce the incident — kill a node with `kill -9` mid-write (e.g., under heavy `PROPOSE` load), observe auto-repair on restart, confirm catch-up from leader

## Follow-ups (tracked, not in this PR)
- Prometheus counter for WAL-repair-invocations (operability signal — frequent repair = underlying reliability problem)
- Companion PR: memwatch-based graceful shutdown (#612) avoids triggering this path in the first place

### DIFF
--- a/internal/raftengine/etcd/wal_store.go
+++ b/internal/raftengine/etcd/wal_store.go
@@ -107,6 +107,28 @@ func bootstrapWalState(logger *zap.Logger, walDir, snapDir, fsmSnapDir string, f
 }
 
 func loadWalState(logger *zap.Logger, walDir, snapDir, fsmSnapDir string, fsm StateMachine) (*diskState, error) {
+	// The outer repair retry covers BOTH loadPersistedSnapshot (which
+	// scans WAL via wal.ValidSnapshotEntries) and openAndReadWAL's
+	// ReadAll. Either can surface io.ErrUnexpectedEOF when the kernel
+	// OOM-killer SIGKILLed the process mid-WAL-write, leaving the last
+	// preallocated segment with a partial trailing record. wal.Repair
+	// truncates that partial record once and is idempotent, so the
+	// retry is bounded.
+	ds, err := tryLoadWalState(logger, walDir, snapDir, fsmSnapDir, fsm)
+	if err == nil || !errors.Is(err, io.ErrUnexpectedEOF) {
+		return ds, err
+	}
+	logger.Warn("WAL tail truncated, repairing",
+		zap.String("dir", walDir),
+		zap.Error(err),
+	)
+	if !wal.Repair(logger, walDir) {
+		return nil, errors.Wrap(err, "WAL unrepairable")
+	}
+	return tryLoadWalState(logger, walDir, snapDir, fsmSnapDir, fsm)
+}
+
+func tryLoadWalState(logger *zap.Logger, walDir, snapDir, fsmSnapDir string, fsm StateMachine) (*diskState, error) {
 	snapshotter := snap.New(logger, snapDir)
 	snapshot, err := loadPersistedSnapshot(logger, walDir, snapshotter)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bootjp:
Addressed review feedback in f3f922d7:
- HIGH: hoisted repair to `loadWalState` level so it also catches `io.ErrUnexpectedEOF` from `loadPersistedSnapshot` → `wal.ValidSnapshotEntries` (previous version only retried around `openAndReadWAL`, leaving a window where a torn snapshot-record aborted startup before repair ran).
- MEDIUM: `w.Close()` errors now logged instead of silently dropped in both error branches.
- MEDIUM: replaced `Size()-(64KB+16)` with `truncateInsideLastRecord` which scans the preallocated tail backwards past the zero padding and truncates 5 bytes *inside* real framing. The previous truncation likely landed in zero padding → test was passing for the wrong reason (no `ErrUnexpectedEOF`). With the new helper the repair path is actually exercised.

/gemini review
@codex review

bootjp:
Addressed in 18007b3a:
- Gemini Medium: second tryLoadWalState failure is now wrapped as "WAL unrepairable after repair", matching the PR description.

/gemini review
@codex review

bootjp:
Round 3 in d3fa0388:
- Codex P2 + Gemini Medium: scoped wal.Repair to WAL-only reads. The round-2 refactor wrapped tryLoadWalState which also ran restoreSnapshotState; a truncated .fsm payload returning io.ErrUnexpectedEOF would have triggered wal.Repair on an otherwise-healthy WAL. Split into two per-path helpers (`loadPersistedSnapshotWithRepair`, `openAndReadWALWithRepair`); FSM restore runs between them and its errors propagate without invoking repair.

/gemini review
@codex review

bootjp:
Round 4 in 261b47b1:

- Gemini Medium (wal_store.go:132 and :160): shared the snapshotter across `loadWalState` and `loadPersistedSnapshotWithRepair` — the helper now takes `*snap.Snapshotter` from the caller instead of instantiating its own. One allocation per open instead of two.

Build + `go test ./internal/raftengine/etcd/...` pass; golangci-lint clean on the changed package (the pre-existing goconst warning on `wal_purge.go:140` is unrelated to this PR).

No other actionable feedback this round.

/gemini review
@codex review

PR #614, opened 2026-04-24
test(adapter): widen raft election budget + retry gRPC on leader churn

bootjp:
## Background

PR #602 added a `waitForStableLeader` step to absorb setup-side leader churn for the redis tests, but CI kept failing on `Test_consistency_satisfy_write_after_read_sequence`. Reproduced locally with `go test -race ./adapter/ -count=1`:

```
grpc_test.go:192: Put RPC failed: etcd raft engine is not leader
grpc_test.go:195: Put RPC failed: leader not found
grpc_test.go:198: Get RPC failed: leader not found
panic: runtime error: invalid memory address or nil pointer dereference
  [signal SIGSEGV] github.com/bootjp/elastickv/adapter.Test_consistency_satisfy_write_after_read_sequence
      adapter/grpc_test.go:200
```

The failing test drives 9,999 `Put / Put / Get` RPCs against the leader in a loop (`Test_grpc_transaction` and `Test_consistency_satisfy_write_after_read_for_parallel` have the same shape with `Put / Get / Delete / Get` and 1,000× parallel). If the leader loses quorum **mid-run** the RPC returns `"etcd raft engine is not leader"` or `"leader not found"` and the test crashes -- and because `assert.NoError` is non-fatal, a nil response is then dereferenced at `resp.Value` → SIGSEGV.

## Root cause

`adapter/test_util.go` was configured with:

```go
testEngineTickInterval  = 10 * time.Millisecond
testEngineHeartbeatTick = 1   // 10 ms heartbeat
testEngineElectionTick  = 10  // 100 ms election timeout
```

etcd/raft's `CheckQuorum` only holds ~1 election-timeout worth of heartbeat-response history, so **any** goroutine-scheduler pause > 100 ms on a `-race` GHA runner makes the leader conclude quorum is lost and step down ("quorum is not active"). Under three parallel 9,999-iteration tests that is very easy to hit. PR #602's setup-side stability window can't help here -- the flip happens mid-test.

## Fix

1. **Widen the test election budget** to the 10× etcd/raft-recommended ratio with a 500 ms election timeout:
   ```go
   testEngineHeartbeatTick = 5   // 50 ms
   testEngineElectionTick  = 50  // 500-1000 ms randomised
   ```
   Absorbs typical `-race` scheduler jitter without slowing tests noticeably. As a bonus, `LeaseDuration = 500 - 300 = 200 ms` (was 0), so the lease-read fast path is actually exercised instead of always falling through to ReadIndex.

2. **Extend `isTransientNotLeaderErr`** to also match `"leader not found"` (`ErrLeaderNotFound`), which the coordinator emits during the re-election window -- distinct from `"not leader"` that PR #596's matcher already handled.

3. **Add gRPC `*Eventually` wrappers** (`rawPutEventually`, `rawGetEventually`, `txnPutEventually`, `txnGetEventually`, `txnDeleteEventually`) that reuse `doEventually`, mirroring the `rpushEventually` / `lpushEventually` pattern from #596 for the gRPC paths called out as "out of scope" in #602.

4. **Rewrite the three consistency tests** (`Test_consistency_satisfy_write_after_read_sequence`, `Test_consistency_satisfy_write_after_read_for_parallel`, `Test_grpc_transaction`) to use the `*Eventually` helpers. This also fixes the nil-pointer panic -- the helpers only return non-nil responses.

## Test plan

- [x] `go test -race -run '^(Test_consistency_satisfy_write_after_read_sequence|Test_grpc_transaction|Test_consistency_satisfy_write_after_read_for_parallel|TestRedis_LuaRPopLPushBullMQLikeLists)$' ./adapter/ -count=1` — 3/3 green consecutive runs (~210 s each).
- [x] `go test -race ./adapter/ -count=1` (full adapter package) — ok, 252 s.
- [x] `go vet ./...`, `go build ./...` — clean.
- [ ] Re-run CI `test (ubuntu-latest)` (only blocker was local golangci-lint vs. Go 1.26 mismatch).

Complements #602 (#602 fixed setup-side churn, this PR fixes the mid-test churn that #602 called out as a separate follow-up).

---
_Generated by [Claude Code](https://claude.ai/code/session_01J5wXeApZcnXCxu2zYsz9MZ)_


bootjp:
@codex review

PR #619, opened 2026-04-24
docs(design): propose workload-class isolation after 2026-04-24 XREAD starvation

bootjp:
## Summary
Design doc (only — no code in this PR) for a four-layer workload-isolation model, prompted by the 2026-04-24 incident's afternoon phase.

**Problem:** Today, one client host with 37 connections running a tight XREAD loop consumed 14 CPU cores on the leader via `loadStreamAt → unmarshalStreamValue → proto.Unmarshal` (81% of CPU per pprof). Raft goroutines couldn't get CPU → step_queue_full = 75,692 on the leader (vs 0-119 on followers) → Raft commit p99 jumped to 6-10s, Lua p99 stuck at 6-8s. Follower replication was healthy (applied-index within 34 of leader); the damage was entirely CPU-scheduling on the leader.

**Gap:** elastickv has no explicit workload-class isolation. Go's scheduler treats every goroutine equally; a single heavy command path can starve unrelated paths (raft, lease, Lua, GET/SET).

## Four-layer defense model
- **Layer 1 — heavy-command worker pool**: gate XREAD / KEYS / SCAN / Lua onto a bounded pool (~`2 × GOMAXPROCS`); reply `-BUSY` when full. Cheap commands keep their own fast path.
- **Layer 2 — locked OS threads for raft**: `runtime.LockOSThread()` on the Ready loop + dispatcher lanes so the Go scheduler can't starve them. **Not v1** — only if measurement after Layer 1 + 4 still shows `step_queue_full > 0`.
- **Layer 3 — per-client admission control**: per-peer-IP connection cap (default 8). Extends, doesn't replace, roadmap item 6's global in-flight semaphore.
- **Layer 4 — XREAD O(N) → O(new)**: entry-per-key layout (`!redis|stream|<key>|entry|<id>`) with range-scan, dual-read migration fallback, legacy-removal gated on `elastickv_stream_legacy_format_reads_total == 0`. Hashes/sets/zsets share the same one-blob pattern and are called out as follow-up.

## Recommended sequencing
Layer 4 (correctness bug, concentrated change) → Layer 1 (generic defense for next unknown hotspot) → Layer 3 (reconcile with roadmap item 6) → Layer 2 (only if forced by measurement).

## Relationship to other in-flight work
- Complements (does not replace) `docs/design/2026_04_24_proposed_resilience_roadmap.md` item 6 (admission control). This doc's Layer 3 focuses on per-client fairness; the roadmap's item 6 is global in-flight capping. Both are needed.
- Consistent with memwatch (#612): Layer 3 admission threshold should fire **before** memwatch's shutdown threshold — flagged as an open question in the doc.
- Assumes WAL auto-repair (#613), GOMEMLIMIT defaults (#617) are landed so the cluster survives long enough to matter.

## Open questions called out in the doc
- Static vs dynamic command classification (Layer 1)
- `-BUSY` backoff semantics — how do we avoid client retry spinning becoming the new hot loop?
- Number of locked OS threads on variable-core hosts (Layer 2)
- Stream migration soak window before removing legacy-format fallback (Layer 4, currently 30 days, arbitrary)

## Deliverable
`docs/design/2026_04_24_proposed_workload_isolation.md` — 446 lines, dated-prefix / `**Status: Proposed**` convention matching the rest of `docs/design/`. No code.

## Test plan
- [x] File paths and function references in the doc spot-checked against `origin/main`
- [x] Cross-references to `2026_04_24_proposed_resilience_roadmap.md` reconciled (complements, doesn't duplicate)
- [ ] Design review — decide on the open questions before implementing Layer 4 (which blocks Layer 1 on XREAD specifically)

### DIFF
--- a/docs/design/2026_04_24_proposed_workload_isolation.md
+++ b/docs/design/2026_04_24_proposed_workload_isolation.md
@@ -78,16 +78,21 @@ commands stay on the accept-goroutine path. Pool full → reply
 `-BUSY server overloaded` and return. Redis clients already treat
 `-BUSY` as retryable; reusing it means no client-library changes.
 
-Static v1 classification:
+Static v1 classification (name-based only — no argument inspection,
+so the dispatcher can gate on the command byte without allocating):
 
 - **Pool-gated:** `XREAD`, `XRANGE`, `XREVRANGE`, `KEYS`, `SCAN`,
-  `HGETALL`, `SMEMBERS`, full-range `ZRANGE`/`ZRANGEBYSCORE`,
-  `EVAL`/`EVALSHA`, `FCALL`/`FCALL_RO`, the `*SCAN` family.
+  `HGETALL`, `HVALS`, `HKEYS`, `SMEMBERS`, `SUNION`, `SINTER`,
+  `ZRANGE`, `ZRANGEBYSCORE`, `ZRANGEBYLEX`, `EVAL`/`EVALSHA`,
+  `FCALL`/`FCALL_RO`, the `*SCAN` family.
 - **Ungated:** `GET`, `SET`, `DEL`, `EXISTS`, `INCR`, `EXPIRE`, `TTL`,
   `HGET`, `HSET`, `LPUSH`/`RPUSH`, `XADD`, single-key fast paths.
 
-Dynamic (observed-cost) classification is a follow-up; v1 bias is a
-boring, reviewable list.
+The entire `ZRANGE` family is gated, not only "full-range" variants —
+arg inspection (e.g., detecting `LIMIT 0 N`) breaks the "classify by
+command byte" simplicity, and a bounded `ZRANGE 0 10` contributes at
+most one unmarshal per request (cheap). Dynamic (observed-cost)
+classification is a follow-up; v1 bias is a boring, reviewable list.
 
 ### Tradeoffs
 
@@ -123,6 +128,30 @@ point, sized `2 × runtime.GOMAXPROCS(0)` (env-overridable). Gated
 commands in `dispatchCommand` call `Submit`; ungated stay
 synchronous. Static list lives next to `dispatchCommand`. Pool-full →
 `-BUSY server overloaded`. Lua follows option (A).
+
+**Container-aware sizing.** `runtime.GOMAXPROCS(0)` on Linux returns
+the host CPU count, not the container's cgroup quota (unless
+`GOMAXPROCS` is set explicitly or something has configured it).
+Operators running under Kubernetes/Docker with a CPU limit should
+either (a) set `GOMAXPROCS` in the deploy environment to match the
+cgroup quota (the project's rolling-update script is the natural
+place), or (b) wire `go.uber.org/automaxprocs` into `main.go` so the
+correction happens at startup. v1 prefers (a) for auditability;
+(b) is acceptable as a follow-up if operators want it automatic.
+
+**Single pool vs per-class sub-pools.** v1 uses a single global pool.
+The risk: a burst of `KEYS *` or `SCAN` from a management client can
+exhaust all slots and force `-BUSY` onto latency-sensitive `XREAD` or
+Lua requests. Two mitigations exist: (i) classify gated commands into
+priority tiers and reserve a minimum slot share per tier (e.g., 50%
+data-path, 25% scan, 25% Lua), (ii) ship separate sub-pools per
+tier. Both add complexity that's only justified if we actually
+observe a scan-command burst displacing data-path work. **v1 defers
+sub-pools; observability must call this out so the need is
+measurable.** New metric `elastickv_heavy_command_pool_submit_total`
+labelled by command name is sufficient: if pool-full rejections
+concentrate on `XREAD` while `KEYS` dominates successful submissions,
+the tier split is warranted.
 
 ### Where in the code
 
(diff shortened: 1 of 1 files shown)


bootjp:
Addressed Gemini design review in 9d44b6c1:
- HIGH (line 315, sync migration): added **chunked migration** section (`STREAM_MIGRATION_CHUNK=1024` default). Scoped explicitly as a stacked follow-up to PR #620 which ships the simple one-txn version. Doc now acknowledges that a 100k-entry legacy stream cannot be rewritten in a single XADD txn without reproducing the CPU spike.
- MEDIUM (line 84, ZRANGE ambiguity): widened gating to the whole `ZRANGE` family. No arg-inspection → preserves static byte-level classification. A bounded `ZRANGE 0 10` pays one pool submit for one unmarshal; still cheaper than the dispatcher branch that would distinguish bounded vs unbounded.
- MEDIUM (line 122, container-awareness): called out the GOMAXPROCS vs cgroup-quota mismatch. v1 recommends operator-set `GOMAXPROCS` at the rolling-update layer (auditable); `uber-go/automaxprocs` is an accepted follow-up.
- MEDIUM (line 139, single-pool starvation): kept single pool for v1 but required `elastickv_heavy_command_pool_submit_total` labelled by command so a sub-pool split becomes a measurable decision rather than a guess. Sub-pools / slot reservation documented as the named follow-up.
- MEDIUM (line 252, RESP reject): changed v1 from "close TCP without RESP" to "accept, write `-ERR max connections per client exceeded`, close". Client sees a protocol-level signal instead of an opaque connection reset.

/gemini review
@codex review

bootjp:
Addressed round-2 in 731475eb:
- Gemini HIGH line 87 (XREAD BLOCK): carved blocking variants out of pool gating — they bypass the pool and re-gate on wake-up.
- Gemini HIGH line 345 + Codex P1: split the migration section into Mode A (simple, PR #620) and Mode B (chunked, follow-up). Mode B MUST use an always-merge read rule to avoid the mixed-state data loss reviewers flagged.
- Gemini Medium line 117: documented the ctx-sentinel mechanism that makes Lua recursion option (A) implementable.
- Gemini Medium line 283: added reject-storm mitigation (rate-limit the reject with RST fallback + client-side backoff in runbook).
- Gemini Medium line 431: resolved the XREAD gated-vs-cheap contradiction — kept gated in v1 for three concrete reasons, promotion to ungated is metric-driven.
- Codex P2 line 134: corrected GOMAXPROCS guidance — Go 1.25+ already derives from cgroup v2 quota, so the previous automaxprocs recommendation is stale for this repo.

/gemini review
@codex review

bootjp:
Round 3 in 104237dd:
- Codex P1 + Gemini Medium: retracted the "blocking variants bypass pool" decision — current adapter busy-polls XREAD BLOCK / B*POP, so bypass would hand them unbounded CPU. v1 keeps blocking gated; condvar wake-up is listed as the prereq for an honest bypass.
- Gemini Medium: documented that `runLuaScript` currently rebuilds its ctx from `context.Background()`, clobbering the `ctxKeyInPoolSlot` sentinel. Called out explicitly that Layer 1 MUST NOT ship until this is fixed or Lua inside the pool self-deadlocks.
- Gemini HIGH: Mode B cost model now states the O(N_suffix) unmarshal cost during the migration window; added read-driven drain mitigation and top-N legacy-suffix-size metric.
- Codex P1: removal criterion now requires BOTH `elastickv_stream_legacy_format_reads_total` AND a new `elastickv_stream_legacy_format_keys_total` (from a periodic prefix scan) to be zero. Catches cold streams that a read-only metric would miss.

/gemini review
@codex review

bootjp:
Round 4 in 2ae40564:

- Gemini HIGH (line 89 XADD starvation): explicitly flag that Mode A's first-XADD migration rewrite is O(N) and therefore the "ungated" classification is a latent incident if Layer 4 Mode A lands before Layer 1. Added two ordering options — ship together (preferred, simpler) or dynamically gate XADD while the legacy-keys counter is non-zero.
- Gemini Medium (line 438 protobuf partial decode): corrected the Mode B read rule. Protobuf cannot decode a repeated field partially, so the suffix blob is fully unmarshaled and then filtered to the requested ID range in memory. This matches (not contradicts) the Mode B cost model.
- Gemini HIGH (line 475 scan cost) + Codex P2 (line 474 key pattern): the prefix-scan approach to counting legacy keys would iterate every entry of every migrated stream because entries share the `!redis|stream|` prefix; the prior wording also compared `!redis|stream|` keys to a phantom `!stream|meta|` key that isn't in the layout. Replaced both with either a sidecar legacy-index (`!redis|stream_legacy_index|<key>` tombstone) or a per-stream layout walk over `!redis|stream|<key>|meta` keys — O(num_streams), not O(num_entries). Retracted the wrong prefix and noted the authoritative layout.
- Gemini Medium (line 350 RST overhead): pure-Go `net` cannot emit a true kernel RST without `Accept()`; clarified that (a) "cheap" means "skip the RESP write + `SetLinger(0)` close" (fewer syscalls per reject), not "no accept." Real RST-before-accept via raw listener or eBPF is deferred to v2.
- Codex P1 (line 597 removal criterion contradiction): open-question 5 referenced only `reads_total == 0`, contradicting the Layer 4 rule that requires BOTH `reads_total` and `keys_total`. Tied them together.

Doc-only PR; no build needed.

/gemini review
@codex review

bootjp:
**Codex P1 (line 528, legacy-key walk via `|meta`):** the design doc at the HEAD of this branch (commit 2ae40564, round 4) already retracts the incorrect "walk `|meta`" wording. The current text explicitly states:

> A "walk `|meta` keys" alternative was considered and **rejected**: untouched legacy streams never acquire a `|meta` key (Mode A's read rule explicitly expects `|meta` to be absent and falls back to the legacy blob). Iterating `|meta` would miss exactly the cold legacy streams that this counter exists to surface, reintroducing the same blind spot the paired counter is closing.

The authoritative option is a `!redis|stream_legacy_index|<logical>` sidecar marker written when a legacy blob is created and deleted in the same commit as the Mode A migration write. The counter becomes `SCAN !redis|stream_legacy_index|`, bounded by legacy-blob count rather than total entries.

Codex appears to be reviewing a pre-round-4 snapshot. No code or doc change needed.

/gemini review
@codex review

PR #620, opened 2026-04-24
perf(redis): stream entry-per-key layout for O(new) XREAD

bootjp:
## Summary
- Store each Redis stream entry under its own key (`!stream|entry|<userKey><StreamID(16B)>`) with a separate meta key (`!stream|meta|<userKey>` -> `Length | LastMs | LastSeq`); XREAD / XRANGE become bounded range scans that unmarshal only the selected entries, XADD is two small writes, XLEN is a single meta read.
- Migration is dual-read with write-time rewrite: reads bump `elastickv_stream_legacy_format_reads_total` when they fall through to the legacy blob; writes convert the legacy blob to the new layout in the same transaction and delete the blob, so XLEN never double-counts.
- StreamID suffix is binary big-endian `ms || seq` so lex order over entry keys matches the numeric `(ms, seq)` order the client sees; parsing the "ms-seq" string form for range bounds is done once up front to compute the scan bounds.

## Motivation
Incident 2026-04-24: one client doing 11 XREAD/s on a large stream consumed 14 CPU cores on the leader, starving raft and Lua paths. Per the CPU profile, `proto.Unmarshal` alone took 59% of 14 cores because `loadStreamAt` read the entire stream as a single protobuf blob and `unmarshalStreamValue` re-parsed every entry on every read.

## Design
- Layout: meta key holds a fixed 24-byte binary record (`Length(8) | LastMs(8) | LastSeq(8)`); entry keys embed the StreamID in big-endian binary so a prefix range scan is enough to serve XRANGE / XREAD without loading the whole stream. `LastMs` / `LastSeq` track the highest ID ever assigned, so XADD `*` stays monotonic even after XTRIM removes the current tail.
- Commands: XADD / XREAD / XRANGE / XREVRANGE / XLEN / XTRIM all rewritten to use the new layout. XADD with MAXLEN trims the head by scanning exactly `count` entry keys and emitting Dels in the same txn as the new Put.
- Migration: reads try the new meta first; on miss, fall back to the legacy blob and increment the counter. Writes that observe the legacy blob re-emit every legacy entry as a `!stream|entry|…` put, emit the new meta, and delete the legacy blob, all in a single coordinator-dispatched transaction. The legacy blob and the new layout never coexist in a committed state. `redisStreamKey` is marked Deprecated and kept only for the dual-read fallback.

## Commands touched
`XADD`, `XREAD`, `XRANGE`, `XREVRANGE`, `XLEN`, `XTRIM`. `deleteLogicalKeyElems` also learns to clean up the new meta + every entry key so `DEL` and overwrite paths stay correct.

## Test plan
New tests in `adapter/redis_compat_commands_stream_test.go`:
- [x] `TestRedis_StreamXAddXReadRoundTrip` — XADD then XREAD `0` returns all entries.
- [x] `TestRedis_StreamXReadLatencyIsConstant` — 10 000 XADDs, then 100 XREADs from `$` must not grow beyond `2 * first + 10ms`. On the old blob path this failed catastrophically.
- [x] `TestRedis_StreamXTrimMaxLen` — XADD 100, XTRIM MAXLEN=10, XLEN==10, XRANGE returns the last 10.
- [x] `TestRedis_StreamXRangeBounds` — inclusive / exclusive (`(`) bounds on both sides, plus XREVRANGE.
- [x] `TestRedis_StreamMigrationFromLegacyBlob` — seed a legacy blob directly, XREAD bumps the counter, XADD migrates and deletes the blob, subsequent XREAD hits the new layout and does NOT bump the counter, XLEN is original + new (no double count), XADD `*` is > the pre-migration last ID.
- [x] `TestRedis_StreamAutoIDMonotonicAfterTrim` — XTRIM that empties the stream must not rewind XADD `*`.

`go build ./... && go vet ./... && go test -short ./adapter/... ./store/...` all pass locally.

## Follow-ups (not in this PR)
- Migrate hash / set / zset single-blob fallbacks off the entry-per-key layouts the same way (Layer 4 design doc already lists these).
- XINFO, XACK, consumer groups — none exist today.
- Delete the `redisStreamKey` helper + legacy blob fallback once `elastickv_stream_legacy_format_reads_total` stays at zero across production nodes for long enough to guarantee no surviving legacy streams.

## Summary by CodeRabbit

* **New Features**
  * Improved stream add semantics with stricter ID validation and guaranteed monotonic auto-IDs.
  * Multikey stream deletions now reliably remove all stream data within transactions.

* **Bug Fixes**
  * Corrected XRANGE/XREVRANGE inclusive/exclusive boundaries and COUNT clamping.
  * Ensured XLEN/XTRIM reflect actual entries and legacy-format remnants are cleaned on write.

* **Performance**
  * Reduced blocked read latency and stabilized BLOCK timeout behavior to return null on timeout.

* **Tests**
  * Added extensive end-to-end and unit tests covering XADD/XREAD/XTRIM/XRANGE, deletion, and edge cases.


bootjp:
Pushed `66a073a5` addressing the review feedback. Summary:

### Critical correctness fixes
- **Codex P1 (full-stream scan before every XADD)**: `streamWriteBase` now splits into a fast path and a migration path. When the stream is already in the new layout we do a single meta-only read — no entry scan — so post-migration XADD is back to O(1) in stream size. The migration path (legacy blob present) is the only code that still loads entries, and only once, which is the original design intent.
- **Gemini critical + CodeRabbit critical (ctx / readTS not threaded into migration reads)**: `xaddTxn`, `xtrimTxn`, `streamWriteBase`, and `buildXTrimHeadElems` now all use the outer transaction's `ctx` and `readTS`. The previous code mixed `context.Background()` and a fresh `r.readTS()` inside the helpers, which broke MVCC atomicity: the commit landed at `readTS` while the scan saw a later snapshot, so concurrent entries in the gap could be trimmed out from under the caller.
- **CodeRabbit critical (MULTI/EXEC DEL leaves new namespace behind)**: `txnContext` now has a `streamDeletions` set. `stageKeyDeletion` populates it (so `MULTI/EXEC DEL stream` or `EXPIRE stream 0` marks both the legacy blob and the wide-column layout for tombstoning), and `commit()` expands it at the transaction's `startTS` via `deleteStreamWideColumnElems`. `trackTypeReadKeys` also tracks `store.StreamMetaKey(key)` so concurrent writers trigger OCC conflicts.
- **CodeRabbit major (KEYS / reverse-mapping)**: `redisInternalPrefixes` includes `store.StreamMetaPrefix`; `mergeInternalNamespaces` scans it via the length-prefix-aware bound; `extractRedisInternalUserKey` and `wideColumnVisibleUserKey` reverse-map meta keys back to the user key and classify entry keys as internal-only. Meta is the uniqueness anchor, so each migrated stream appears exactly once (not once per entry).

### Linter issues flagged by reviewdog
- **gci**: reformatted the touched `adapter/`, `store/stream_helpers.go`, and `monitoring/redis.go` via `gci write -s standard -s default -s "prefix(github.com/bootjp/elastickv)"`.
- **gocritic appendAssign** at 3 call sites (migrationElems): rewrote to pre-allocate `elems` and copy via `append(elems, migrationElems...)`.
- **gocritic ifElseChain**: converted the ms/seq overflow chain in `streamBoundHigh` to a `switch`.
- **gosec G115**: centralised the int64→uint64 time-conversion in test helper `nowNanos(t)` with an explicit positive bound.
- **mnd**: extracted `streamMetaLengthSignBit` (63) in `store/stream_helpers.go` and `xaddFixedElemCount` (2) in `xaddTxn`.
- **staticcheck SA9003**: removed the empty branch in `scanStreamEntriesAt`; documented `expectedLen` as a reserved hint.
- **wrapcheck**: wrapped `store.MarshalStreamMeta` / `UnmarshalStreamMeta` / `cockerrors.Newf` with `errors.WithStack`.
- **cyclop**: split `xaddTxn` into `xaddTxn` + `xaddTrimIfNeeded`; split `resolveXReadAfterIDs` into `resolveXReadAfterIDs` + `resolveXReadDollarID` + `dollarIDFromState`; converted `extractRedisInternalUserKey` to a table-driven trim-prefix pass.
- **goconst**: `"0-0"` is now the package constant `streamZeroID`.

### New tests
- `TestRedis_StreamMigrationWithMaxLenTrim`: seeds a legacy blob and issues `XADD ... MAXLEN ...` small enough to drop some migrated entries in the same transaction. Documents the apply-order contract that the coordinator's sequential Put-then-Del resolution provides.
- `TestRedis_StreamMultiExecDelRemovesWideColumnLayout`: regression guard for the MULTI/EXEC DEL wide-column cleanup path.

### Verification
- `go build ./...` clean
- `go vet ./...` clean
- `golangci-lint run ./...` returns zero issues
- Stream test suite (including `TestRedis_StreamXReadLatencyIsConstant`'s 10k-entry O(new) guard) passes locally

Other review nits addressed: the dead `hasLast` tracking in `streamWriteBase` was removed as part of the migration-path rewrite. The `<< 1` readTS in `TestRedis_StreamMigrationFromLegacyBlob` is now `nowNanos(t) + uint64(time.Minute)` per the CodeRabbit nitpick.

/gemini review
@codex review

bootjp:
Round 3 in 5ea7f5e4:
- Gemini Medium (resolveXReadDollarID:4130): threaded caller's ctx through resolveXReadAfterIDs → resolveXReadDollarID → dollarIDFromState. Cancellation of the outer BLOCK deadline now tears down the in-flight dollar-resolve scan.
- Gemini Medium (xaddTxn:3714): extracted `estimateXAddTrimCount` to pre-size the elems slice when MAXLEN will trim; avoids repeated slice growth on large streams hitting the cap. Kept xaddTxn within the cyclop budget.
- Bonus: xread now derives ctx from the BLOCK deadline so dollar resolve + the busy-poll share one cancellation path.

/gemini review
@codex review

bootjp:
Round 4 — no new actionable feedback since round 3 (5ea7f5e4).

No new inline comments from Gemini, Codex, or CodeRabbit since the latest bootjp comment. CI is still running on round 3's commit. Leaving the branch at 5ea7f5e4.

/gemini review
@codex review

bootjp:
Round 4 in 50def0d7 (additional, see prior comment for no-reviewer-feedback state):

Stabilised `TestRedis_StreamXReadLatencyIsConstant` which was flaking red on CI under `-race` against the branch. The failure was a ~33% overshoot of a 2x-first-sample ceiling, not an algorithmic regression — a cold-first + max-of-100-under-race combination was absorbing scheduler jitter into the budget.

Rework:
- 8-probe warmup; baseline = median of warmups (not a single cold sample).
- Assert `median <= 3*baseline + 20ms` as the O(new) guard.
- Assert `p95 <= 6*baseline + 40ms` for scheduler-jitter headroom.

Under the old blob implementation every probe was O(stream_size), so a 10k-entry stream still produces the original 10x+ regression signal — the widened thresholds do not hide the class of bug this test guards against. Verified locally (non-race): test passes in 142s.

/gemini review
@codex review

bootjp:
Addressed in 8bc3eebe:

**Codex P2 line 4338 (BLOCK timeout → error regression, introduced in round 3):** `$` resolution now uses its own `redisDispatchTimeout` context; the BLOCK deadline only governs the busy-poll loop. Short BLOCK windows no longer surface `context deadline exceeded` to clients.

**Comments I'm declining with justification:**

- **Gemini HIGH line 3725 (xaddTxn maxWideColumnItems enforcement):** already present at line 3699 — `xaddEnforceMaxWideColumn` rejects XADD that would push `currentLength` past `maxWideColumnItems` unless a MAXLEN clause will cap the growth. The reviewer is reading stale state.
- **Gemini HIGH line 4256 / Codex P1 line 4256 (XREAD without COUNT silent truncation):** `scanStreamEntriesAfter` already raises `ErrCollectionTooLarge` at line 4270-4272 when an unbounded scan would exceed `maxWideColumnItems`. Not silent; explicit error.
- **Gemini HIGH line 4519 (rangeStreamNewLayout):** same guard at line 4572-4574.
- **Gemini Medium line 4288 / Codex P2 line 4287 (streamScanStartForAfter max-ID edge):** handled at line 4322-4326: when both ms and seq are `^uint64(0)`, returns `PrefixScanEnd(prefix)` so the inclusive start + half-open end yields an empty scan.
- **Gemini Medium line 4590 (streamBoundLow/streamBoundHigh max-ID edge):** same pattern at line 4641-4642 and 4669-4670.
- **Gemini Medium line 3773 (MAXLEN 0 trim-to-empty):** `parseXAddMaxLen` (line 3562) returns 0 for both "MAXLEN absent" and "MAXLEN 0"; treating `maxLen == 0` as trim-to-empty would make every un-MAXLEN'd XADD trim. Proper fix needs the parser to use a sentinel (-1 for absent) and every caller of `xaddRequest.maxLen` audited. Out of scope for this PR; tracked as follow-up.
- **Gemini HIGH line 3729 (migration + MAXLEN redundant ops):** the end-state is correct per `TestRedis_StreamMigrationWithMaxLenTrim` (later Del tombstones earlier Put at the same commitTS). Skipping the migrated-then-trimmed entries would require plumbing `maxLen` into `streamWriteBase`; deferred as a perf optimization since the transaction size is bounded by `maxWideColumnItems` in the cap check above.

/gemini review
@codex review

bootjp:
Addressed both Codex P2 regressions in a6b176d4:

**Shorthand stream IDs in XRANGE/XREVRANGE bounds (`redis_compat_commands.go:4637-4639`):** `streamBoundLow` / `streamBoundHigh` now accept both strict `ms-seq` and the shorthand `ms` form. New `parseStreamBoundID` helper expands shorthand to either `(ms, 0)` or `(ms, MaxUint64)` depending on which side of the ±1 exclusive shift produces the Redis-specified "match all entries with this ms" half-open range. New regression guards in `TestRedis_StreamXRangeBounds`:
- `XRANGE k 0 +` (lower-inclusive shorthand)
- `XRANGE k 1001 1002` (both-inclusive shorthand)
- `XRANGE k - (1002` (upper-exclusive shorthand)

**Hard 100k cap in `scanStreamEntriesAt` (`redis_compat_helpers.go:562-563`):** this is the path `loadStreamAt` uses to rebuild full stream state for the Lua bridge (`streamState`). Legacy blob reads had no size cap; the new-layout scan failed any migrated stream with >100k entries, which is a real regression for large-stream Lua scripting. Sized the scan from `expectedLen + 64` (meta.Length at the caller's snapshot + slack for concurrent writes) and dropped the hard cap. `scanStreamEntriesAfter` and `rangeStreamNewLayout` keep their caps because they serve user-driven unbounded requests (XREAD/XRANGE without COUNT) where OOM protection is needed.

/gemini review
@codex review

bootjp:
Addressed Codex P1 in 96ca38d3:

`deleteStreamWideColumnElems` previously routed through `scanAllDeltaElems`, which caps at `maxWideColumnItems` and errors once that threshold is reached. After migration, a stream that legally grew past the cap in legacy blob form became undeletable — `DEL`, `EXPIRE 0`, and the `MULTI/EXEC` path via `txnContext.buildStreamDeletionElems` all returned `ErrCollectionTooLarge` and left keys stranded in storage.

Fix: introduce `scanStreamEntryKeysForDelete` — paginated but uncapped walk over the entry prefix, used only from the delete path. User-driven unbounded scans (XREAD/XRANGE without COUNT) still cap via `scanStreamEntriesAfter` / `rangeStreamNewLayout` because there OOM protection is meaningful; DEL has no such concern (the legacy blob DEL also had no cap).

Pagination page size remains `MaxDeltaScanLimit` so the server never materialises more than one page in the `ScanAt` call; the returned Del elems are handed to the coordinator in one transaction. Large streams produce large transactions, matching legacy behaviour where DEL on a 1MB blob also emitted a single transaction.

/gemini review
@codex review

bootjp:
Addressed the three Copilot review comments:

**1. `deleteStreamWideColumnElems` uses `scanAllDeltaElems` (100k cap)** — already fixed in **96ca38d3** (preceded the Copilot review by ~10 min). `scanStreamEntryKeysForDelete` is a new uncapped paginated walk specifically for the delete path; user-driven unbounded scans (XREAD/XRANGE without COUNT) still cap in `scanStreamEntriesAfter` / `rangeStreamNewLayout`.

**2. `xlen` narrows int64 → int** (fixed in **8cf12e29**): switched to `conn.WriteInt64(meta.Length)`. Prevents overflow on 32-bit targets or any stream whose length legitimately exceeds `MaxInt32`. The legacy fallback path (`len(stream.Entries)`) is updated to `int64(…)` for consistency.

**3. `marshalStreamEntry` doc comment inaccurate** (fixed in **8cf12e29**): chose to update the comment rather than drop Id from the stored value. Keeping Id in the value lets `unmarshalStreamEntry` return a fully-formed entry without making every caller (XREAD, XRANGE, XREVRANGE, Lua `streamState`) parse the storage key back. Cost is ~16 bytes per entry; benefit is zero key-parsing plumbing at call sites. Comment now explains the intentional duplication.

/gemini review
@codex review

bootjp:
@copilot apply changes based on the comments in [this thread](https://github.com/bootjp/elastickv/pull/620#pullrequestreview-4172339726)

bootjp:
@copilot apply changes based on the comments in [this thread](https://github.com/bootjp/elastickv/pull/620#pullrequestreview-4172535084)

bootjp:
@copilot apply changes based on the comments in [this thread](https://github.com/bootjp/elastickv/pull/620#pullrequestreview-4172647774)  and https://github.com/bootjp/elastickv/pull/620#discussion_r3139651007

bootjp:
@copilot+claude-opus-4.6 [adapter/redis_compat_commands.go](https://github.com/bootjp/elastickv/pull/620/files/88ffdfa0048a6b71744934f5c2172547d706044b#diff-287419c0e8d28197a616be1e7634e27990db84456c96cd725ba3835b083c2523)
Comment on lines +3857 to +3858
	if len(entriesToMigrate) > maxWideColumnItems {
		entriesToMigrate = entriesToMigrate[len(entriesToMigrate)-maxWideColumnItems:]
@[chatgpt-codex-connector](https://github.com/apps/chatgpt-codex-connector)
chatgpt-codex-connector Bot
[23 分前](https://github.com/bootjp/elastickv/pull/620#discussion_r3139651007)
P1 Badge Do not silently truncate legacy streams during migration

streamWriteBase unconditionally slices legacy streams down to maxWideColumnItems before returning migration elems, which means commands that should be no-ops can drop data. For example, on a pre-migration stream with 150k entries, XTRIM key MAXLEN 200000 will report removed=0 in xtrimTxn but still migrate only the newest 100k entries because this truncation already happened. That is silent data loss on a client-visible no-op path; this should fail explicitly (or require explicit trim) rather than rewriting a truncated stream.[adapter/redis_compat_commands.go](https://github.com/bootjp/elastickv/pull/620/files/88ffdfa0048a6b71744934f5c2172547d706044b#diff-287419c0e8d28197a616be1e7634e27990db84456c96cd725ba3835b083c2523)
Comment on lines +3857 to +3858
	if len(entriesToMigrate) > maxWideColumnItems {
		entriesToMigrate = entriesToMigrate[len(entriesToMigrate)-maxWideColumnItems:]
@[chatgpt-codex-connector](https://github.com/apps/chatgpt-codex-connector)
chatgpt-codex-connector Bot
[23 分前](https://github.com/bootjp/elastickv/pull/620#discussion_r3139651007)
P1 Badge Do not silently truncate legacy streams during migration

streamWriteBase unconditionally slices legacy streams down to maxWideColumnItems before returning migration elems, which means commands that should be no-ops can drop data. For example, on a pre-migration stream with 150k entries, XTRIM key MAXLEN 200000 will report removed=0 in xtrimTxn but still migrate only the newest 100k entries because this truncation already happened. That is silent data loss on a client-visible no-op path; this should fail explicitly (or require explicit trim) rather than rewriting a truncated stream.

bootjp:
@copilot apply changes based on the comments in [this thread](https://github.com/bootjp/elastickv/pull/620#pullrequestreview-4172826300)

bootjp:
@copilot apply changes based on the comments in [this thread](https://github.com/bootjp/elastickv/pull/620#pullrequestreview-4172938769)

bootjp:
@copilot apply changes based on the comments in [this thread](https://github.com/bootjp/elastickv/pull/620#pullrequestreview-4173120168)

bootjp:
Per the operator directive "pre-migration data is expendable; clear explicitly," dropped the dual-read / migration machinery from this PR in **bedb3de7**. Rebased on top of the Copilot SWE-agent's recent fixes (MAXLEN sentinel, shorthand IDs, 0-0 rejection, prefix precompute).

**Behaviour change (streams not yet written to in the new layout):**
- Reads on a legacy-only stream return empty, not the legacy entries. `XREAD`, `XRANGE`, `XLEN`, `XREVRANGE` all report "stream does not exist."
- The next write against the stream (`XADD` or `XTRIM`) deletes the legacy blob in the same transaction as the new-layout meta+entry write. Entries are NOT materialised from the legacy value.
- `MULTI/EXEC DEL` / `EXPIRE 0` also sweeps the legacy blob via `deleteStreamWideColumnElems` so operator-driven deletes leave no orphans on disk.

**Code changes:**
- `loadStreamAt`: new-layout only; legacy `GetAt(redisStreamKey)` fallback removed.
- `streamWriteBase` now returns `(legacyCleanup, meta, found, err)` — no more `migrationElems` / `existingEntries`. New `legacyStreamCleanupElems` helper emits a Del for the legacy key when present.
- `xaddTxn` / `xtrimTxn`: wire `legacyCleanup` into the elem list.
- `buildXTrimHeadElems` signature loses `migrationActive` and `existingEntries`; always uses the bounded range scan over the new layout.
- `resolveXAddID` signature loses `existingEntries`; auto-ID derived solely from `meta.LastMs` / `meta.LastSeq`.
- `deleteStreamWideColumnElems` also deletes the legacy key in the same commit.
- Removed: `observeLegacyStreamRead`, `streamLegacyReadObserver`, `WithStreamLegacyFormatReadObserver`, `StreamLegacyFormatReadObserver`, `elastickv_stream_legacy_format_reads_total`, and the registry accessor. The counter's invariant (non-zero ⇒ fallback still exercised) is now structurally enforced by the code.

**Test changes:**
- `TestRedis_StreamMigrationFromLegacyBlob` → replaced by `TestRedis_StreamLegacyDataIsDiscarded`, which proves the new contract: legacy-only stream reads empty, next write deletes the legacy blob, post-write state is only the freshly-added entry.
- `TestRedis_StreamMigrationWithMaxLenTrim` removed — no migration path to cover.
- `TestRedis_StreamMultiExecDelRemovesWideColumnLayout` unchanged — still guards DEL touching meta + entry keys + the legacy key.
- Kept the Copilot SWE-agent's `TestRedis_StreamXAddMaxLenZero`.
- Also in: `appendMaxLenZeroSelfDel` helper to keep `xaddTxn` under the cyclop budget after the MAXLEN-0 self-delete branch landed.

Build / vet / lint / targeted stream tests clean.

/gemini review
@codex review

bootjp:
Addressed the Copilot round-3 review in 428888ae — every fix ships with a matching test as requested.

**Correctness fixes (all four int-overflow / wrap concerns):**

1. **`nextXAddID` seq wrap at `MaxUint64`** — extracted `bumpStreamID` helper that carries seq→ms+1 on overflow, returns `ERR The stream has exhausted the ID space` when both are at max instead of silently producing a duplicate `0-0`. New `TestBumpStreamID` covers all three branches; new `TestNextXAddID_Monotonic` exercises the path through `nextXAddID` with a far-future `lastMs` so the "nowMs < lastMs" branch is forced.
2. **`estimateXAddTrimCount` `int(nextLen) - maxLen` wrap** — arithmetic now in `int64`, clamped to `[0, math.MaxInt]`. New `TestEstimateXAddTrimCount` covers MAXLEN=0, MAXLEN-at-cap, the MaxInt64 corrupt-meta case.
3. **`xaddTrimIfNeeded` `int(candidateLen) - maxLen` wrap** — same `int64` arithmetic + clamp.
4. **`xtrimTxn` `removed = int(meta.Length) - maxLen` wrap** — same treatment. Also extracted `streamTypeForWrite` + `flushLegacyCleanupOnTrimNoOp` helpers to keep `xtrimTxn` under the cyclop budget after the clamp added a branch.

**Stale comments fixed:**

- `dollarIDFromState`: the doc-comment claimed a legacy-blob fallback that no longer exists; the body also had a dead legacy branch. Both dropped; the function now returns `streamZeroID` when meta is absent and comments match.
- `readStreamAfter`: same — doc and body both said "falls through to the legacy blob"; now just returns `nil, nil` when meta is absent.
- `redis_compat_commands_stream_test.go`: the header comment for `TestRedis_StreamMultiExecDelRemovesWideColumnLayout` had been prefixed by a stale comment describing the removed `TestRedis_StreamMigrationWithMaxLenTrim`. Dropped.

**Declined (resolved in an earlier commit):**

- **"xreadOnce uses `context.Background()`"** — already resolved in 3dd2f790. `xread` derives a per-iteration ctx from the BLOCK deadline and threads it through `xreadOnce → readStreamAfter → scanStreamEntriesAfter`. The reviewer appears to be reading a pre-3dd2f790 snapshot.
- **"`extractRedisInternalUserKey` returns nil for stream entry keys"** — already returns `store.ExtractStreamUserKeyFromEntry(key)` at line 282. Same "reading older snapshot" situation.
- **"PR description says dual-read/rewritten"** — the current PR body accurately describes the "discard-on-read, delete-on-write" behavior per the operator directive. No change needed.

Build / vet / lint / targeted tests all clean.

/gemini review
@codex review

bootjp:
Addressed Codex P1 (BLOCK-expired-on-entry returns error instead of null) in f3c52649.

Guard added at the top of each busy-poll iteration: `iterTimeout := time.Until(deadline); if iterTimeout <= 0 { conn.WriteNull(); return }`. Pre-fix, very small BLOCK windows (`BLOCK 1`) or windows where $-resolution consumed most of the budget could enter the loop with non-positive remaining time → `context.WithTimeout(0)` returned an already-expired context → `xreadOnce` surfaced `DeadlineExceeded` → the command replied with `-ERR context deadline exceeded`, breaking the Redis-spec null-on-timeout contract.

Test: `TestRedis_StreamXReadShortBlockReturnsNullNotError` does `XADD` of one entry then `XREAD BLOCK 1 ... \$`. Pre-fix returned a redis error; post-fix returns `redis.Nil`. Targeted test passes; build / vet / lint clean.

/gemini review
@codex review

bootjp:
Addressed Gemini's round-4 feedback (4 HIGH + 3 MEDIUM) in 9a8ec49b — every change ships with a matching test as requested.

**HIGH (4) — clamp at `maxWideColumnItems`, not `math.MaxInt`:**

- **`estimateXAddTrimCount` (line 3824)** — capacity hint now clamps at `maxWideColumnItems`. A `math.MaxInt` hint would let `make()` try to allocate ~16 EiB on 64-bit targets (panic / OOM).
- **`xaddTrimIfNeeded` (line 3854)** — actual trim count handed to `buildXTrimHeadElems` clamped at `maxWideColumnItems`. A single XADD must not emit more than that many Del ops in one Raft commit (msg-size / memory pressure).
- **`buildXTrimHeadElems` ScanAt (line 3940)** — defense-in-depth clamp at the boundary that actually allocates. Callers already cap; this guards against any future caller that forgets.
- **`xtrimTxn removed` (line 4062)** — same `maxWideColumnItems` clamp on the planned removal count.

**MEDIUM (3) — `meta.Length` / XTRIM return must follow `len(trim)`:**

- **`xaddTrimIfNeeded` (line 3859)** — returns `candidateLen - int64(len(trim))` instead of `int64(maxLen)`. When the cap kicks in or the scan returns fewer rows than requested, the planned and actual counts diverge; tracking `len(trim)` keeps `meta.Length` consistent with on-disk rows. MAXLEN 0 stays at 0 explicitly because `appendMaxLenZeroSelfDel` removes the freshly-added entry separately.
- **`xtrimTxn meta.Length -= ...` (line 4070)** — switched from `int64(removed)` (planned) to `int64(len(trim))` (actual).
- **`xtrimTxn` return value (line 4076)** — switched from the planned `removed` to `len(trim)`. XTRIM's wire-level reply now matches physical reality.

**Tests added:**
- `TestEstimateXAddTrimCount`: added `maxWideColumnItems`-boundary cases (passes-through at the cap, clamps when over).
- `TestRedis_StreamXTrimReturnAndMetaTrackLenTrim`: end-to-end guard that XTRIM's returned count equals the actual deletions, that XLEN matches XRANGE post-trim, and that an above-length MAXLEN reports zero deletions without drifting `meta.Length`.

Build / vet / lint / targeted stream tests clean.

/gemini review
@codex review

bootjp:
Addressed CodeRabbit + Gemini round 5 feedback in 6e5364e8 — every change ships with a matching test, per the standing rule.

**CodeRabbit Major — `redis.go` `buildStreamDeletionElems` ran on server-lifetime ctx:**

- `commit()` now derives the `redisDispatchTimeout`-bounded ctx **before** calling `buildStreamDeletionElems` and threads it through, instead of `buildStreamDeletionElems` reaching for `t.server.handlerContext()`. The same ctx now bounds both the paginated `ExistsAt` + `scanAllDeltaElems` inside `deleteStreamWideColumnElems` *and* the final `Dispatch`. A pathological staged-stream count (slow disk, hot key) can no longer hold the EXEC handler open past the per-request budget.
- Test: `TestRedis_StreamMultiExecDelHonoursDispatchTimeout` — smoke-level guard that MULTI/EXEC DEL still removes a populated stream and `XLEN` reports 0 afterwards on the now-bounded path.

**CodeRabbit Minor — `store/stream_helpers.go` uint32 overflow:**

- `ExtractStreamUserKeyFromMeta` / `ExtractStreamUserKeyFromEntry` previously compared `uint32(wideColKeyLenSize)+ukLen` (and `+uint32(StreamIDBytes)`) against `len(trimmed)` in `uint32` arithmetic. A corrupted `ukLen` near `math.MaxUint32` would wrap, slip past the guard, and panic on the `trimmed[lo:hi]` slice. Bounds check now done in `uint64`.
- Tests: `TestExtractStreamUserKeyFromMeta_BoundsOverflow` and `TestExtractStreamUserKeyFromEntry_BoundsOverflow` craft a key with `ukLen=math.MaxUint32` and assert nil (no panic). Round-trip happy paths included as a regression net.

**Gemini Medium x2 — XREAD / XRANGE / XREVRANGE COUNT must clamp at `maxWideColumnItems`:**

- `parseXReadCountArg` (line 4125) and `parseRangeStreamCount` (line 4552) silently clamp client-supplied `COUNT` at `maxWideColumnItems`. Without this, a client sending `COUNT` near `math.MaxInt` would force pre-allocation of a maxInt-sized `[]redisStreamEntry` slice and pull more entries than the equivalent uncapped scan would have returned. The `-1` sentinel (no COUNT) is preserved so the unbounded path still trips `rangeStreamNewLayout`'s `maxWideColumnItems` guard.
- Tests:
  - `TestParseXReadCountArg_Clamp` / `TestParseRangeStreamCount_Clamp` exercise the parse-level boundary (valid `<= cap` passes through, above-cap clamps, `0`/`-1`/garbage still error).
  - `TestRedis_StreamXReadAndRangeCountClamp` end-to-end check that `XREAD`, `XRANGE`, and `XREVRANGE` with `COUNT = maxWideColumnItems*4` all succeed and return the expected entries.

**Codex P1 "preserve legacy stream data":** declined again, consistent with the prior rounds and the explicit user directive on this PR — the migration path was deliberately removed in favour of "discard pre-migration data on next write." Live legacy data is not preserved by design.

Build / vet / lint / targeted stream tests clean.

/gemini review
@codex review

bootjp:
Addressed Codex P1 round 6 in 23c38a2d — the last item on this PR per the standing instruction.

**Fix:** when the per-iteration ctx fires *during* `xreadOnce` (mid-call, e.g. `XREAD ... BLOCK 10` against a busy / slow node where ScanAt overshoots its slice of the BLOCK budget), `xreadOnce` returns `context.DeadlineExceeded` — bare, `cockroachdb/errors`-wrapped, or gRPC `status.Error(codes.DeadlineExceeded)` after going through `coordinator.Dispatch`. The busy-poll loop previously surfaced that as `-ERR context deadline exceeded`, which violates the Redis BLOCK-timeout contract (null reply on timeout).

The earlier f3c52649 covered the `iterTimeout <= 0` *entry* path; this round closes the *mid-call* path that survives that early-return.

`isXReadIterCtxError` now silently translates all four ctx shapes — bare sentinel, `cockerrors.Is`, and gRPC `codes.DeadlineExceeded` / `codes.Canceled` — into "empty iteration"; the loop falls through to the deadline check and replies null. Non-ctx errors (Unavailable, Internal, wrong-type, etc.) still propagate to the client unchanged. The busy-poll body is extracted into `xreadBusyPoll` to keep `xread` under the cyclop budget.

**Tests:**

- `TestRedis_StreamXReadIterCtxDeadlineReturnsNull`: `BLOCK 30 ms` reading past the only existing entry, repeated 5 times. iterTimeout starts > 0 each call (so the existing `iterTimeout <= 0` early-return does not catch it); on a busy CI runner the ScanAt round trip can land near or past the iter budget, exercising the mid-call DeadlineExceeded path. The reply must always be `redis.Nil`.
- `TestIsXReadIterCtxError`: parameterised over all four error shapes the helper has to handle plus negative cases (Unavailable, Internal, plain non-ctx errors must NOT be swallowed).

Build / vet / lint / `-race` tests clean.

/gemini review
@codex review

bootjp:
**Re: Codex P2 "Narrow internal key prefix to stream meta/entry namespaces" — 対応必要でした。** 追加で CI で見えた Lua bridge の隠れバグも直しました。両方 a31080f3 で修正、それぞれ回帰テスト付きです。

**Codex P2 (`adapter/redis_compat_types.go`):** `knownInternalPrefixes` の `!stream|` は内部名前空間として広すぎ。本当に内部なのは `!stream|meta|` と `!stream|entry|` だけで、`!stream|foo` のようなユーザーキーは Redis-protocol 上ふつうに合法。広い `!stream|` umbrella だと `txnContext.load` が `redisStrKey()` ラップをスキップして bare `!stream|foo` 鍵に直接アクセスしてしまい、MULTI/EXEC 内の SET/GET/DEL が内部名前空間に書き込む or データを取りこぼす可能性があった。

→ Fix: `[]byte(store.StreamMetaPrefix)` と `[]byte(store.StreamEntryPrefix)` の2つに分割。

→ Tests:
- `TestIsKnownInternalKey_StreamPrefixNarrowed` (unit) — narrowed prefix list の境界を網羅。
- `TestRedis_UserKeyShadowingStreamPrefixSurvivesMultiExec` (e2e) — MULTI/EXEC で `SET "!stream|foo" "user-value"` → `GET "!stream|foo"` が round-trip すること。

**ついでに見つけた CI regression — `TestRedis_LuaBullMQLikeHelpers`:**

Lua bridge の `streamCommitElems` が round-5 以降も**まだレガシー単一 blob 形式**で書き込んでいた (`redisStreamKey` + `marshalStreamValue`)。一方 read path (`loadStreamAt` / `scanStreamEntriesAt`) は PR #620 以来 "new-layout only, discard-on-read"。結果、Lua script から `redis.call("XADD", ...)` した直後の `XRANGE` / `XLEN` が空を返し、CI の bullmq テストが落ちていた。

→ Fix: `streamCommitElems` を新 wide-column layout (`StreamEntryKey` Put × N + `StreamMetaKey` Put) に書き換え。caller (`commitPlanForKey`) が事前に `deleteLogicalKeyElems` で全レイアウトをクリアしてくれるので、フルリライトで安全。

→ Test: 既存の `TestRedis_LuaBullMQLikeHelpers` がそのまま regression test。pre-fix で fail、post-fix で pass。

Build / vet / lint clean.

/gemini review
@codex review

PR #638, opened 2026-04-25
feat(sqs): finish Milestone 1 - purge, batch, tags, redrive, FIFO, retention reaper

bootjp:
## Summary

Closes the remaining `NotImplemented` gaps on the SQS adapter so it can serve real client workloads, not just the catalog and the single-message hot path.

- **Queue ops**: PurgeQueue (with AWS's 60s cooldown persisted on the meta record), TagQueue / UntagQueue / ListQueueTags. GetQueueAttributes now returns ApproximateNumberOfMessages / NotVisible / Delayed, QueueArn, CreatedTimestamp, LastModifiedTimestamp.
- **Message ops**: SendMessageBatch / DeleteMessageBatch / ChangeMessageVisibilityBatch with per-entry validation. MessageAttributes is fully wired: structured input, AWS-canonical MD5 (length-prefixed sorted encoding + type byte), filtered echo on Receive.
- **DLQ redrive**: RedrivePolicy is parsed, persisted, and enforced — receive transactionally moves a message to the DLQ when it would exceed maxReceiveCount.
- **FIFO**: per-queue monotonic SequenceNumber, 5-minute dedup window (`!sqs|msg|dedup|`), and group lock (`!sqs|msg|group|`) that keeps the head pinned across visibility expiries — released only on Delete / DLQ redrive / retention.
- **Retention reaper**: new `!sqs|msg|byage|` send-age index plus a leader-side goroutine that drops expired records (data + vis + byage + optional group lock) under one OCC dispatch per record. Per-queue budget prevents starvation; followers short-circuit so only the leader emits Dispatches.

## Test plan

- [x] `go vet ./...`
- [x] `go test -run 'TestSQS' ./adapter/...`
- [x] 11 new integration tests covering purge cooldown, batch shape errors, MD5 round-trip, malformed attributes, DLQ redrive, FIFO sequence / dedup / content-based dedup, FIFO group lock pinning across visibility expiry, retention reaper.
- [x] `golangci-lint run --config=.golangci.yaml ./adapter/...` clean.

## Out of scope

- Long-poll cross-node notifier (still polls every 200ms; design §7.3 considers polling acceptable for Milestone 1).
- Operator console UI (`adapter/console.go`).
- Jepsen workload (`jepsen/sqs/`).
- Query-protocol XML compatibility (JSON-1.0 only).


bootjp:
@claude[agent] review

bootjp:
Fixes pushed in `f17601f9` addressing the three concerns from this round:

1. **Race in Run/Stop** (CI failure on `test (ubuntu-latest)`): the reaper context is now allocated in `NewSQSServer` so a `Stop()` that lands before `Run()` reads a stable cancel func. `go test -race ./adapter/...` is now clean (was flagging `TestSQSServer_StopShutsDown` and any test going through `shutdownNode`).
2. **FIFO send retry must re-read meta** (chatgpt-codex P1, `adapter/sqs_messages.go:387`): on OCC conflict the loop now re-reads meta against the new readTS so a concurrent `PurgeQueue` / `DeleteQueue` / `SetQueueAttributes` cannot let the retry commit under a stale generation. Aborts with `QueueDoesNotExist` if the queue vanished mid-retry.
3. **FIFO batch dedup + sequence** (chatgpt-codex P1, `adapter/sqs_messages_batch.go:164`): `SendMessageBatch` on FIFO queues now routes per-entry through a shared `runFifoSendWithRetry` helper, honoring per-entry dedup window and per-queue monotonic `SequenceNumber`. Standard queues keep the single-OCC fast path. `SequenceNumber` added to the batch result shape; new test `TestSQSServer_SendMessageBatchFifoDedupAndSequence` pins both invariants.

@claude[agent] review

bootjp:
Pushed `04593f44` addressing the three follow-up findings from the Codex review on `f17601f9`:

1. **FIFO send first-attempt meta/readTS mismatch (P1, `adapter/sqs_messages_batch.go:333`)**: `runFifoSendWithRetry` now re-loads queue metadata at the same `readTS` used for the dispatch on every attempt — including attempt 1 — so a concurrent PurgeQueue / DeleteQueue / SetQueueAttributes cannot let a stale `meta` slip past `ReadKeys` on the first try. The (meta, readTS) pair is always coherent.
2. **Reaper missed prior generations (P1, `adapter/sqs_reaper.go:102`)**: the byage scan now spans every generation under the queue. Live-generation entries are gated by retention; prior-generation entries are unconditional orphans, since PurgeQueue / DeleteQueue advance the gen counter without walking the keyspace. Without this, every purge leaked storage permanently. New `parseSqsMsgByAgeKey` decodes the (gen, ts, msg_id) triple from the key.
3. **Expired dedup records (P2, `adapter/sqs_fifo.go:108`)**: `reapExpiredDedup` now sweeps the `!sqs|msg|dedup|...` prefix once per queue per reaper tick and deletes rows whose `ExpiresAtMillis` has passed. The send-side treatment of expired entries is unchanged (still treated as misses); the storage reclamation now runs on the leader.

Two new integration tests pin the contracts:
- `TestSQSServer_RetentionReaperReclaimsPurgedGenerations` — three sends, then PurgeQueue, then one reaper pass; asserts no prior-gen byage rows survive.
- `TestSQSServer_RetentionReaperDropsExpiredFifoDedup` — backdates a dedup record's `ExpiresAtMillis` and asserts it is gone after one reaper pass.

`go test -race ./adapter/...` and `golangci-lint run` both clean.

@claude[agent] review

bootjp:
Pushed `f2a94f22` addressing the three follow-up findings on `04593f44`:

1. **Reaper missed deleted queues (P1, `adapter/sqs_reaper.go:79`)**: `DeleteQueue` now writes a `!sqs|queue|tombstone|<queue><gen>` marker in the same OCC transaction. The reaper enumerates tombstones every tick and reaps the (queue, gen) cohort across byage / dedup / group keyspaces. Once everything is empty, the tombstone itself is deleted so the next tick does not re-walk an empty queue forever.
2. **FIFO retry derived dedup/delay from stale meta (P2, `adapter/sqs_messages.go:393`)**: `runFifoSendWithRetry` now re-derives `dedupID` and `delay` from the fresh meta on every attempt. The (meta, readTS, dedupID, delay) tuple is always coherent, so a `SetQueueAttributes` mid-retry that flips `ContentBasedDeduplication` or `DelaySeconds` cannot let the send commit under the previous generation's rules. Caller-side validation in `sendOneFifoBatchEntry` and `sendMessageFifoLoop` is reduced to cheap pre-flight shape checks.
3. **Batch entry Id grammar (P2, `adapter/sqs_messages_batch.go:565`)**: `validateBatchEntryShape` now rejects entry Ids that don't match `^[a-zA-Z0-9_-]{1,80}$` with `InvalidBatchEntryId` — arbitrary input (whitespace, multi-byte unicode, slashes, oversized strings) no longer slips through to per-entry processing.

New tests:
- `TestSQSServer_ReaperCleansDeletedQueueOrphans` — DeleteQueue + one reaper pass leaves no byage rows and no tombstone.
- `TestSQSServer_SendMessageBatchRejectsInvalidEntryId` — exercises every malformed-Id shape.

`go test -race ./adapter/...` and `golangci-lint run` both clean.

@claude[agent] review

bootjp:
Pushed `701a21a4` (rebased on the new claude-review CI workflow) addressing the three follow-up findings on `f2a94f22`:

1. **SendMessageBatch attribute bytes in size cap (P1, `adapter/sqs_messages_batch.go:91`)**: `totalBatchPayloadBytes` now sums name + DataType + StringValue + BinaryValue across every entry's attributes in addition to the body. The 256 KiB cap can no longer be bypassed by stuffing the data into a BinaryValue.
2. **Self-referential RedrivePolicy (P1, `adapter/sqs_redrive.go:124`)**: rejected at attribute-apply time on both CreateQueue and SetQueueAttributes; defense-in-depth check in `redriveCandidateToDLQ` so even an existing record with a self-pointing policy cannot trigger the loop. Without this, redrive would delete and rewrite the same record with reset receive counters forever.
3. **FIFO batch preflight used stale meta (P2, `adapter/sqs_messages_batch.go:300`)**: every meta-dependent rule (MaximumMessageSize, validateSendFIFOParams, dedup id, delay) is now evaluated inside the per-attempt path via `resolveFreshFifoSnapshot`. The batch preflight keeps only meta-independent shape checks (empty body, MessageAttributes format), so a SetQueueAttributes commit racing the batch cannot fail an entry that the per-attempt path would accept.

New tests:
- `TestSQSServer_SendMessageBatchAttributesContributeToSizeCap` — two entries with 200 KiB BinaryValues + tiny bodies → BatchRequestTooLong.
- `TestSQSServer_RedrivePolicyRejectsSelfReference` — exercises the validator on CreateQueue and SetQueueAttributes.

`go test -race ./adapter/...` and `golangci-lint run` both clean.

@claude review

bootjp:
Pushed `88cc76da` — comment-only fix for the one minor observation in your review (referenced `reapDeletedQueueOrphans` instead of `reapTombstonedQueues`).

Thanks for the thorough review. With your "ready to merge" verdict and all CI green, the loop has reached its exit condition (review OK). I'll stop the periodic check now; please merge at your convenience.

@claude review

bootjp:
Pushed `e6e0d0b8` (rebased on `f7fe55d4`) addressing all five findings from the latest Codex review:

1. **ReceiveMessage paging past skip-heavy candidates (P1, `adapter/sqs_messages.go:699`)**: scan and rotate are now interleaved. `scanAndDeliverOnce` pages the visibility index, delivers what survives the per-candidate filters (FIFO group lock, retention expiry, DLQ redrive), and pages forward until `opts.Max` is reached, the index is drained, or a 100 ms wall-clock budget elapses. Hot-FIFO-group fan-in and poison-message backlogs no longer cause false-empty responses.
2. **Batch missing-queue → request-level error (P2, `adapter/sqs_messages_batch.go:467` and `:559`)**: both `DeleteMessageBatch` and `ChangeMessageVisibilityBatch` now do an upfront `requireQueueExists` check; a missing queue surfaces as HTTP 400 `QueueDoesNotExist`, not per-entry failures inside a 200 envelope.
3. **FIFO DLQ rejects empty-MessageGroupId redrive (P2, `adapter/sqs_redrive.go:140`)**: when the DLQ is FIFO and the source record has no `MessageGroupId`, `redriveCandidateToDLQ` refuses the move with `InvalidAttributeValue`. Without this gate, Standard-source → FIFO-DLQ redrives would land records the FIFO receive path treats as non-FIFO, bypassing group-lock semantics.
4. **CreateQueue tag cap (P2, `adapter/sqs_catalog.go:520`)**: `CreateQueue` now applies the same 50-tag cap that `TagQueue` enforces; a caller can no longer bypass the post-creation gate by stuffing tags into the initial create.

New tests:
- `TestSQSServer_ReceivePagesPastFifoGroupLockSkips` — 30 msgs in one FIFO group block the head; second group's head still surfaces and the second receive correctly returns 0.
- `TestSQSServer_CreateQueueRejectsTooManyTags` — 60-tag create → HTTP 400.
- `TestSQSServer_BatchOnMissingQueueIsRequestLevelError` — both batch APIs return request-level error on a missing queue.
- `TestSQSServer_RedrivePolicyFifoDlqRejectsStandardSource` — Standard source → FIFO DLQ does not produce DLQ records.

`go test -race ./adapter/...` and `golangci-lint run` both clean.

@claude review

bootjp:
Thanks for the thorough re-review on `e6e0d0b8`. With your "Ready to merge" verdict and all CI green, the loop has reached its exit condition (review OK). I'll stop the periodic check now.

The two observations you flagged (source-message liveness assertion in the FIFO-DLQ rejection test, and the inherent TOCTOU on the batch queue-exists check) were both noted as low-priority / not blocking, so I'll leave them as follow-ups unless you'd prefer them addressed before merge — happy to do either.

Please merge at your convenience.

bootjp:
@claude review

bootjp:
Pushed `31a083f3` addressing the new P1 from your review of `e6e0d0b`:

**Storage orphan on Purge → Delete races** (`adapter/sqs_purge.go:81`): `tryPurgeQueueOnce` now writes a `sqsQueueTombstoneKey(queueName, lastGen)` entry in the same OCC OperationGroup that bumps the generation counter. Any pre-bump cohort is now reachable by `reapTombstonedQueues` even if `DeleteQueue` lands before the next reaper tick. The fence (StartTS + ReadKeys on metaKey/genKey) is unchanged; the new Put rides on the existing transaction so we still get all-or-nothing semantics.

**Regression test** (`adapter/sqs_extra_test.go: TestSQSServer_PurgeThenDeleteOrphansCleaned`) drives the exact Purge → Delete → Reaper sequence Lens 5 called out as missing. It fails on the prior code with 3 leaked byage entries and passes after the fix. Sibling tests (`RetentionReaperReclaimsPurgedGenerations`, `ReaperCleansDeletedQueueOrphans`) stay green — the new tombstone is drained by the existing live-queue reap path before `reapTombstonedQueues` runs, so neither test sees a leftover marker.

`go test -race -run "TestSQS" ./adapter/` is clean, lint passes (extracted three helper functions to keep the new test under the cyclop=10 ceiling per project rules — no `//nolint`).

The two prior low-priority observations (FIFO-DLQ rejection liveness assertion, batch queue-exists TOCTOU) are still left as follow-ups per the previous round's agreement.

@claude review

bootjp:
Pushed `9e8148c1` — comment-only fix for the minor observation in your review (`SqsQueueTombstonePrefix` doc now records both DeleteQueue and PurgeQueue as tombstone writers, matching the actual invariant).

With your "Ready to merge" verdict on `31a083f3` and CI green, the loop has reached its exit condition (review OK). I'll stop the periodic check after this round; please merge at your convenience.

@claude review

bootjp:
Pushed `cdb3c87d` addressing the new P2 from the latest Codex review on `31a083f3`:

**FIFO -> Standard redrive direction now rejected** (`adapter/sqs_redrive.go`): the existing guard only rejected Standard -> FIFO. The symmetric mismatch silently moved a FIFO source record (carrying `MessageGroupId`) into a Standard DLQ, after which `tryDeliverCandidate` enabled FIFO group-lock semantics on a queue meant to behave as Standard.

`redriveCandidateToDLQ` now loads the source meta in addition to the DLQ meta and requires `srcMeta.IsFIFO == dlqMeta.IsFIFO` before building the OCC dispatch. The existing `dlqMeta.IsFIFO && srcRec.MessageGroupId == ""` rail stays as defense-in-depth for malformed FIFO -> FIFO records that would slip past the type-equality check. Validation extracted into `validateRedriveTargets` to keep `redriveCandidateToDLQ` under cyclop=10 per project rules — no `//nolint`.

**Regression test** (`adapter/sqs_extra_test.go: TestSQSServer_RedrivePolicyStandardDlqRejectsFifoSource`) mirrors the existing Standard -> FIFO test in the opposite direction. On the prior code it failed by observing the Standard DLQ holding a record with `MessageGroupId:g1`; after the fix the DLQ stays empty.

`go test -race` on every `Redrive*` / `DLQ*` test green; lint clean.

@claude review

bootjp:
Thanks for the thorough re-review on `cdb3c87d`. With your second consecutive "Ready to merge" verdict (after `9e8148c1` and now `cdb3c87d`), all P1/P2 findings resolved across the entire round, and CI green on every shipped check, the loop has reached its exit condition. Stopping the periodic check now; please merge at your convenience.

The pre-existing minor gap you noted (source-message liveness assertion in the redrive rejection tests) remains as a low-priority follow-up per the earlier round's agreement.

bootjp:
@claude review

bootjp:
Round-N Codex 2 件 (P1 + P2) を 0bd04e92 で対応。テスト 2 本付き。

| Severity | Location | Issue | Fix |
|---|---|---|---|
| **P1** | `sqs_redrive.go:265-272, 213` | FIFO→FIFO redrive で `SequenceNumber=0` + DLQ seq counter 未更新 → 後続の normal FIFO send が低い seq を取り、AWS の monotonic 不変条件を破る | `redriveCandidateToDLQ` で `dlqMeta.IsFIFO` のとき `loadFifoSequence(DLQ, readTS)+1` を計算、`buildDLQRecord` (record に stamp) と `buildRedriveOps` (seq Put + ReadKeys に `sqsQueueSeqKey(DLQ)` 追加) に thread。Standard DLQ は `dlqSeq=0` 通過で挙動不変 |
| **P2** | `sqs_messages.go:1523-1527` | Binary attribute が `StringValue` 同時送信を reject しない → 不正 byte 列が record に永続化されて ReceiveMessage で MD5 不一致 | `validateMessageAttributeValuePair` を抽出し、Binary は `StringValue != ""` を reject、対称的に String/Number も `len(BinaryValue) > 0` を reject。split で cyclop budget も維持 |

## Tests

- `TestSQSServer_FifoFifoRedriveAssignsSequenceNumber`: FIFO src→FIFO DLQ で redrive、DLQ message の `SequenceNumber > 0`、後続 FIFO send が `redrivenSeq + 1` (monotonic 継続) を取ることを assert
- `TestSQSServer_SendMessageRejectsBinaryWithStringValue`: 3 ケース (Binary+StringValue, String+BinaryValue, Number+BinaryValue) すべて HTTP 400 / InvalidAttributeValue

1. **Data loss**: なし。body / attributes 不変、SequenceNumber と DLQ counter のみ変更。
2. **Concurrency**: `sqsQueueSeqKey(DLQ)` を ReadKeys に追加したので、並行 FIFO→FIFO redrive は OCC で write-write-conflict serialize される (concurrent FIFO send と同じ挙動)。
3. **Performance**: FIFO redrive 1 回あたり `loadFifoSequence` 1 回追加。`validateRedriveTargets` 既存読み込みに比べて無視できる。
4. **Data consistency**: DLQ の SequenceNumber が redrive / normal-send 両経路で monotonic、同一 queue 上で seq 衝突しない。
5. **Test coverage**: 両 fix とも共置された regression test。既存 FIFO/redrive テスト群 (`FifoSequenceNumberMonotonic`, `DLQRedriveOnMaxReceiveCount`, `RedrivePolicy*`) 全 pass。

@claude review

PR #644, opened 2026-04-25
admin: AdminForward follower-side client + handler integration (P1, partial)

bootjp:
Phase 1 + 2 of Task #26: the follower-side `LeaderForwarder` client and its integration into the dynamo HTTP handler. Builds on the AdminForward leader-side dispatcher landed via #635.

## Summary

- New `LeaderForwarder` interface (`internal/admin/forward_client.go`) decouples the dynamo HTTP handler from `pb.AdminForwardClient`. The handler stays proto-free; the bridge in `main_admin.go` (next phase) plugs in the gRPC-backed implementation.
- `gRPCForwardClient` translates a `CreateTableRequest` / table-name into an `AdminForwardRequest`, dials via a `GRPCConnFactory` (production wraps `kv.GRPCConnCache`), and re-shapes the response into `ForwardResult` (status, payload, content-type).
- `forwarded_from = nodeID` is populated so the leader's audit log carries the trace (criterion 6, leader-side already shipped in #635).
- Defensive: `status_code == 0` upgrades to `502 Bad Gateway`; missing `ContentType` fills the JSON default. Both surface transport bugs rather than producing silently-malformed SPA responses.
- `ErrLeaderUnavailable` sentinel signals the "no leader known" case so the handler can map to 503 + `Retry-After: 1` (criterion 3).
- `DynamoHandler` gains a `forwarder` field and `WithLeaderForwarder` option. When set, `handleCreate` / `handleDelete` catch `ErrTablesNotLeader` from the source and forward to the leader transparently — the SPA cannot tell forwarded from leader-direct.
- `writeForwardResult` re-emits the leader's structured response verbatim (status + payload + content-type), so a forwarded `409 Conflict` from the leader stays `409` on the wire — no re-classification.
- `writeForwardFailure` maps `ErrLeaderUnavailable` (election in flight) and gRPC transport errors to 503 + `Retry-After: 1`. `ErrLeaderUnavailable` is intentionally NOT logged at error level (elections are routine); transport errors are logged at LevelError so operators can investigate.

## What is NOT in this PR

- gRPC server registration in `main.go` (production wiring of the `ForwardServer` from #635) — comes in the next phase.
- The bridge that wraps `kv.GRPCConnCache` and supplies `LeaderAddressResolver` — same phase.
- Election-period retry-loop on the client side (criterion 3 partial: this PR returns 503 + Retry-After; the SPA / client retries the request; criterion 3 fully needs the production bridge to dial actual leader-discovery).

## Test plan

[...]

PR #635, opened 2026-04-25
admin: AdminForward RPC + leader-side dispatcher (P1, partial)

bootjp:
Stacked on #634. Lays down the proto contract and the leader-side dispatcher for the follower-to-leader forwarding path mandated by design Section 3.3. The follower-side client + bridge wiring lands in a separate PR to keep this one reviewable.

## Summary

- New `proto/admin_forward.proto` with `service AdminForward` carrying an `AdminPrincipal` + opaque JSON payload + the operation enum. Operation-agnostic so adding a new admin endpoint does not require a wire-format bump.
- `internal/admin/forward_server.go` is the leader-side gRPC handler. `validatePrincipal` re-evaluates the inbound access-key against the leader's own `RoleStore` and demotes a follower-claimed full role when the leader sees read-only — this is the design's *adapter side is the source of truth for authz* invariant for forwarded calls (Section 3.3.1).
- `forwardErrorResponse` mirrors the HTTP handler's `writeTablesError` mapping (403 / 503 / 404 / 409 / 400 / 500) so a forwarded call is indistinguishable from a leader-direct call to the SPA. Raw adapter error text never leaves the leader; everything is encoded into the structured response payload.
- `forwarded_from` is echoed into the leader's audit log line on every successful forwarded operation (acceptance criterion 6).

## What is NOT in this PR

- Follower-side AdminForward client + bridge wiring (acceptance criterion 2). The follower's HTTP handler still returns `503 leader_unavailable` + `Retry-After: 1` from #634 — this PR is a no-op on the request path until the client wiring lands.
- Election-period 503 handling (criterion 3) — comes with the client wiring.
- Rolling-upgrade compatibility flag `admin.leader_forward_v2` (criterion 5) — depends on a Raft-level cluster version bump that is independently substantial.
- gRPC server registration in `main.go` — wired in the next PR alongside the client.

## Acceptance criteria coverage

| # | Criterion | This PR |
|---|---|---|
| 1 | Leader direct write | ✓ existing (PR #634) |
| 2 | Follower forwards transparently | ⏳ next PR (client) |
| 3 | Election period 503 + retry | ⏳ next PR (client) |
| 4 | Leader demotes stale full role | ✓ test added |
| 5 | Rolling-upgrade compat flag | ⏳ deferred |
| 6 | `forwarded_from` in audit log | ✓ test added |

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run` (admin package: 0 issues)
[...]

PR #648, opened 2026-04-26
admin: AdminForward Phase 3 — register gRPC server + wire follower-side bridge

bootjp:
Phase 3 of Task #26 / AdminForward design 3.3. Closes the wiring loop opened by #635 (leader-side dispatcher) and #644 (follower-side client + handler integration): the admin HTTP DynamoHandler now hands `ErrTablesNotLeader` writes off to the in-process `LeaderForwarder`, and the leader-side gRPC `AdminForward` service is registered on every node so the forwarded RPC actually lands somewhere.

## Summary

- **`admin.ServerDeps.Forwarder`** (3315bf43) — new optional field on the high-level admin server composition surface. `admin.NewServer` calls `WithLeaderForwarder` when set; nil leaves the existing 503 + Retry-After:1 fallback untouched. Single-node / leader-only deployments are unaffected.
- **`main_admin_forward.go`** (f3e529c1) — production bridge between `kv.GRPCConnCache` and the admin layer's typed `PBAdminForwardClient` (via `pb.NewAdminForwardClient`). Includes `buildLeaderForwarder`, `roleStoreFromFlags`, `adminForwardServerDeps`, and a `readyForRegistration` gate.
- **Runtime wiring** (19f6b70a) — `runtimeServerRunner.start()` now creates `dynamoServer` BEFORE `startRaftServers`, so the leader-side `ForwardServer` can use it as its `TablesSource`. `startRaftServers` accepts the new `adminForwardServerDeps` bundle and delegates to `registerAdminForwardServer`. `startServers` constructs a single `*kv.GRPCConnCache` and forwards it (with the `Coordinator`) to `startAdminFromFlags`, which in turn builds the production `LeaderForwarder` via `buildAdminLeaderForwarder` and passes it through `ServerDeps.Forwarder`.

## What is NOT in this PR

- Rolling-upgrade compatibility flag (criterion 5) — still deferred behind a cluster-version bump.
- Election-period retry loop on the follower's bridge — the handler returns 503 + Retry-After:1, the SPA / curl re-issues, and the bridge dials again on the next attempt. Inline retry inside the handler would just hide the latency from operators.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `golangci-lint run` (main + admin packages: 0 issues)
- [x] `go test ./internal/admin/ -count=1 -race` — full admin suite passes (existing 21 forward tests + 1 new `TestServer_ServerDepsForwarderIsWired`)
- [x] `go test . -count=1 -race` — main package passes (4 new bridge / role-store / readyForRegistration tests)
- [ ] `adapter` package times out at 240s on this branch AND on `main` — pre-existing flake unrelated to this PR (verified by checking out main and running the same suite)
- [ ] End-to-end smoke against a 3-node cluster — needs the local Jepsen runner; blocked on the next PR or a manual run

## Acceptance criteria coverage

| # | Criterion | This PR |
|---|---|---|
| 1 | Leader direct write | ✓ (in main since #634) |
| 2 | Follower forwards transparently | ✓ — wiring complete; both leader-side gRPC service AND follower-side bridge are now plumbed |
| 3 | Election-period 503 + retry | ✓ — handler still returns 503 + Retry-After:1 when no leader is known; the SPA / client retries and the next attempt re-dials |
| 4 | Leader demotes stale full role | ✓ (in main since #635) |
| 5 | Rolling-upgrade compat flag | ⏳ deferred (cluster-version bump) |
| 6 | `forwarded_from` in audit log | ✓ (in main since #635 — bridge now stamps `--raftId` onto every forwarded request) |

## Self-review (5 lenses)

1. **Data loss**: No FSM / Raft / Pebble path changes; admin writes still go through the same `AdminCreateTable` / `AdminDeleteTable` Raft proposal once they reach the leader.
2. **Concurrency**: `runner.start()` reorder confirmed safe — `startDynamoDBServer` only depends on `coordinate` + `shardStore` (already constructed); raft TCP listeners are independent of the dynamo HTTP listener. Worst case during the 100 ms-or-so reorder window is the same "no leader yet" 503 the old order would have produced if a request landed before raft converged.
3. **Performance**: One additional `&kv.GRPCConnCache{}` allocation per process; one closure for the resolver. No hot-path changes.
4. **Data consistency**: Forwarded requests re-validate the principal at the leader (criterion 4, already shipped); commit-ts ordering is unchanged because the leader's `AdminCreateTable` path does its own `HLC.Next`.
5. **Test coverage**: 4 new main-package tests + 1 new admin-package server-level test. `TestServer_ServerDepsForwarderIsWired` is the regression test for "future refactor drops the forwarder before it reaches the dynamo handler" — exactly the failure mode the wiring change in this PR introduces the risk of.

## Summary by CodeRabbit

* **New Features**
  * Added leader-forwarding capability for admin requests sent to follower nodes
  * Implemented role-based access control for admin API operations
  * Enhanced DynamoDB request routing to automatically direct operations to leader nodes

* **Tests**
  * Added comprehensive integration and unit tests validating leader-forwarding paths

### DIFF
--- a/internal/admin/server.go
+++ b/internal/admin/server.go
@@ -39,6 +39,14 @@ type ServerDeps struct {
 	// cluster page deploy without standing up the dynamo bridge.
 	Tables TablesSource
 
+	// Forwarder is the LeaderForwarder that the Dynamo handler hands
+	// off ErrTablesNotLeader writes to (design 3.3, AdminForward).
+	// Optional: a nil value disables follower→leader forwarding, in
+	// which case the handler surfaces 503 + Retry-After: 1 directly.
+	// Single-node and leader-only deployments leave this nil; multi-
+	// node clusters wire the production gRPC client.
+	Forwarder LeaderForwarder
+
 	// StaticFS is the embed.FS (or any fs.FS) backing the SPA. May be
 	// nil during early development; the router renders 404 for
 	// /admin/assets/* and the SPA fallback in that case.
@@ -110,9 +118,13 @@ func NewServer(deps ServerDeps) (*Server, error) {
 		// operators must restart the listener for revocation to
 		// take effect, but the JWT no longer extends a revoked
 		// key past the next request.
-		dynamo = NewDynamoHandler(deps.Tables).
+		dynamoHandler := NewDynamoHandler(deps.Tables).
 			WithLogger(logger).
 			WithRoleStore(MapRoleStore(deps.Roles))
+		if deps.Forwarder != nil {
+			dynamoHandler = dynamoHandler.WithLeaderForwarder(deps.Forwarder)
+		}
+		dynamo = dynamoHandler
 	}
 	mux := buildAPIMux(auth, deps.Verifier, cluster, dynamo, logger)
 	router := NewRouter(mux, deps.StaticFS)
--- a/internal/admin/server_test.go
+++ b/internal/admin/server_test.go
@@ -453,6 +453,57 @@ func TestServer_DynamoDeleteTable_FullRoleHappyPath(t *testing.T) {
 	_ = resp.Body.Close()
 }
 
+// TestServer_ServerDepsForwarderIsWired confirms that a non-nil
+// ServerDeps.Forwarder reaches the DynamoHandler so a follower-side
+// CreateTable that the source returns ErrTablesNotLeader for is
+// transparently forwarded. Without the wire, the request would 503 +
+// Retry-After:1 — the no-forwarder fallback path. With it, the
+// leader's response status (here, 201) is replayed verbatim.
+func TestServer_ServerDepsForwarderIsWired(t *testing.T) {
+	src := &notLeaderSource{}
+	fwd := &stubLeaderForwarder{createRes: &ForwardResult{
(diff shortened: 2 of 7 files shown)


bootjp:
@claude review

bootjp:
@claude review

964ecd42 addresses all four findings from the previous review pass:

- **Codex P1 / CodeRabbit Major**: `roleStore` + `connCache` are now built only when `*adminEnabled` is true. With admin disabled, `forwardDeps.roles` stays nil and `registerAdminForwardServer`'s gate skips the gRPC bind entirely — the AdminForward service is no longer reachable when the HTTP admin surface is off.
- **Claude Issue 1 / CodeRabbit Minor**: shutdown goroutine drains `connCache` via `Close()` on context cancellation; gRPC connections no longer leak.
- **Claude Issue 2**: dropped `//nolint:nilnil` (the linter isn't enabled and CLAUDE.md disallows nolint directives); replaced the inline justification with a clear comment.
- **Claude Issue 3**: `TestBuildAdminLeaderForwarder_NilGateReturnsNoForwarder` covers all three nil combinations + the empty-nodeID error path.

PR #649, opened 2026-04-26
feat(admin): React + Vite SPA for the admin dashboard (P3)

bootjp:
## Summary

Phase **P3** of [docs/design/2026_04_24_proposed_admin_dashboard.md](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_04_24_proposed_admin_dashboard.md). The Go side (auth / JWT / cluster handler / dynamo handler / follower→leader forwarding / `embed.go` skeleton) shipped in earlier PRs (#644 et al.) but `StaticFS` was wired as `nil`, so `/admin/*` answered 404 for any non-API path. This PR ships the React bundle and flips `StaticFS` on.

- React 18 + TypeScript + Vite, Tailwind CSS for tokens (Section 5.1). No UI framework dependency — bundle stays at **~60 KB gzip** vs the doc's 150 KB ceiling (Section 7.2).
- `vite.config.ts` writes its build output straight into `internal/admin/dist/` so `npm run build && go build` produces a single-binary release.
- Five pages per Section 5.2: Login, Dashboard, DynamoDB list/detail (with Create + Delete dialogs), S3 list/detail (placeholder — endpoints not yet wired, the SPA renders a soft "endpoint pending" notice on the 404).
- Cookie-based session (Section 6): `admin_session` (HttpOnly) handled by the browser; `admin_csrf` value echoed in `X-Admin-CSRF` for every POST/PUT/DELETE per the double-submit pattern. Read-only sessions hide write affordances; backend re-evaluates the role.
- `.gitignore` keeps the placeholder `internal/admin/dist/index.html` committed (so `//go:embed all:dist` always has a match in a fresh clone) and excludes the hashed Vite asset outputs and npm/vite caches — committed bundles drift from source.

## Test plan

- [x] `cd web/admin && npm install && npm run build` — 192 KB JS / 60 KB gzip + 14 KB CSS
- [x] `cd web/admin && npm run lint` — `tsc --strict --noUnusedLocals --noUncheckedSideEffectImports`
- [x] `go build ./...` — clean (placeholder + built bundle both compile)
- [x] `go test -race -count=1 ./internal/admin/...`
- [x] `golangci-lint run ./internal/admin/...`
[...]

PR #650, opened 2026-04-26
fix(admin/spa): drop redundant abort guard and add Modal a11y

bootjp:
## Summary

Two follow-ups from the [#649](https://github.com/bootjp/elastickv/pull/649) review that were intentionally deferred to keep that PR focused on the functional bugs.

**Stacked on `feat/admin-dashboard-spa`** (#649). Once #649 merges, this rebases cleanly onto `main`.

### 1. `useApi.ts` — drop the redundant abort guard

The `catch` handler had two checks back-to-back:

```ts
if (cancelled) return;
if (ctrl.signal.aborted) return;  // ← unreachable
```

The cleanup function sets `cancelled = true` *before* calling `ctrl.abort()`, so any abort path is already covered by the first check. Dropped the second guard and recorded *why* one check is enough in a comment.

### 2. `Modal.tsx` — full WAI-ARIA dialog contract + focus trap

Added:

- `role="dialog"`, `aria-modal="true"`, `aria-labelledby` on the dialog root (uses `useId()` for the title id).
- Focus management:
  - On open: store the previously-focused element, focus the first focusable inside the dialog.
  - Tab / Shift+Tab cycle within the dialog instead of escaping to the page underneath.
  - On close: restore focus to whoever opened the dialog (skipped if that element was unmounted by the dialog action — e.g. the dialog deleted the row whose button opened it).
  - Empty-dialog fallback focuses the dialog container itself with `tabindex=-1`.

Vanilla DOM implementation — `focus-trap-react` / `react-focus-lock` would have been 5–7 KB gzip for what is a ~40 line solution.

**Bundle delta**: +0.4 KB gzip (60.23 → 60.59 KB JS, 14.55 → 14.57 KB CSS).

## Test plan

- [x] `cd web/admin && npm install && npm run build`
- [x] `cd web/admin && npm run lint` — `tsc --strict`
- [x] `go build ./...`
- [ ] Manual smoke (after #649 merges and the dashboard is visitable):
  - Open Create Table / Delete Table dialog → focus lands inside the dialog.
  - Tab past the last button → wraps back to the close button.
  - Shift+Tab from the close button → wraps to the last form field.
  - Escape closes (when not `busy`) and focus returns to the trigger button.
  - Screen reader announces the dialog name from the title text.

### DIFF
--- a/web/admin/src/components/Modal.tsx
+++ b/web/admin/src/components/Modal.tsx
@@ -30,16 +30,41 @@ export function Modal({ title, open, onClose, children, busy }: ModalProps) {
   const previouslyFocusedRef = useRef<HTMLElement | null>(null);
   const titleId = useId();
 
+  // Focus capture / restore is tied to `open` ONLY. Folding `busy`
+  // into the same effect as the previous version did caused the
+  // cleanup to fire on every busy toggle (e.g. the user clicks Save
+  // and the parent flips busy=true): focus would briefly leave the
+  // dialog, previouslyFocusedRef would get clobbered with the
+  // trigger button, then the next run would snap focus back. Real
+  // bug for screen-reader users — the dialog became "exited" mid-
+  // operation. Splitting the effects keeps each invariant tied to
+  // its own state slice (Claude review on #650).
   useEffect(() => {
     if (!open) return;
-
     previouslyFocusedRef.current = (document.activeElement as HTMLElement | null) ?? null;
-    // Focus the first focusable element so keyboard users land
-    // inside the dialog instead of the page underneath. Run on the
-    // next tick so the dialog DOM exists and any autofocus has had
-    // a chance to settle.
+    // queueMicrotask defers focus until after sibling useEffect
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
@claude review

685f410f addresses the Modal focus-management bug from your previous review. The single `useEffect` is now split into two:

- **Focus capture/restore** — deps `[open]` only. `previouslyFocusedRef` stays valid across the whole dialog lifetime; no clobber on `busy` toggle.
- **Keyboard handler** — deps `[open, busy, onClose]`. Re-binds on those changes; only side effect is one window listener swap.

The dist/ placeholder `index.html` is intentionally NOT regenerated (per .gitignore: assets are built at deploy time, not committed).

bootjp:
@claude review

Rebased onto main (086480c7) now that #649 is merged. Dropped the 2 commits that landed via #649's squash merge; kept the 2 commits unique to this PR (45691918 abort-guard cleanup + Modal a11y, 086480c7 effect-split focus fix). Build passes.

PR #661, opened 2026-04-26
feat(kv): observe lease + linearizable reads via keyviz Sampler

bootjp:
## Summary

Wires read sampling into the shard coordinator's two key-routed read entry points so the heatmap's `Reads` / `ReadBytes` series get populated (design §5.1: "Reads are sampled by the node that actually serves the read"). Mirrors the write-side `observeMutation` wiring; both feed the same `keyviz.Sampler` the coordinator already holds.

- New `observeReadKey` helper, called from `LinearizableReadForKey` and `LeaseReadForKey` after the route is resolved.
- `valueLen` is always 0 at this layer — the consistency check doesn't fetch data; the actual `MVCCStore.GetAt` happens further down the stack and is sampled separately. Adapter direct-read paths (Redis / DynamoDB / S3 hitting `MVCCStore.GetAt` without going through the coordinator) still bypass keyviz; sampling those is a future slice — see the `observeReadKey` doc comment.
- Leadership-check entries (`IsLeaderForKey`, `VerifyLeaderForKey`, `RaftLeaderForKey`) intentionally do NOT call `observeReadKey` — they're internal routing checks, not user-facing data reads.

Closes the read-side gap noted on the `observeMutation` Phase-2 TODO. The matrix's `Reads` and `ReadBytes` series are no longer permanently zero.

## Test plan

- [x] `TestShardedCoordinatorObservesLeaseAndLinearizableReads` — drives one of each through a real raft + store, asserts exactly one `Observe(routeID, OpRead, len(key), 0)` per call.
- [x] `TestShardedCoordinatorSkipsObserveForLeadershipChecks` — pins the negative contract that the three leadership checks stay silent.
- [x] `go test -race -count=1 -run TestShardedCoordinator ./kv/...` clean.
- [x] `golangci-lint run ./kv/...` clean.

### DIFF
--- a/kv/sharded_coordinator.go
+++ b/kv/sharded_coordinator.go
@@ -750,11 +750,11 @@ func (c *ShardedCoordinator) RaftLeaderForKey(key []byte) string {
 }
 
 func (c *ShardedCoordinator) LinearizableReadForKey(ctx context.Context, key []byte) (uint64, error) {
-	g, ok := c.groupForKey(key)
+	routeID, g, ok := c.routeAndGroupForKey(key)
 	if !ok {
 		return 0, errors.WithStack(ErrLeaderNotFound)
 	}
-	c.observeReadKey(key)
+	c.observeRead(routeID, len(key))
 	return linearizableReadEngineCtx(ctx, engineForGroup(g))
 }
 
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
@claude review

PR #659, opened 2026-04-26
feat(sqs): Phase 3 — Approximate counters + admin SPA queues pages

bootjp:
## Summary

Phase 3.A and 3.E of [`docs/design/2026_04_24_partial_sqs_compatible_adapter.md`](https://github.com/bootjp/elastickv/blob/feat/sqs-phase3-counters-console/docs/design/2026_04_24_partial_sqs_compatible_adapter.md) (renamed from `_proposed_` in this PR — Phase 1+2 already landed). Phase 3.B / 3.C / 3.D are recorded as TODOs in §16.4 / §16.5 / §16.6 of the same doc with the rough scope each needs (XML query protocol, per-queue throttling, split-queue FIFO).

### 3.A — Approximate counters on `GetQueueAttributes`

`GetQueueAttributes` now returns `ApproximateNumberOfMessages`, `ApproximateNumberOfMessagesNotVisible`, and `ApproximateNumberOfMessagesDelayed` (previously: none of the three). Implementation walks the visibility index at the read TS and bucket-classifies each entry by `(visible_at, available_at)` vs `now`:

- **Visible** = `available_at ≤ now ∧ visible_at ≤ now` — ready for the next `ReceiveMessage`.
- **NotVisible** = `available_at ≤ now ∧ visible_at > now` — in flight after a receive bumped vis.
- **Delayed** = `available_at > now` — sent with `DelaySeconds`, not yet eligible.

Bounded at **5,000 vis-index entries per call** so a runaway queue cannot turn `GetQueueAttributes` into an O(stream) scan; over the budget the counts are reported as a lower bound and `Truncated=true` so callers can log. The counters fire **only when the caller asks for them or for `All`** — config-only callers still pay zero scan cost (matches AWS's "approximate" contract and keeps the existing `GetQueueAttributes` callers cheap).

### 3.E — Admin SPA queues pages (Section 13 superseded)

The original Section 13 console UI design (separate `--consoleAddress` listener, bearer token, dedicated `go:embed` bundle) is **superseded** — see §16.2 of the design doc for the rationale. Operator UI for SQS lives on the existing admin dashboard SPA (#649) instead, reusing its JWT cookie session, CSRF double-submit, role model, and embed.FS.

Backend (`adapter/sqs_admin.go` + `internal/admin/sqs_handler.go`):
[...]

PR #662, opened 2026-04-26
feat(sqs): Phase 3.B — XML query-protocol support (3-verb proof)

bootjp:
## Summary

Phase **3.B** of [`docs/design/2026_04_24_partial_sqs_compatible_adapter.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_04_24_partial_sqs_compatible_adapter.md) §16.4. Adds the AWS SQS **query protocol** (form-encoded request, XML response) alongside the existing JSON protocol on the same listener — older `aws-sdk-java` v1, `boto < 1.34`, and AWS CLI clients can now talk to elastickv without modification. Detection happens per-request via `Content-Type` / `X-Amz-Target` / `Action` presence; no flag, no separate port. See the new proposal doc committed alongside.

This is an **architectural proof PR**: dispatch + decoding + encoding + error envelope + the `*Core` refactor pattern are all in place, with **three verbs** wired end-to-end as concrete proof. Each follow-up verb is a parser + response struct + one switch arm — no further design work needed.

### Verbs in this PR

| Verb | Why it's in the proof set |
|---|---|
| `CreateQueue` | Exercises the `Attribute.N.{Name,Value}` indexed-collection parser. |
| `ListQueues` | Exercises the repeated-element XML response shape. |
| `GetQueueUrl` | Exercises the `<ErrorResponse>` envelope path via `QueueDoesNotExist`. |

Every other verb returns a structured **501 `NotImplementedYet`** XML envelope so operators see the gap explicitly. `SendMessage` / `ReceiveMessage` / `DeleteMessage` are the highest-priority follow-ups (they need the same `*Core` refactor on the FIFO send loop).

### Key design points

- **No new listener / no flag.** `pickSqsProtocol(*http.Request)` decides per request. JSON and Query share the SQS port and the SigV4 path.
- **Wire-format-free cores.** `createQueue` / `listQueues` / `getQueueUrl` are now `decode → core → encode` with `core(ctx, in) (out, error)`. The JSON wrappers are unchanged in behavior; existing JSON tests pass without modification.
- **DoS protection inherited.** Body read is bounded by the same `sqsMaxRequestBodyBytes` the JSON path uses.
- **SigV4 unchanged.** The signed canonical request includes the form-encoded body, so the existing SigV4 middleware verifies query requests without code changes.
- **Error parity.** `<Code>` reuses the existing `sqsErr*` constants. HTTP status mirrors what the JSON path returns, so SDK retry classifiers work across protocols.
- **Cyclomatic budget honoured.** `handle()` was refactored to extract `handleQueryProtocol` — `cyclop ≤ 10` per project rules, no `//nolint`.

### Known limitation (design §11.4)

`proxyToLeader`'s error writer always emits the JSON envelope, so a query-protocol client hitting a follower during a leader flip sees one JSON error before retry lands on the new leader. Follow-up PR threads the detected protocol onto the request context so the proxy emits matching XML.

## Test plan

- [x] `go test -count=1 -race -run "TestSQS|QueryProtocol|TestPickSqs|TestCollectIndexedKV|TestWriteSQSQueryError" ./adapter/` — passes
- [x] `golangci-lint run ./adapter/...` — `0 issues.`
- [x] `pickSqsProtocol` table tests cover documented edge cases (header precedence, charset suffix, GET with Action, missing Action, nil request).
- [x] `collectIndexedKVPairs` tests cover happy path, orphan Name, empty input, unrelated prefix.
- [x] End-to-end via the in-process listener: CreateQueue / ListQueues / GetQueueUrl round-trip on the query side.
- [x] **Cross-protocol parity**: a queue created via Query is visible via JSON `GetQueueUrl` with the same URL.
- [x] Error envelope: 4xx maps to `<Type>Sender</Type>`, 5xx to `<Type>Receiver</Type>`, namespace pinned, `x-amzn-ErrorType` header set.
- [x] Unknown verb returns 501 with the `NotImplementedYet` XML envelope.
- [x] Missing `Action` parameter returns 400 (per design §3).

## Self-review (5 lenses)

1. **Data loss** — Wire-format change only. Cores are byte-for-byte identical to the previous handler bodies; no Raft / OCC / MVCC code is touched.
2. **Concurrency** — No new shared state. Detection is request-local. Body parsing is bounded.
3. **Performance** — One additional `Content-Type` string compare per request on the dispatch hot path. Negligible.
4. **Data consistency** — `*Core` returns the same business-logic outputs as before; the JSON tests are the regression net for parity. Cross-protocol parity test pins behaviour.
5. **Test coverage** — 10 new test cases cover detection, parsing, envelope shape, and three end-to-end verbs. Existing `TestSQS*` race suite passes on the refactor.

## Stacking

This PR is **independent** — branched from current `main` (which has #638 + #649 merged). It does not depend on PR #650 / PR #659. Merge whenever ready.

### DIFF
--- a/adapter/sqs.go
+++ b/adapter/sqs.go
@@ -154,6 +154,20 @@ func (s *SQSServer) handle(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// pickSqsProtocol decides between the JSON path (X-Amz-Target +
+	// JSON body, the existing default) and the query path (form-
+	// encoded body, XML response) on a per-request basis. See
+	// docs/design/2026_04_26_proposed_sqs_query_protocol.md for the
+	// detection rules.
+	if pickSqsProtocol(r) == sqsProtocolQuery {
+		s.handleQueryProtocol(w, r)
+		return
+	}
+	// JSON / Unknown both fall through to the JSON path: the JSON-
+	// style 400 is the most informative error for a client that
+	// has not picked a codec yet (§3 of the design doc). The
+	// dispatch table below stays the single decision point.
+
 	if r.Method != http.MethodPost {
 		w.Header().Set("Allow", http.MethodPost)
 		writeSQSError(w, http.StatusMethodNotAllowed, sqsErrMalformedRequest, "SQS JSON protocol requires POST")
@@ -172,6 +186,27 @@ func (s *SQSServer) handle(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	handler(w, r)
+}
+
+// handleQueryProtocol owns the query-protocol leg of handle(): method
+// gating, SigV4 authorisation against the form body, and dispatch
+// into per-verb handlers. Pulled out of handle() so the dispatcher
+// stays under cyclop=10 even as more wire formats are added.
+func (s *SQSServer) handleQueryProtocol(w http.ResponseWriter, r *http.Request) {
+	// GET is legal for query (some legacy ListQueues callers).
+	// POST is the common case. Anything else (PUT/DELETE) is
+	// outside the SQS surface entirely.
+	if r.Method != http.MethodGet && r.Method != http.MethodPost {
+		w.Header().Set("Allow", "GET, POST")
+		writeSQSQueryError(w, newSQSAPIError(http.StatusMethodNotAllowed, sqsErrMalformedRequest,
+			"SQS query protocol requires GET or POST"))
+		return
+	}
+	if authErr := s.authorizeSQSRequest(r); authErr != nil {
+		writeSQSQueryError(w, newSQSAPIError(authErr.Status, authErr.Code, authErr.Message))
+		return
+	}
+	s.handleQuery(w, r)
 }
 
 func (s *SQSServer) serveHealthz(w http.ResponseWriter, r *http.Request) bool {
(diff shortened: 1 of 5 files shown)


bootjp:
Pushed addressing the converging Gemini + Codex findings:

**Bugs fixed:**
- Tags now parsed via `.Key`/`.Value` (was `.Name`/`.Value`) — CodexP1 + Gemini high. Query `CreateQueue` with tags previously silently dropped them.
- Indexed pair iteration now deterministic: pairs gathered, sorted by integer index, lower-index-wins on duplicates — CodexP2 + Gemini high.
- `RequestId` now actually 22 chars (was 26) — Gemini medium. 14 random bytes -> base32 no-pad -> trim.

**Drive-by:**
- `readQueryForm` adopts `r.URL.Query()` directly instead of copying — Gemini high.

**Refactor:**
- `collectIndexedKVPairs` split into gather + flatten halves with an `indexedPairKeyToIdx` helper to keep cyclop ≤ 10 (no `//nolint`).

3 new tests pin the fixes (Tag suffix, duplicate-key determinism over 64 iters, RequestId length). `go test -race` clean, `golangci-lint run` 0 issues.

@claude review

bootjp:
Thanks for the thorough re-review on `459242a2`. With your "code is ready to merge" verdict and CI green, this PR has reached its exit condition. The four minor non-blocking observations you noted (cosmetic `make([]…, 0)` → `var`, accepted negative indices, intentional JSON envelope on missing Action, proxy JSON-only error in the leader-flip window) all match either project conventions or documented design decisions; recording them as follow-up work in the partial doc rather than amending this proof PR.

Please merge at your convenience.

PR #663, opened 2026-04-26
perf(redis): event-driven XREAD BLOCK wake-up (Phase 1)

bootjp:
## Summary

- XREAD's busy-poll loop was the leader's #1 CPU consumer in production (45% of n3's 1530% / 14 CPUs from a 30s pprof). Each iteration re-issued `keyTypeAt + scanStreamEntriesAfter` every 10 ms regardless of whether any new entry had been written.
- Replace the `time.Sleep(redisBusyPollBackoff)` with an in-process notification: `streamWaiterRegistry` tracks BLOCKed XREADs by stream key; the local XADD path `Signal`s after `dispatchElems` returns (FSM applied → entry visible). The wait loop selects on `signal | fallback timer (100ms) | handlerCtx | deadline`.
- Notification-only — data plane is unchanged. Stream entries still flow through Raft → FSM → Pebble; the registry holds nothing but `chan struct{}` signals.

## Design

- **Phase 1 (this PR)**: signal fires from `xaddTxn` after `dispatchElems` returns. Covers the dominant production case (Redis-protocol XADD on the leader → XREAD BLOCK on the same leader).
- **Phase 2 (follow-up)**: FSM `ApplyObserver` hook. Covers (a) follower-side XREAD waiters, (b) Lua-script XADD, (c) future direct-mutation paths. Tracked in `docs/design/2026_04_26_proposed_fsm_apply_observer.md`.
- **100 ms fallback timer**: 10× reduction from the prior 10 ms busy-poll for paths that bypass the signal. Stays in place after Phase 2 as a safety net.

## CPU savings (production-measured)

- Signal-driven wake (leader XADD → leader XREAD): one wake per XADD instead of 100 polls/sec per blocked reader. Cuts XREAD's share of leader CPU from 45% to a small constant proportional to the XADD rate.
- Fallback poll (10 ms → 100 ms): 10× reduction for follower-side and Lua-driven writers.

1. **Data loss** — None. Notifications are advisory; data still flows through Raft → FSM → Pebble. A lost signal is bounded by the 100 ms fallback, after which `xreadOnce` re-reads the visible state.
2. **Concurrency** — `streamWaiterRegistry.mu` is held briefly: Register inserts under the lock; Signal copies the per-key waiter set out of the lock and then non-blocking-sends. Sends are non-blocking so XADD never stalls on a slow waiter. Registration completes before the first `xreadOnce` call so a signal that fires between check and wait is preserved in the size-1 channel buffer.
3. **Performance** — Wait-loop overhead is one timer + one select per iteration (was: one `time.Sleep`). Hot-path XADD adds a single non-blocking channel send per `Signal`; `Signal` short-circuits when no waiters are registered for the key.
4. **Data consistency** — XREAD's `readTS` is taken after wake, so the woken iteration sees a strictly fresh MVCC snapshot that includes the just-applied entry. Signal/visibility ordering is established by `dispatchElems` blocking until FSM apply.
5. **Test coverage** — unit tests cover `Register/Signal/release` semantics including the size-1 coalescing invariant and concurrent stress; end-to-end test exercises the leader-local XADD-wakes-XREAD path. Phase 2 (FSM ApplyObserver) extends coverage to follower-side and Lua-driven write paths.

## Test plan

- [x] `go test -race -run "TestRedis_Stream|TestStreamWaiter|TestRedis_BullMQ|TestNextXAddID|TestXAddEnforce" ./adapter/...` — passes (276s)
- [ ] CI: full adapter test suite under `-race`
- [ ] CI: jepsen redis suite (no behavior change to BLOCK/null contract; existing regression tests `TestRedis_StreamXReadShortBlockReturnsNullNotError` and `TestRedis_StreamXReadIterCtxDeadlineReturnsNull` cover the contract)

@claude review

### DIFF
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -221,6 +221,11 @@ type RedisServer struct {
 	// Exposed via HELLO / CLIENT ID.
 	connIDSeq atomic.Uint64
 
+	// streamWaiters lets XADD wake an XREAD BLOCK waiting on the same
+	// node, replacing what was a 10 ms time.Sleep busy-poll. See
+	// redis_stream_waiters.go.
+	streamWaiters *streamWaiterRegistry
+
 	route map[string]func(conn redcon.Conn, cmd redcon.Command)
 }
 
@@ -348,6 +353,7 @@ func NewRedisServer(listen net.Listener, redisAddr string, store store.MVCCStore
 		traceCommands:   os.Getenv("ELASTICKV_REDIS_TRACE") == "1",
 		baseCtx:         baseCtx,
 		baseCancel:      baseCancel,
+		streamWaiters:   newStreamWaiterRegistry(),
 	}
 	r.relay.Bind(r.publishLocal)
 
--- a/adapter/redis_compat_commands.go
+++ b/adapter/redis_compat_commands.go
@@ -28,7 +28,16 @@ const (
 	pubsubPatternArgMin  = 3
 	pubsubFirstChannel   = 2
 	redisBusyPollBackoff = 10 * time.Millisecond
-	redisKeywordCount    = "COUNT"
+	// redisStreamWaitFallback is the safety-net poll interval that fires
+	// in xreadBusyPoll when no XADD signal arrives. The signal path covers
+	// all in-process XADDs on the same node; the fallback only matters
+	// when the entry was applied via a path that does not call Signal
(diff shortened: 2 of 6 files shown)


bootjp:
@claude review

bootjp:
@claude review

PR #664, opened 2026-04-26
docs(sqs): proposals for Phase 3.C (throttling) + 3.D (split-queue FIFO)

bootjp:
## Summary

**Docs-only PR.** Two design proposals for the remaining Phase 3 SQS items per [`docs/design/2026_04_24_proposed_sqs_compatible_adapter.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_04_24_proposed_sqs_compatible_adapter.md) Section 14 (Phase 3 bullets). Both items were explicitly called out as needing separate design docs before any implementation work; this PR lands those proposals so the implementation PRs have a reviewed architecture to build on.

### 3.C — Per-queue throttling and tenant fairness ([proposal](docs/design/2026_04_26_proposed_sqs_per_queue_throttling.md))

Per-queue token-bucket throttling configured on queue meta (no separate keyspace), evaluated at the SQS adapter layer on the leader (no Raft per request), surfaced as the AWS `Throttling` error envelope so SDK retry/backoff just works.

Key decisions:
- **Default-off**. Existing queues are unaffected; operators opt in per queue via `SetQueueAttributes`.
- **Per-action buckets** (Send / Receive / Default) so a slow consumer cannot pin the producer.
- **Per-leader buckets, no replication**. Worst case on failover: one extra burst on the new leader. Acceptable per AWS-equivalent behaviour at region failover boundaries; replicating would cost a Raft commit per `SendMessage`.
- **Batch verbs charge by entry count**, not call count, with all-or-nothing rejection (matches AWS).
- **Admin-only configuration plane**. Standard SQS clients see `InvalidAttributeName` on the `Throttle*` attributes (matches AWS behaviour for unknown attributes); the data-plane enforcement runs for everyone.

### 3.D — Split-queue FIFO (high-throughput FIFO) ([proposal](docs/design/2026_04_26_proposed_sqs_split_queue_fifo.md))

Per-`MessageGroupId` hash partitioning across multiple Raft groups, mirroring AWS High Throughput FIFO. Within-group ordering preserved; across-group throughput scales with the partition count.

Key decisions:
- **Existing single-partition FIFO queues stay byte-identical** (`PartitionCount = 0` path is the legacy layout; no migration runs implicitly).
- **Power-of-two partition counts only** (1, 2, 4, 8, 16, 32) so the routing step is `hash & (N-1)` and future offline rebuilds stay tractable.
- **Partition count is immutable after first SendMessage**. Live re-partitioning would break ordering for in-flight messages of every group whose hash bucket changed; out of scope.
- **Multi-PR rollout plan** with an explicit "gate of no return" called out at PR 5 (the data-plane PR). PRs 1–4 are reversible no-ops on data layout; once a partitioned FIFO holds real data, rollback means draining and recreating the queue.
- **FNV-1a hash** (deterministic across processes / Go versions / architectures). Risk of attacker-controlled `MessageGroupId` pinning all traffic to one partition is documented and accepted (the feature is for cooperative operators).

## Test plan

- [x] Markdown renders correctly on GitHub (manually previewed).
- [x] Cross-references resolve (the partial-doc rename in PR #659 is in flight; both proposals reference the partial filename — they will resolve once #659 merges, otherwise resolve to the current `_proposed_` filename for a 1-character mismatch that is fine to leave for now).
- [x] No code changes; CI is irrelevant beyond the markdown lint pass.

## Self-review

This is a docs-only PR; the 5-lens self-review collapses to:

1. **Data loss / Concurrency / Performance / Consistency** — N/A, no code touched.
2. **Test coverage** — N/A, no code touched. The proposals themselves include a Testing Strategy section (§6 / §9) so the implementation PRs have explicit acceptance criteria.

## Stacking

This PR is **independent** of #650, #659, and #662. Branched from current `main`. Merge whenever ready — landing the proposal docs early lets reviewers comment on the architecture before the implementation PRs go up.

## Summary by CodeRabbit

* **Documentation**
  * Added detailed designs for optional per-queue token-bucket throttling (rules, batch semantics, AWS-shaped Throttling errors with Retry-After, metrics, rollout default-off) and for HT‑FIFO “split-queue” partitioning (routing, immutability, gating/rollout).
* **New Features**
  * Optional queue-level send/receive throttling with batch-aware charging and Retry-After; HT‑FIFO partitioning with deterministic partition routing and immutable partition attributes.
* **Tests**
  * Extensive unit and integration tests for throttling, HT‑FIFO partitioning, attribute validation, immutability, idempotency, and cache invalidation.


bootjp:
Pushed addressing all four review findings (Codex + Gemini converged):

**3.D split-queue FIFO doc:**
- **Keyspace backward-compat (CodexP1 + GeminiHigh)**: §3.1 rewritten — legacy queues use today's keys byte-identically, partitioned queues use a new `!sqs|msg|data|p|...` prefix with an explicit `p|` discriminator that cannot collide with legacy keys.
- **ReceiveMessage false-empty (CodexP1 + GeminiMedium)**: §4.2 rewritten — proxies non-local-leader partition scans via the existing leader-proxy (extended to take a partition argument) so a consumer pinned to one endpoint still sees every partition's messages. Added partitionOrder rotation to avoid head-of-line bias.

**3.C per-queue throttling doc:**
- **Hot-path lock contention (GeminiMedium)**: §3.1 switched to `sync.Map` lookup + per-bucket `sync.Mutex` for charge/refill. Cross-queue traffic never serialises on the same lock. Three-step charge operation documented explicitly.
- **Metric type (CodexP2)**: `sqs_throttle_tokens_remaining` is now correctly a **gauge** (token budgets go up *and* down). `sqs_throttled_requests_total` stays a counter.

Docs-only changes; no implementation gate. `+64/-22` lines across the two files.

@claude review

bootjp:
Pushed addressing all five findings from your follow-up review (1 medium, 4 low):

**3.D split-queue FIFO doc:**
- (medium) §8.1 stale failure-mode description rewritten — was contradicting §4.2's proxy fix. Now describes the actual remaining failure (proxy RTT cost under spread deployment) instead of the eliminated false-empty risk.
- (low) §3.1 variadic `sqsMsgDataKey` replaced with two named constructors (`legacyMsgDataKey` / `partitionedMsgDataKey`) + explicit call-site dispatch — no silent argument loss.
- (low) §3.1 `p|` discriminator safety made explicit: queue-name validator's `|` prohibition + base32-encoded segments never starting with literal `p` together guarantee no collision.
- (low) §8.5 upgrade-gate mechanism now concrete: `/sqs_health` advertises `capabilities: ["htfifo"]`; catalog polls every registered node, fail-closed on timeout. References PR #644 as prior art.

**3.C per-queue throttling doc:**
- (low) §3.4 `Retry-After` now computed per-queue from `ceil((1 - currentTokens) / refillRate)` (floored at 1 s for HTTP integer-second granularity) instead of the constant `1`. Slow-refill queues no longer cause busy-loops.

Docs-only changes; no implementation gate. `+24/-13` lines net across the two files.

@claude review

bootjp:
Pushed addressing all three findings from your third review (1 medium, 2 low):

**3.D split-queue FIFO doc:**
- (medium) §3.3 `partitionFor` now also short-circuits when `FifoThroughputLimit == "perQueue"` — fixes the contradiction with §3.2's documented semantics. Codex P2 from the second-pass review, now resolved.
- (low) §4.1 step 5 pseudocode now dispatches through the §3.1 named constructors (`legacyMsgDataKey` / `partitionedMsgDataKey`) instead of the old variadic form. No more conflicting guidance between §3.1 and §4.1.

**3.C per-queue throttling doc:**
- (low) §3.4 `Retry-After` formula now uses `requestedCount` instead of the hardcoded `1.0` numerator. `SendMessageBatch` of 10 against `refillRate=1.0` correctly returns `Retry-After: 10` instead of `1`.

Docs-only changes; `+18/-7` lines net.

@claude review

bootjp:
Pushed addressing all four findings from your fourth review (2 medium, 2 low):

**3.C per-queue throttling doc:**
- (medium) §3.1 now requires `SetQueueAttributes` to call `buckets.Delete(key)` after the Raft commit so a config change takes effect on the next request, not after the 1h idle-eviction window. Race with `LoadOrStore` is benign (rebuilt bucket starts at full capacity, same as failover).
- (low) §4.3 forward-compat fixed: bucket key for split-FIFO is `(queueName, partitionID)`, not `MessageGroupId` — the latter would create unbounded buckets per group value.

**3.D split-queue FIFO doc:**
- (medium) §5 now uses a new dedicated `--sqsFifoPartitionMap` flag for partition-to-Raft-group assignment instead of overloading `--raftSqsMap` (which keeps doing endpoint mapping for `proxyToLeader`). Eliminates the parsing ambiguity. §11 rollout table updated.
- (low) §9 unit test list now includes `FifoThroughputLimit=perQueue with PartitionCount=8 routes every group to partition 0` to lock down the §3.3 short-circuit guard against regression.

Docs-only changes; `+9/-6` lines net.

@claude review

bootjp:
Pushed addressing all three findings from your fifth-round review (2 medium, 1 low):

**3.C per-queue throttling — §3.1 `DeleteQueue` bucket lifecycle (medium):**
- New paragraph immediately after the `SetQueueAttributes` invalidation block. `DeleteQueue` now also calls `buckets.Delete(key)` for every `bucketKey` of the deleted queue (`Send`, `Recv`, `Default`), so a `DeleteQueue`+`CreateQueue` with the same name no longer inherits the previous incarnation's token balance/limits for up to 1h. Documented the `generation`-in-`bucketKey` alternative and the reason for picking the explicit `Delete` (cheaper, mirrors the `SetQueueAttributes` pattern).

**3.D split-queue FIFO — §4.2 `WaitTimeSeconds` budget across fanout (medium):**
- Pseudocode reworked: step 3 sets `deadline := start + WaitTimeSeconds (cap 20s)`; step 4a/4c/4d compute `remainingWait := max(0, deadline - now())` and pass it as the per-sub-call long-poll budget. Total wall-clock wait is now bounded by the original `WaitTimeSeconds` regardless of `PartitionCount`. Without this, an empty queue with `WaitTimeSeconds=20` + `PartitionCount=8` would hold the connection for up to 160s.
- Added a "Why the shared deadline matters" paragraph documenting the bug class and the rejected alternative (parallel probe with shared cancellation — rejected because it erases the `partitionOrder` rotation).
- §8.1 dropped the contradictory "long-poll mitigates the RTT cost" sentence; long-poll was actually the trigger for the bug.

**3.D split-queue FIFO — §3.4 `--raftSqsMap` reference (low):**
- Rephrased to distinguish the router infrastructure (`kv/shard_router.go`, which already supports multi-group routing by key) from the configuration surface (`--sqsFifoPartitionMap`, the new flag from §5). Calls out explicitly that conflating the two would parse partition assignments as endpoint pairs and route to the wrong leader. Aligns with §5.

@claude review

bootjp:
Pushed addressing all three findings from your sixth-round review (2 medium Codex P1 + 1 low Claude):

**3.C per-queue throttling — §3.2 capacity ≥ max batch charge (Codex P1, medium):**
- New fourth bullet on the `SetQueueAttributes` validator: when `SendCapacity > 0` it must also be `≥ 10`, and likewise `RecvCapacity > 0 → ≥ 10`. Justified inline against the §3.3 charging table (`SendMessageBatch` / `DeleteMessageBatch` charge up to 10 each). Documented the failure mode the rule prevents (permanently unserviceable batches with no recovery short of re-running `SetQueueAttributes`) and the rejection envelope (`InvalidAttributeValue` with the per-bucket minimum named so the operator sees the cause). Default* is exempt because no batch verb is in scope of the catch-all bucket today; explicit forward note for when one moves there.

**3.D split-queue FIFO — §8 item 5 runtime safeguard for downgraded leaders (Codex P1, medium):**
- New paragraph after the capability advertisement mechanism. Describes the failure sequence the create-time gate misses (downgrade after a partitioned queue exists → false-empty reads + key-prefix corruption from the rolled-back leader). Specifies the safeguard: on startup and on leadership acquisition for an SQS Raft group, enumerate local-shard queues with `PartitionCount > 1`; if any exist and the binary lacks `htfifo`, refuse leadership with an explicit log line and `TransferLeadership` away (group becomes leaderless if no peer is willing — desired fail-closed behaviour). Hooks into the existing leadership-acquisition path in `kv/lease_state.go` so steady-state cost is zero. Explicitly required before the §11 PR 4 rollout step that marks the binary `htfifo`-eligible.

**3.C per-queue throttling — §3.1 `bucketKey` text/struct consistency (Claude low):**
- The DeleteQueue paragraph said "we keep `bucketKey = (queueName, partitionID, action)` for now" but the struct definition above has only `queue` + `action` (no `partitionID` until Phase 3.D per §4.3). Corrected to "for Phase 3.C we keep `bucketKey = (queueName, action)` per the struct definition above — the `partition` field is added later by Phase 3.D, see §4.3".

@claude review

bootjp:
Pushed addressing all three findings from your seventh-round review (2 medium Codex P1 + 1 low Claude):

**3.C per-queue throttling — §3.1 action key naming consistency (Codex P1, medium):**
- Both invalidation paragraphs (`SetQueueAttributes`, `DeleteQueue`) now read `(`Send`, `Receive`, `*` — the canonical action values from the `bucketKey` struct above)`. The prior `(Send, Recv, Default)` was the JSON config field-name vocabulary, not the in-memory map key — implementations following the struct would have left `Receive` and `*` buckets stale.
- Added a new "Config-field → bucket-action mapping" paragraph immediately after the `sqsQueueThrottle` struct definition. Explicitly maps `Send*` → `bucketKey{action:"Send"}`, `Recv*` → `bucketKey{action:"Receive"}`, `Default*` → `bucketKey{action:"*"}`, and pins which vocabulary to use where (config-field for the JSON contract / `GetQueueAttributes` response; bucket-action for in-memory bucket-store ops). Implementation must apply the mapping after `SetQueueAttributes` commit.

**3.D split-queue FIFO — §3.2 immutability extended to FifoThroughputLimit + DeduplicationScope (Codex P1, medium):**
- Sentence rewritten: "PartitionCount, FifoThroughputLimit, and DeduplicationScope are all immutable after first SendMessage." Three bullets explain the failure mode for each:
  - `PartitionCount` — re-hashing every message, breaks in-flight ordering, multi-second op (existing reasoning).
  - `FifoThroughputLimit` — flipping `perMessageGroupId ↔ perQueue` activates/deactivates the §3.3 short-circuit; in-flight messages stay where they are, new sends for the same group route to a different partition, within-group FIFO is silently violated.
  - `DeduplicationScope` — changing scope reshapes the dedup key (`(queue, dedupId)` vs `(queue, partitionId, dedupId)` vs `(queue, MessageGroupId, dedupId)`); narrowing scope can resurrect previously-deduped duplicates, widening can suppress legitimately-new sends that collide with still-cached prior-scope entries.
- Concluding sentence pins the unifying invariant: each attribute participates in a routing or dedup decision whose correctness depends on every existing message having been written under one consistent value, and live mutation creates incompatible "before"/"after" sets the runtime cannot reconcile without a full drain.

**3.C per-queue throttling — §6 item 2 test capacity (Claude low):**
- Bumped from `SendCapacity=5 SendRefillPerSecond=1` / 5 messages / 6th throttled to `SendCapacity=10 SendRefillPerSecond=1` / 10 messages / 11th throttled. Added a parenthetical inline note explaining why the value changed (the new §3.2 validator would have rejected `SendCapacity=5` at setup).

@claude review

bootjp:
Pushed addressing both findings from your eighth-round review (1 medium + 1 low):

**3.D split-queue FIFO — §3.2 immutability gate moved from "after first SendMessage" to "from CreateQueue onward" (Claude P1, medium):**
- You're right that the prior wording required a non-trivial detection mechanism (Pebble range scan or `firstSendAt` meta field with a concurrent-send race). I went with option (a) — align with AWS. AWS freezes `FifoThroughputLimit` and `DeduplicationScope` at `CreateQueue` time; SDK clients expect the same rejection envelope on the same request, so the alignment is also better wire-compatibility.
- New "Enforcement (gate is `CreateQueue`, not first send)" paragraph spells out the validator: load the current `sqsQueueMeta` (one Raft-consistent point read — already required for the OCC compare-and-set) and reject with `InvalidAttributeValue` if the request value differs from the meta. No range scan, no extra meta field, no concurrent-send race.
- Documented the small loss explicitly: the corner case "operator creates with `PartitionCount=8`, changes mind before first send" is solvable by `DeleteQueue`+`CreateQueue` and rare enough that the simplicity of a stateless validator beats it.
- Updated the `PartitionCount` struct comment from "immutable after first SendMessage commits" → "immutable thereafter" pointing at the new §3.2 enforcement paragraph.

**3.D split-queue FIFO — §3.4 "no router changes are needed" contradicted §11 PR 4 (Claude low):**
- Replaced the contradictory sentence. New text: the router's *dispatch algorithm* (`kv/shard_router.go`) does not change (the byte-key dispatch core already supports multi-group routing), but PR 4 does touch `shard_router.go` to wire the `(queueName, partition)` composite key format and load per-partition shard assignments from `--sqsFifoPartitionMap`. An implementor lining §3.4 up with §11 PR 4 now sees consistent scope.

@claude review

bootjp:
Pushed addressing all three low-severity findings from your ninth-round review:

**3.D split-queue FIFO — §4.4 receipt-handle 4-byte vs 8-byte (low):**
- Right call — `partitionIndex` is `uint32` throughout the doc (`partitionFor` returns `uint32`, `PartitionCount` is `uint32`), so 4 bytes is the natural fixed-width encoding. Changed to "4-byte big-endian segment" with the rationale inline (the 8-byte version would waste 4 per handle and double this field's contribution to the base64-encoded handle the SDK exposes to operators).

**3.C per-queue throttling — §6 immediate bucket invalidation test (low):**
- New test item #3 between the existing E2E and config round-trip. Pins the §3.1 `buckets.Delete`-after-Raft-commit contract so a refactor cannot silently drop it.
- Three sub-cases: exhaust Send bucket → raise capacity → immediate next send must succeed (no sleep); same shape for `ReceiveMessage` (`RecvCapacity`); `DeleteQueue` + `CreateQueue` same-name lifecycle gets a fresh full-capacity bucket, not the stale empty one from the previous incarnation.
- Renumbered following items 4–7.

**3.D split-queue FIFO — §9 immutability enforcement test (low):**
- New test item #4 in the testing section. Pins the §3.2 enforcement rule (`PartitionCount` / `FifoThroughputLimit` / `DeduplicationScope` reject changes with `InvalidAttributeValue`) against the regression class you flagged — a one-point-read comparison that could be silently dropped if the validator's attribute dispatch changes.
- Sub-cases: per-attribute change (each named in the rejection message so the operator sees the cause); combined change in one request; mutable-mixed-with-immutable rejects without partially committing the mutable changes; same-value "no-op" succeeds (only differing values rejected). Renumbered Jepsen and metrics items.

Verified: `golangci-lint run` (pre-commit) → 0 issues.

@claude review

bootjp:
Pushed addressing both low findings from your tenth-round review:

**3.D split-queue FIFO — §3.2 all-or-nothing semantics in the enforcement paragraph (low):**
- Right call — the rule was pinned by the §9 test but the §3.2 spec required a cross-reference. New sentence in the enforcement paragraph: *"`SetQueueAttributes` is **all-or-nothing**: if any immutable attribute in the request carries a differing value, the entire request is rejected before any attribute is persisted — including the *mutable* attributes in the same call (e.g. `VisibilityTimeout` paired with an attempted `PartitionCount` change is rejected as a whole; the `VisibilityTimeout` change does not commit on its own). The §9 immutability test pins this rule."* An implementor now sees the all-or-nothing contract directly in §3.2.

**3.D split-queue FIFO — §4.2 `remainingWait=0` loop behaviour (low):**
- Picked option 2 (optimistic drain — your suggested intent for the proxy fanout). Removed "the deadline expires" from the loop terminator, and step 4a now reads: *"When remainingWait reaches 0, **continue iterating remaining unvisited partitions with WaitTimeSeconds=0 (non-blocking point-read)** rather than breaking immediately — a consumer must not miss a message that is already visible on an unvisited partition simply because the long-poll budget ran out mid-loop."* Plus an explicit line that the deadline degrades long-poll → point reads but does not cut the fanout short.

@claude review

bootjp:
Pushed addressing the last low finding from your eleventh-round review.

**3.D split-queue FIFO — §9 `WaitTimeSeconds` shared-deadline timing test (low):**
- New §9 item #5. Pins the §4.2 step 3 contract that total wall-clock wait is bounded by the original `WaitTimeSeconds` regardless of `PartitionCount`. The pseudocode you suggested becomes the test:
  - `PartitionCount = 4`, empty queue, `ReceiveMessage(WaitTimeSeconds = 2)` → assert returns ≤ 3 s. A naive refactor that drops the `remainingWait` threading would time out at 4 × 2 = 8 s.
  - Companion: `WaitTimeSeconds = 0` exercises the non-blocking-point-read fanout from §4.2 step 4a so the optimistic-drain path also has coverage.
- Renumbered Jepsen and metrics items.

You also said "the docs are ready for implementation PRs to proceed; neither this finding nor anything else in either doc represents a design gap that would block PR 2 from starting." Reading that as the review-OK signal for the design proposal — the new test item is the only remaining loose end and is now landed.

@claude review

bootjp:
Pushed addressing the two Codex P1 mediums you flagged as still open from the tenth-round Codex review on `14b4d88a82`. You're right — both were genuinely missed and both need to land before PR 2.

**3.D split-queue FIFO — control-plane gate for `{PartitionCount > 1, DeduplicationScope = "queue"}` (Codex P1, medium):**
- New "Cross-attribute validation at `CreateQueue` and `SetQueueAttributes`" paragraph in §3.2. Rejection envelope is `InvalidParameterValue` with the AWS-shaped reason: *"queue-scoped deduplication is incompatible with multi-partition FIFO because the dedup key cannot be globally unique across partitions without a cross-partition OCC transaction."*
- Reframed the §4.1 paragraph from a runtime rejection to a "cannot reach this code path because §3.2 rejects at control-plane time" pointer. The runtime rejection is now impossible — the operator gets the validation error at `CreateQueue` where they can cheaply fix the config, not at the first `SendMessage`.

**3.D split-queue FIFO — explicit PR 2 dormancy gate (Codex P1, medium):**
- PR 2 row now describes a temporary `CreateQueue` rejection: `PartitionCount > 1` returns `InvalidAttributeValue("PartitionCount > 1 requires HT-FIFO data plane — not yet enabled")`. The schema field exists in the meta type but cannot land in production data.
- PR 3 + 4 rows note the gate is still in place.
- PR 5 row explicitly **removes the gate in the same commit that wires the fanout** — the gate-and-lift are atomic so a half-deployed cluster can never accept a partitioned queue without the data plane to serve it.
- New "Why the temporary gate" paragraph below the rollout table walks through the failure sequence the gate prevents (PR 2–4 cluster accepts `CreateQueue(PartitionCount=4)` → `SendMessage` writes to legacy `!sqs|msg|data|<queue>|…` prefix → PR 5 reader looks for partitioned `!sqs|msg|data|p|<queue>|<partition>|…` and finds nothing → silent message loss invisible to reaper too).

Both fixes lock down real production-time bugs the design would otherwise leave unspecified for the implementer to discover.

@claude review

bootjp:
Pushed addressing both low findings from your thirteenth-round review:

**3.D split-queue FIFO — §3.2 cross-attribute rule at `SetQueueAttributes` is dead code (low):**
- New clarifying paragraph after the cross-attribute validation block: "In practice this rule is only reachable at `CreateQueue` time. `SetQueueAttributes` cannot present the `{PartitionCount > 1, DeduplicationScope = "queue"}` combination because both `PartitionCount` and `DeduplicationScope` are immutable: any change to either is caught by the immutability check (described in the next paragraph) before the cross-attribute check has the chance to run. Implementing the cross-attribute rule on the `SetQueueAttributes` path would therefore be dead code — it is harmless to add for symmetry, but PR 2's author should know it is unreachable rather than spend time hunting for the test case that hits it."

**3.D split-queue FIFO — §11 PR 4 "Mixed-version gate" expanded (low):**
- PR 4 row's "Mixed-version gate" parenthetical now explicitly lists both components required before the binary is marked `htfifo`-eligible: (1) §8.5 capability advertisement via `/sqs_health` + catalog polling for `CreateQueue` gating, **and** (2) the §8 leadership-refusal hook in `kv/lease_state.go` that calls `TransferLeadership` when a non-`htfifo` binary discovers a partitioned queue in its shard on startup or leadership acquisition. PR 4 author cannot now satisfy the table by delivering only the capability badge.

@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
Round 7 (`53ee4d42`) で Codex round-6 の P1 (`Include queue generation in throttle bucket keys` / `Invalidate throttle caches on leader changes`) を対応:

ご指摘どおり、`bucketKey` が `(queue, action)` のみだったため、

1. **DeleteQueue+CreateQueue cycle**: 同名 recreate で旧 incarnation の bucket (token balance, capacity 含む) を継承 → 新 queue が drained 状態でスタート、または逆に unexpected burst
2. **Cross-leader stale cache**: 旧 leader が cache に保持した `(queue, action)` bucket → idle eviction 前に再 leader 化 → 別 incarnation のトラフィックを処理してしまう

という correctness gap がありました。Per-process な `invalidateQueue` は処理した leader でしか走らないため、これは構造的な解決ではなく hot-path 最適化に過ぎませんでした。

**修正:**
- `bucketKey { queue, action, generation }` (Generation = `sqsQueueMeta.Generation`)
- `charge` / `loadOrInit` / `chargeBucket` / `chargeQueueWithThrottle` / `chargeQueue` が generation を貫通
- `queueThrottleConfig` は `(*sqsQueueThrottle, generation)` を返却 (meta-load hot path は single-trip 維持)
- `invalidateQueue` は固定 action list iteration から map.Range に変更 → 同 queue の **全 generation** を一括 drop (SetQueueAttributes は同 gen の bucket を、DeleteQueue/CreateQueue cycle は旧 gen も含めて drop)
- `throttleAllActions` は不要になり削除

Generation-keyed structure と `invalidateQueue` は complementary:
- `invalidateQueue`: 同 generation の SetQueueAttributes 反映 (hot-path optimization, in-memory map をコンパクト保つ)
- Generation in key: cross-leader incarnation isolation (correctness guarantee)

**新規 test:**
- `TestBucketStore_GenerationKeyedDoesNotReuseAcrossIncarnations`: gen=1 を drain → gen=2 が fresh full bucket で起動することを assert + gen=1 が drained のままであることも pin (cross-pollination 不在を確認)
- `TestBucketStore_InvalidateQueueClearsAllGenerations`: orders@gen=1, orders@gen=2, events@gen=1 を populate → `invalidate("orders")` で orders 系統 2 entries drop、events 残存

**Design doc 更新:**
- §3.1 `bucketKey` 構造体に generation 追加
- DeleteQueue / SetQueueAttributes / CreateQueue cache-invalidation paragraph を range-by-queue + generation-keyed の complementary mechanisms として書き直し
- "Cross-leader incarnation isolation" paragraph 追加 (該当 failure mode を文書化)

**残り findings (re-post):** §3.1 line 110 (Invalidate cached buckets on throttle config updates)、line 159 (Require bucket capacity to cover max batch token cost)、line 306 (Enforce HT-FIFO capability beyond queue creation) はいずれも prior round で対応済 (Codex round 1 の P1 / round 5 の P1 / Phase 3.D PR 5+ scope)。

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
Round 8 (`f664d283`) で round-7 CodeRabbit findings に対応:

**(1) Major (`adapter/sqs_catalog.go:510` Throttle attributes are exposed to every caller):**

ご指摘どおり、§5 で「admin principal の access-key-to-role mapping で Throttle* を gate する」と書いていた箇所は **aspirational だった** (実装されていない)。確認したところ:
- SQS HTTP 層には access-key → role mapping は存在せず、`WithSQSStaticCredentials` は access-key → secret のみ
- AdminPrincipal が乗るのは `AdminDeleteQueue` などの operator-only admin RPC で、HTTP route には届かない

Role layer 追加は separable change (independent design + access-key meta scheme + audit) なので、本 PR の scope では doc を実装と整合させる方を選択:
- §5 を rewrite: Throttle* は他の queue attribute と同じ SigV4 / IAM-policy controls 配下に置く (= AWS 自体の access control 粒度と同じ。SetQueueAttributes 権限を持つ principal が触れる)
- §6 test item 4 を変更: admin-vs-standard round-trip → unknown attribute name で `InvalidAttributeName` reject (AWS behaviour)
- §11 future-work に「admin-principal layer 追加」の note を残す parenthetical を §5 末尾に挿入

**(2) Minor (`docs/design/.../proposed_sqs_per_queue_throttling.md:30` Use JSON error shape, not query/XML one):**

ご指摘どおり Goal 3 が `<Code>Throttling</Code>` 1 形式しか書いていなかったので、§3.4 (per-protocol envelope) を参照する形に rewrite。両 protocol を明示。

**(3) Minor (`docs/design/.../proposed_sqs_split_queue_fifo.md:184` MD040 fenced-code-language):**

§4.1, §4.2, §4.3 の 3 つの unlabeled fence に `text` tag を付与。

**(4) Nitpick (`adapter/sqs_partitioning.go:97-102` Prefer checked cast over //nolint:gosec):**

ご指摘どおり、CLAUDE.md は `//nolint` を避けることを要求しています。元の 64-bit FNV-1a は `uint64 → uint32` narrowing が必要で gosec G115 を silence するために `//nolint` していた。

修正: **32-bit FNV-1a に書き換え** (offset `0x811c9dc5`, prime `0x01000193`)。32-bit hash は MessageGroupId を最大 32 partitions に spread するに十分なエントロピーを持ち、loop body は同形 (1 xor + 1 multiply per byte, allocation 0)、結果が native uint32 なので narrowing なし → `//nolint` 完全除去。`TestPartitionFor_*` の property tests (determinism, ±5% uniformity, mask=mod equivalence) はいずれも specific hash value を pin していないので green。Wire-protocol determinism contract (same input → same output) も維持。

**Round-7 Codex P1 re-posts:**
すべて prior round で対応済:
- "Invalidate cached buckets on throttle config updates" → round 1 (`SetQueueAttributes` 後の `invalidateQueue`)
- "Require bucket capacity to cover max batch token cost" → round 5 (`validateThrottlePair(requireBatchCapacity=true)` for Send/Recv/Default*)
- "Enforce HT-FIFO capability beyond queue creation" → Phase 3.D PR 5+ scope (CreateQueue gate は §11 PR 4 で document)
- "Invalidate throttle caches on leader changes" → round 7 (generation in bucketKey で cross-leader incarnation isolation)
- "Include queue generation in throttle bucket keys" → round 7 (上記)

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
Round 9 (`044128e5`) で round-8 Claude review の 2 件 (Medium + Low) に対応:

**(1) Medium — §3.2 Default* exemption parenthetical contradicts implementation:**

ご指摘どおり、§3.2 line 163 の末尾 "(`Default*` is exempt because the catch-all action set has no batch verb in scope for this PR…)" は**round 5 で fix された実装と矛盾していました**。`validateThrottleConfig` (sqs_catalog.go:656) は `requireBatchCapacity=true` を Default* にも渡しています — 理由は `resolveActionConfig` (sqs_throttle.go) が Send*/Recv* unset 時に Send/Receive traffic を Default bucket に fall-through させるため、batch verb (SendMessageBatch / DeleteMessageBatch、charge=10) が Default bucket に着弾する。`DefaultCapacity < 10` だと `SendCapacity < 10` と同じ permanently-unserviceable trap になる。

修正:
- §3.2 line 163: parenthetical を rewrite して fall-through reasoning を明記、earlier-draft の誤りを認める
- `validateThrottlePair` docstring (sqs_catalog.go:690-693): 三つの action set すべてが `requireBatchCapacity=true` を渡している現状を説明 + parameter は将来の non-batch action 用に残す旨を note

**(2) Low — §3.1 line 119 "fresh bucket at full capacity on failover" is over-broad:**

ご指摘どおり、3 つの failover path のうち 2 つしか説明していなかった。3 つに分けて enumerate:

1. **First-time leader**: cache miss → fresh full-capacity bucket
2. **Re-elected, config changed**: capacity/refillRate mismatch → reconciliation evicts → fresh full-capacity replacement
3. **Re-elected, same generation, same config**: cached bucket retained, next charge runs standard refill (`tokens += elapsed * refillRate`, capped at capacity)

Path 3 が **correctness feature** (operator が leadership を bouncing して burst を triggering できる経路を防ぐ) であることも明記。Worst-case burst は capacity 1 個分 (path 1 / 2) で path 3 は local cache の token-bucket semantics をそのまま継承、と本来の reasoning に繋げて書き直し。

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。Comment-only change (sqs_catalog.go) + design doc 2 site の編集のみ。

@claude review

bootjp:
@claude review

bootjp:
Round 10 (`34cb485e`) で round-9 review の 5 件に対応:

**(1) Codex P1 (`adapter/sqs_catalog.go:1424` Avoid resetting buckets on no-op throttle updates):**

ご指摘どおり、`setQueueAttributes` が `Throttle*` キーの presence のみで invalidate していたため、同値の no-op write でも bucket が full capacity に reset され、caller が自分の current config を tight loop で再投入することで rate limit を bypass できる経路がありました。

修正: `trySetQueueAttributesOnce` が `applyAttributes` の前に `meta.Throttle` を `snapshotThrottle` で snapshot → apply 後に `throttleConfigEqual` で比較 → `throttleChanged` を retry chain 経由で handler に伝搬 → handler は値が actually 変わった時のみ `invalidateQueue`。同 readTS での snapshot なので writer が race している値そのものと比較。

Test: `TestSQSServer_Throttle_NoOpSetQueueAttributesPreservesBucket` — drain → 同値 SetQueueAttributes → 次の send が依然 reject されることを assert。旧 code では allow されていた。

**(2) Codex P2 (`adapter/sqs_throttle.go:601` Do not fail open on throttle metadata read errors):**

ご指摘どおり、`queueThrottleConfig` が `loadQueueMetaAt` error を `(nil, 0)` に converted して fail-open になっていた経路を修正:
- 戻り値を `(*sqsQueueThrottle, uint64, error)` に変更
- `chargeQueue` は err != nil の時 HTTP 500 sqsErrInternalFailure で fail-closed
- `(nil, 0, nil)` は legitimate "queue does not exist" path として残し、surrounding handler が QueueDoesNotExist を返せるように

**(3) Codex P2 (`adapter/sqs_throttle.go:284` Avoid resetting throttle buckets on PurgeQueue generation bumps):**

ご指摘どおり、`PurgeQueue` が `Generation` を bump し、bucket key が `Generation` を含んでいたため、purge ごとに bucket re-key → fresh full bucket → caller が repeated PurgeQueue (60s rate-limited だが real bypass) で rate limit 回避できる経路がありました。

修正: `sqsQueueMeta.Incarnation` を新規導入。
- `CreateQueue` 時のみ bump (`requested.Incarnation = lastGen + 1`)
- `PurgeQueue` / `SetQueueAttributes` は read-modify-write で **value を preserve** (overwrite しない)
- `DeleteQueue` は meta を削除するが `sqsQueueGenKey` が persist するので、次の `CreateQueue` の `Incarnation` は monotonic に確保

Bucket key を `incarnation` field に rename。`charge` / `loadOrInit` / `chargeBucket` / `chargeQueueWithThrottle` / `chargeQueue` / `queueThrottleConfig` の signature 全部が `incarnation uint64` を渡すように変更。`PurgeQueue` は throttle bucket に一切 touch しない設計。

Test: `TestBucketStore_PurgeQueueDoesNotResetBucket` — drain → 同 incarnation で再 charge (purge 後に meta が持つ値) → 依然 reject を assert。Generation-keyed code では allow されていた。

既存 test は `TestBucketStore_IncarnationKeyed*` / `TestBucketStore_InvalidateQueueClearsAllIncarnations` に rename + diagnostic message 更新。

**(4) CodeRabbit Major (`docs/design/.../proposed_sqs_split_queue_fifo.md:37` Doc/implementation FNV-1a bit width mismatch):**

ご指摘どおり、§2.2 #4 と §3.3 pseudocode が round 8 で 32-bit に switch した実装と乖離していたので、constants/code を 32-bit に更新 + §3.3 散文に "PartitionCount ≤ 32 なら mask 後に survive する bit は ≤ 5、narrowing 回避が substantive な理由" の説明を追加。

**(5) CodeRabbit Minor (`docs/design/.../proposed_sqs_per_queue_throttling.md:293` Typo reqUest):**

修正済み。

**Design doc 更新:**
- §3.1 `bucketKey` 構造体: `generation` → `incarnation`
- Cache-invalidation paragraph rewrite: Incarnation semantics 説明 + 新規 "Why Incarnation, not Generation" paragraph で PurgeQueue bypass の経緯を文書化
- Failover path 3 を "same incarnation" 表現に統一

**Round-9 re-posts (already addressed):**
- Codex P1 caching/batch/HT-FIFO/leader/generation → round 1 / round 5 / Phase 3.D PR 5+ scope / round 6 / round 7
- CodeRabbit Major Throttle attribute access → round 8 (admin-only model removed)
- CodeRabbit Minor JSON shape → round 8 (Goal 3 protocol-agnostic rewrite)

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
Round 10.1 (`72d6485c`) で round-9 Claude review に残っていた最後の Low (`adapter/sqs_throttle.go:540-544` chargeQueue comment incorrectly claims sendMessageBatch uses chargeQueueWithThrottle) に対応。

ご指摘どおり、`chargeQueue` の docstring が `sendMessageBatch` を「`chargeQueueWithThrottle` を使うべき handler」に列挙していましたが、実際 `sqs_messages_batch.go:97,459,528` は `chargeQueue` を呼んでいました。実装が正しい (batch path は throttle check を per-entry OCC retry loop の外側で1回だけ行うので、charge 時点では meta を pre-load していない) ので、ドキュメントを実装に合わせる方を選択:

- `chargeQueue` を使う handler の列挙を更新: deleteMessage, changeMessageVisibility, **sendMessageBatch, receiveMessageBatch, deleteMessageBatch** (batch family を明示)
- `chargeQueueWithThrottle` を使う handler の列挙を更新: **sendMessage, receiveMessage** のみ
- 「batch handler を refactor して `chargeQueueWithThrottle` にしないのは何故か」の rationale を追加: per-entry meta load が必要になり savings を相殺するため

これで round-9 Claude review の全 4 findings (FNV major、PurgeQueue medium、no-op SetQueueAttributes low、fail-open low、typo low、chargeQueue comment low) を round 10 + 10.1 で完全対応。

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
@claude review

bootjp:
Round 11 (`d1ce7c1e`) で round-10 Codex P1 (`adapter/sqs_throttle.go:327` Hold bucket lock while removing stale bucket entries) に対応:

ご指摘どおり、`loadOrInit` の reconciliation path が unlock-then-CompareAndDelete-then-relock-then-flag の順序になっていたため、CompareAndDelete success と re-lock の間に concurrent charger が `bucket.mu` を取得 → `evicted=false` を観測 → orphaned bucket からトークンを spend、後続 request が fresh full-capacity bucket → SetQueueAttributes 直後 / leader 変更直後に **post-update burst** が configured limit を超える経路がありました。

**修正:** `bucket.mu` を matches check / CompareAndDelete / `evicted=true` の全期間 hold するように変更。sweep / invalidateQueue (rounds 5 / 6.1) と同じ lock-then-delete-then-flag ordering に統一。`mu` で block していた charger は unblock 時点で `evicted=true` を観測 → `chargeBucket` の retry loop 経由で live bucket に converge。

**新規 test:** `TestBucketStore_LoadOrInitReconciliationMarksOrphanEvictedUnderMu` — cfgOld で build した bucket に cfgNew で charge → reconciliation を deterministic に発火 → orphan bucket が `evicted=true` であることを mu 下で assert。Pre-fix code は同条件で gap 内に concurrent charger を入れると `evicted=false` を観測する経路を残していた。

Round-10 残りの Codex/CodeRabbit re-posts (caching / batch capacity / HT-FIFO capability / Throttle attribute access / generation-in-key / JSON shape) はすべて prior round (1, 5, 6, 7, 8, 10) で対応済の re-post です。

`go test -race -count=10 ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
Round 12 follow-up (`55297f6a`) で round-11 CodeRabbit Major review のうち、`642ec0d4` (round 12 main) でカバーされていなかった 2 件に対応:

**(1) CodeRabbit Major (`adapter/sqs_throttle.go:221` chargeBucket fail-open after retry exhaustion):**

ご指摘どおり、`chargeBucket` retry loop の terminal fallback が `chargeOutcome{allowed: true}` を返す fail-open になっており、invalidate/reconcile race を rate limit を enforce すべき path で bypass する経路がありました。

修正: terminal fallback を `{allowed: false, retryAfter: 1s, bucketPresent: false}` に変更。Pathological exhaustion 時に client は normal back-off cue を受け取り、次の attempt は (sweep / reconciliation が同じ fresh bucket を時間軸を跨いで evict できないので) ほぼ確実に live bucket に着弾します。

**(2) CodeRabbit Major (split-queue-fifo §7.2 PartitionCount implicit default):**

`642ec0d4` で §7.2 doc は "PartitionCount required for HT-FIFO" に更新されましたが、validator が wire で enforce していなかった。`validatePartitionConfig` に rule 追加:
- `meta.IsFIFO && meta.FifoThroughputLimit == htfifoThroughputPerMessageGroupID && meta.PartitionCount <= 1` → `InvalidAttributeValue` reject
- 新規 rule の追加で `validatePartitionConfig` が cyclop=10 を越えるため、`validatePartitionShape` (PartitionCount power-of-two + cap) と `validateStandardQueueRejectsHTFIFO` (FIFO-only attr rules) に refactor

新規 unit test: `TestValidatePartitionConfig_PerMessageGroupIDRequiresExplicitPartitionCount` — (PartitionCount ∈ {0, 1, 8}) × (perMessageGroupId / perQueue / 未指定) の accept/reject matrix を網羅。

既存 integration test (Immutability / AllOrNothing / GetQueueAttributesRoundTrip) を `perQueue` に変更 — `perMessageGroupId` は新 rule + dormancy gate (§11 PR 2) の合わせ技で wire 上 reachable でなくなったため。`perQueue` は legitimate な single-partition FIFO 設定で、同じ immutability / roundtrip path を exercise します。

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
Round 13 (`7e434888`) で round-12 Claude review の 4 件 (Medium x3 + Low) に対応:

**(1) Medium (`adapter/sqs_catalog.go:493` //nolint:gosec for PartitionCount):**

ご指摘どおり、CLAUDE.md は `//nolint` を avoid して refactor することを要求しています。`strconv.ParseUint(_, _, 32)` → `ParseUint(_, _, 64)` に変更し、`if n > math.MaxUint32` の explicit bounds check を追加してから narrow。`math` import は既存。gosec G115 は静かになり `//nolint` 不要。

**(2) Medium (`docs/design/.../proposed_sqs_per_queue_throttling.md` §3.1 line 115 doc divergence):**

ご指摘どおり、§3.1 が SetQueueAttributes の cache invalidation を `buckets.Delete(key)` per-action と書いていましたが、実装は `invalidateQueue` (lock-then-CompareAndDelete-then-flag protocol、round 5 / 6.1 で orphan-bucket race を closed) を使っています。Doc を `invalidateQueue` semantics の説明に rewrite し、raw per-key Delete だと closed race を reintroduce してしまうことを併記。Round 12 の no-op gate (same-value SetQueueAttributes が bucket reset しない) も同 paragraph に統合。

**(3) Medium (`docs/design/.../proposed_sqs_split_queue_fifo.md` §4.2 X-Elastickv-Receive-Partition internal-only):**

ご指摘どおり、`X-Elastickv-Receive-Partition` を `proxyToLeader` が設定する hop hint として説明していましたが、internal-only であることを明示していなかったため、malicious client が直接 public endpoint に同 header を送ると single-partition read を強制 / targeted partition の repeated drain などができる経路がありました。明示的な security note を追加: ingress handler が client-supplied 値を strip / unconditionally ignore することを doc 化、Phase 3.D PR 5 の handler change 内で strip を land する旨と integration test 計画への追加を §9 に明記。

**(4) Low (`docs/design/.../proposed_sqs_split_queue_fifo.md` §10 Open Questions stale items 3+4):**

ご指摘どおり、§10 item 3 ("default PartitionCount?") と item 4 ("--sqsDefaultFifoPartitionCount flag?") が §7.2 の "PartitionCount required, no server-side default" 決定と直接矛盾していました。両 item を削除、slot に "これらは §7.2 で resolved" 旨の parenthetical を追加して理由 (CreateQueue idempotency が hidden deployment state に依存する hazard) を pointer 化。元 item 5 (cross-partition ordering) を item 3 に renumber。

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
Round 13.1 (`8d6e2143`) で round-13 Claude review の Low x2 に対応:

**(1) Low (`docs/design/.../proposed_sqs_split_queue_fifo.md` §4.2 dangling §9 cross-reference):**

ご指摘どおり、§4.2 の security note に "the integration-test plan in §9 covers a 'client-set internal header is ignored' case" と書いていましたが、§9 の test list に該当 item が無く dangling pointer 状態でした。実装 PR の author が §9 を読んで guidance とした際に false confidence になる経路。

修正: §9 に新規 item 6 (`X-Elastickv-Receive-Partition` ingress strip) を追加 — 4-partition queue を作成し各 partition に 1 message ずつ着弾、direct client が `X-Elastickv-Receive-Partition: 0` header を付けて `ReceiveMessage(MaxNumberOfMessages=4)` → 4 件全て返ることを assert (full fanout 走行を確認)。Companion variant: header 無しで同じ response が返ることも assert (strip path が divergence を introduce しないこと)。Jepsen item を 6→7、metrics item を 7→8 に renumber。

**(2) Low (`adapter/sqs_catalog.go` PartitionCount parser comment references PR number):**

ご指摘どおり、CLAUDE.md は code comment で task / PR 番号を参照することを禁じています ("they rot as the codebase evolves")。先 round の comment が "(CLAUDE.md asks us to refactor rather than suppress; Claude medium on PR #664 round 12 flagged the suppression)" を含んでいたので削除。残ったのは substantive な WHY のみ ("Parse at uint64 width and bound-check explicitly so the uint32 narrowing below is gosec-clean.")。

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。

@claude review

bootjp:
Round 14 (`213ded44`) で round-13.1 Claude review の Low (pervasive PR-number references in implementation code comments) に対応:

ご指摘どおり、CLAUDE.md は code comment で task / PR 番号 / review attribution を参照することを明示的に禁じており ("they rot as the codebase evolves")、SQS Go files に ~25+ 個の "Codex P1 on PR #679", "CodeRabbit Major on PR #664 round 11" 等の comment が点在していました。

修正: 11 ファイル横断で sweep:

- `adapter/sqs_throttle.go`
- `adapter/sqs_catalog.go`
- `adapter/sqs_partitioning.go`
- `adapter/sqs_messages.go`
- `adapter/sqs_throttle_test.go`
- `adapter/sqs_partitioning_test.go`
- `adapter/sqs_throttle_integration_test.go`
- `adapter/sqs_partitioning_integration_test.go`
- `adapter/sqs_catalog_test.go`
- `adapter/sqs_query_protocol.go`
- `adapter/sqs_query_protocol_test.go`

各 comment は **substantive な WHY** (なぜその設計判断をしたか) を保持し、reviewer/PR number/round への parenthetical attribution のみ削除。Design doc (`.md`) は archival record として attribution が acceptable practice なので exempt (Claude review でも同じ判断)。

`grep -nE "round [0-9]+|PR #[0-9]+|Round [0-9]+|round-[0-9]+" adapter/sqs_*.go` で確認 — 残っている参照はゼロ。

`go test -race ./adapter/...` clean、`golangci-lint ./...` clean。Behaviour 変更なし。

@claude review

PR #673, opened 2026-04-26
admin: AdminForward integration for S3 admin ops (P2 slice 2b)

bootjp:
P2 slice 2b: a follower-side S3 admin write (POST /buckets, PUT /buckets/{name}/acl, DELETE /buckets/{name}) now hands off to the leader transparently, completing the same end-to-end forwarding contract Dynamo writes received in #644 + #648.

**Stacked on #669 (P2 slice 2a).** Once #669 merges, this rebases cleanly onto main.

## Summary

- **Proto** — three new `ADMIN_OP` enum values (`CREATE_BUCKET` / `DELETE_BUCKET` / `PUT_BUCKET_ACL`) appended after the Dynamo block so existing wire integers stay stable. Regenerated with the pinned protoc 29.3 / protoc-gen-go 1.36.11 / protoc-gen-go-grpc 1.6.1.
- **Leader-side** — `ForwardServer.WithBucketsSource` lets deployments wire the S3 dispatcher optionally; Dynamo-only builds keep it nil and the new operations return 501. Three new dispatch arms mirror the leader-direct HTTP path's payload contract (NUL-byte rejection, 64 KiB limit, `DisallowUnknownFields`, trailing-token rejection, slash-in-name rejection). `forwardBucketsErrorResponse` mirrors `forwardErrorResponse`: `ErrBucketsForbidden` / `NotLeader` / `NotFound` / `AlreadyExists` / `NotEmpty` + `*ValidationError` each map to the same HTTP status `writeBucketsError` produces, so forwarded and leader-direct responses are byte-identical from the SPA's view.
- **Follower-side** — `LeaderForwarder` interface gains `ForwardCreateBucket` / `ForwardDeleteBucket` / `ForwardPutBucketAcl`. `ForwardPutBucketAcl` carries both the bucket name (URL path) and the new ACL (request body) in one JSON payload — same approach `ForwardDeleteTable` takes for the table name.
- **Handler integration** — `S3Handler.WithLeaderForwarder` + `tryForwardCreateBucket` / `tryForwardPutBucketAcl` / `tryForwardDeleteBucket` gated on `ErrBucketsNotLeader && forwarder != nil`. `writeForwardResult` / `writeForwardFailure` mirror the Dynamo handler's pattern (nosniff + Cache-Control:no-store + Retry-After:1 on 503).
[...]

PR #669, opened 2026-04-26
admin: S3 bucket write endpoints (P2 slice 2a)

bootjp:
P2 slice 2a of [docs/design/2026_04_24_proposed_admin_dashboard.md](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_04_24_proposed_admin_dashboard.md). Ships the S3 admin write endpoints so the SPA's S3 modals stop receiving 405. Slice 2b will plumb AdminForward so a follower can hand these writes off to the leader transparently.

## Summary

- **`*adapter.S3Server.AdminCreateBucket` / `AdminPutBucketAcl` / `AdminDeleteBucket`** — SigV4-bypass write methods with three in-method guards: principal must be `AdminRoleFull`, the local node must be the verified S3 leader, and bucket-name / ACL must pass the existing validators. `AdminCreateBucket` reuses the same atomic bucket-meta + ACL + generation-key txn the SigV4 path does — no new code path through the storage layer. `AdminDeleteBucket` rejects non-empty buckets (the dashboard cannot force recursive delete by design).
- **`internal/admin` write surface** — `BucketsSource` gains the three write methods + `CreateBucketRequest` / `PutBucketACLRequest` types with the documented JSON shapes. `S3Handler.serveCollection` + `servePerBucket` route POST/PUT/DELETE through dedicated handlers with `principalForWrite` re-validating the role on every request against the live `MapRoleStore`.
- **Strict body decoder** — `decodeAdminS3JSONBody` is generic over the request type, applies `DisallowUnknownFields`, rejects NUL bytes, rejects trailing tokens, and caps at 64 KiB (matches design 4.4). Used by both POST and PUT.
- **`writeBucketsError`** translates the source-side sentinels into the design's HTTP statuses: 403 forbidden / 503 + Retry-After:1 leader_unavailable / 404 not_found / 409 already_exists / 409 bucket_not_empty / 400 invalid_request via `*ValidationError`.
- **Bridge** — `bucketsBridge` gains write methods running through `translateAdminBucketsError`, mirroring `translateAdminTablesError` on the Dynamo side. Leader-churn sentinels from the kv coordinator route to `admin.ErrBucketsNotLeader` so the SPA's retry contract stays intact.

## What is NOT in this PR

- AdminForward integration for S3 admin writes — slice 2b.
- Rolling-upgrade compatibility flag (criterion 5) — still deferred behind a cluster-version bump.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
[...]

PR #674, opened 2026-04-26
docs: add admin dashboard operator guide (P4)

bootjp:
P4 deliverable from `docs/design/2026_04_24_proposed_admin_dashboard.md` Section 8: a single self-contained operator-facing reference for the admin HTTP listener.

**Stacked on #669 (slice 2a) + #673 (slice 2b)** so the API-surface table can describe S3 write endpoints as shipped. Rebases cleanly onto main once those land.

## Sections

- **Quick start** — minimal loopback-dev invocation
- **Configuration reference** — required vs optional flags, with the rationale for each guard
- **Hard-error startup conditions** — the explicit cases where the process refuses to start (missing creds, partial TLS, non-loopback without TLS, role-list overlap)
- **TLS setup** — loopback / TLS / discouraged plaintext-non-loopback topologies
- **Roles** — read-only vs full + how live role re-validation works on every state-changing request
- **API surface** — full `/admin/api/v1/*` table including the slice 2 S3 write endpoints
- **Follower → leader forwarding** — what the SPA sees, the rolling-update caveat, and the election-period 503 + Retry-After contract
- **Audit log** — `admin_audit` slog shape and the `forwarded_from` field
- **Troubleshooting** — the common bring-up failures (missing creds, TLS hard-error, 401 ambiguity, 503 leader_unavailable, bucket_not_empty, placeholder bundle blank-screen)

## What is NOT in this PR

Section 8's P4 plan also called out "TLS, read-only role, CSRF" as deliverables; those are already implemented in main (see `config.go`'s `validateTLS` / `validateAccessKeyRoles`, the role gates in `DynamoHandler.principalForWrite` + `S3Handler.principalForWrite`, and the `CSRFDoubleSubmit` middleware). This doc stitches them into a single reference operators can land on without reading code.

## Test plan

- [x] Markdown renders cleanly (manual check)
- [x] Cross-references match real file paths (design doc, architecture overview, proto file)
- [x] Flag names match `main.go`'s flag definitions verbatim
- [x] The `bucket_not_empty` 409 response shape matches `S3Handler.writeBucketsError`

### DIFF
--- a/docs/admin.md
+++ b/docs/admin.md
@@ -42,7 +42,7 @@ with the access key + secret pair from the credentials file.
 | Flag | Description |
 |---|---|
 | `-adminEnabled` | Master on/off switch. Default `false`. |
-| `-adminSessionSigningKey` *or* `-adminSessionSigningKeyFile` *or* `ELASTICKV_ADMIN_SESSION_SIGNING_KEY` | Cluster-shared base64-encoded HS256 key (≥ 32 raw bytes / 44 base64 chars). **Must be the same on every node** — JWTs minted by node A are verified by node B during follower→leader forwarding, so a mismatch breaks the dashboard's read paths on follower nodes. The `*File` / env-var forms keep the secret out of `/proc/<pid>/cmdline`. |
+| `-adminSessionSigningKey` *or* `-adminSessionSigningKeyFile` *or* `ELASTICKV_ADMIN_SESSION_SIGNING_KEY` | Cluster-shared base64-encoded HS256 key — **exactly 64 raw bytes** (88 base64 chars with standard padding, or 86 with `RawURLEncoding`). The validator rejects any other length at startup with a precise error message. **Must be the same on every node** — JWTs minted by node A are verified by node B during follower→leader forwarding, so a mismatch breaks the dashboard's read paths on follower nodes. The `*File` / env-var forms keep the secret out of `/proc/<pid>/cmdline`. |
 | `-s3CredentialsFile` | JSON file with at least one access key + secret key pair. Same file the S3 adapter uses for SigV4; the admin dashboard reuses it for login authentication. |
 | `-adminFullAccessKeys` *and/or* `-adminReadOnlyAccessKeys` | Comma-separated allow-lists. Only access keys listed here may log into the dashboard, even if their SigV4 secret validates against the credentials file. Keys must not appear in both lists. |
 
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

bootjp:
@claude review

303f8f93 (rebased onto updated #673 `feat/admin-s3-forward`) addresses the four findings:

1. **High**: HS256 key size now correctly says "exactly 64 raw bytes — 88 base64 chars (standard padding) or 86 (RawURLEncoding)". The validator in `internal/admin/config.go` enforces exactly 64; the prior "≥ 32 / 44 chars" would have produced a confusing startup error.
2. **Moderate**: cross-reference link uses `_proposed_` (matches main today) with an inline note that the link will follow once #675 lands.
3. **Moderate**: audit log sample replaced with two real shapes — leader-direct (`operation=create_bucket bucket=...`) and forwarded (`forwarded_from=n2 operation=put_bucket_acl bucket=... acl=...`). An operator building a log parser against the documented contract will now find matches.
4. **Nit**: "data-plane" label for `/admin/api/v1/cluster` replaced with "admin".

bootjp:
@claude review

Addressed both findings from the previous review:

1. **Audit log shapes (Bug A)** — rewrote the section as three explicit shapes: `Audit` middleware (always wraps non-GET requests), `S3Handler` operation line (only S3 admin path emits these), `ForwardServer` operation line (covers both Dynamo and S3 forwarded paths). Removed the misleading `delete_table` leader-direct sample and added a one-sentence note that a single request typically produces two lines.
2. **Login 401 vs 403 (Codex P2)** — split into two troubleshooting entries: 401 invalid_credentials (bad access_key or secret) and 403 forbidden (valid creds but no admin role assignment), each with the remediation specific to that case.

Latest commit: 05560e79.

bootjp:
@claude review

Addressed both Codex P2 findings on c9da9235:

1. **Cookie Secure flag wiring (line 55)** — `-adminAllowPlaintextNonLoopback` only bypasses the startup TLS guard; cookie `Secure` is controlled separately by `-adminAllowInsecureDevCookie` (verified in `auth_handler.go:114` + `main_admin.go:597`). Rewrote the line-55 entry, the loopback section, and the discouraged-plaintext section to scope each flag correctly and call out that a working plaintext rig needs both. Operationally this is the difference between fails-closed and broken-end-to-end (browser refuses to send Secure cookie over HTTP → 401 every request).

2. **SPA Retry-After retry claim (line 172)** — verified in `web/admin/src/api/client.ts:75` and `useApi.ts` that there is no automatic retry. Rewrote the Election-period 503 section to describe current behaviour: server emits `Retry-After: 1` (kept in the contract for future clients / external scripts), SPA surfaces the 503 directly, operator must re-issue.

Latest commit: 02d52dc6. Doc-only; no code changes.

bootjp:
@claude review

Addressed the new Codex P2 finding on `docs/admin.md:205` (audit middleware coverage):

The previous text overstated coverage by claiming the `Audit` middleware fires for every non-GET write. Verified in `internal/admin/server.go:236-296`:

- `/auth/login` uses `publicAuth` (no Audit) — `AuthService` emits its own `action=login` line.
- `/auth/logout` uses `protectNoAudit` (Audit deliberately omitted) — `AuthService` emits `action=logout`; wrapping Audit too would produce two lines per logout.
- The `protect` chain is `BodyLimit → SessionAuth → Audit → CSRFDoubleSubmit`, so unauthenticated requests are rejected at SessionAuth before reaching Audit. CSRF-rejected requests still produce a middleware line (the actor is known by then) — this ordering is preserved in the doc rewrite.

Updated the middleware-shape section to enumerate which endpoints are wrapped and which aren't, plus the intro paragraph (auth endpoints produce one line, protected-chain mutations produce two).

Latest commit: 90100850. Doc-only.

bootjp:
@claude review

Addressed the Codex P2 finding on `docs/admin.md:260`:

The previous text said both login and logout `admin_audit` entries include `claimed_actor`, but `auditLogout` (auth_handler.go:442-449) only emits `action`, `actor`, `remote`, `status`. `claimed_actor` is login-only because it exists to distinguish the typed key from the authenticated one — logout has no claim to verify.

Replaced the one-line summary with two concrete sample audit lines (login with `claimed_actor`, logout without) plus a sentence telling parsers `claimed_actor` is present-only-on-login. Verified against auth_handler.go:432-449.

Latest commit: 85df320f. Doc-only.

PR #676, opened 2026-04-26
feat(raft): learner / non-voter membership (M1 engine + M2 operator surface)

bootjp:
Add Raft learner / non-voter membership to elastickv. Operators can now attach a fresh node as a non-voting replica, let it catch up on the log, and then promote it to voter — without ever shrinking the cluster's effective fault tolerance during catch-up.

The codebase already had partial scaffolding (`Server.Suffrage`, `RaftAdminMember.suffrage`, `applyConfigPeerChangeToMap` keeping learner metadata) waiting for a consumer; this PR turns it on end-to-end.

### Design doc (`docs/design/2026_04_26_proposed_raft_learner.md`)
Five rounds of review. Approved on Round 5: "ready to proceed to Milestone 1 implementation."

### Milestone 1 — engine-level learner support
- `Admin.AddLearner` and `Admin.PromoteLearner` on `internal/raftengine`.
- Persisted peers file bumped to v2 with per-peer suffrage byte; v1 reader preserved for upgrade compat.
- New apply-loop-owned `voterCount` + `isLearnerNode` cache (initialized at Open from snapshot ConfState; rebuilt from scratch — never patched — on every conf change).
- All four `len(e.peers)` overloads switched to `voterCount` (`recordQuorumAck`, `refreshStatus` single-node fast path, `removePeer` post-removal, plus the no-op audit on `nextPeersAfterConfigChange`). Closes the 1-voter+1-learner lease-read stall the design flagged.
- `applyConfigChange` / `applyConfigChangeV2` extended with `confState raftpb.ConfState`. Four-step apply-loop sequence: refresh voter cache → per-peer cleanup → rebuild `e.config.Servers` → bookkeeping.
- `validateConfState` accepts learner-only entries; voter check stays element-wise, learner check is set-based.
- `upsertPeer` no longer owns `e.config.Servers`; the apply loop rebuilds it via `configurationFromConfState`.
- `PromoteLearner` preconditions checked synchronously on the admin loop **before** propose: target must be a learner; leader's `Progress.Match >= minAppliedIndex`.

### Milestone 2 — operator surface
- gRPC `RaftAdmin.AddLearner` and `RaftAdmin.PromoteLearner` RPCs.
- `cmd/raftadmin add_learner <id> <address> [previous_index]` and `cmd/raftadmin promote_learner <id> [previous_index] [min_applied_index]`.
- `--raftJoinAsLearner` flag — operator alarm (not consensus veto): when a node booted with the flag finds itself in `ConfState.Voters`, an `ERROR`-level structured log fires once and `JoinRoleViolationCount` increments. The node keeps running.
- `docs/raft_learner_operations.md` runbook: prerequisites, attach/promote/verify, `remove_server` for learners, common errors.

## Out of scope (deferred to M3 / separate proposal)
- Follower-served reads (separate proposal needed for staleness bounds and adapter routing).
- Joint consensus / atomic multi-member reconfig.
- Jepsen workload exercising attach during partition.
- `Status.PerPeer` exposing `Progress.Match` for `min_applied_index` selection.
- Auto-promotion (engine never decides on its own to promote).

## Test evidence
- `internal/raftengine/etcd/learner_test.go` — 6 new tests:
  - `TestAddLearnerReplicatesWithoutCountingAsVoter`
  - `TestPromoteLearnerSwapsRoleToVoter`
  - `TestPromoteLearnerRejectsNonLearner`
  - `TestPromoteLearnerRejectsNotCaughtUp`
  - `TestRemovePeerLearnerKeepsSingleNodeFastPath` (the headline 1-voter+1-learner lease-read regression)
  - `TestJoinAsLearnerAlarmFiresWhenAddedAsVoter`
  - `TestLinearizableReadOnLearnerForwardsToLeader`
- `internal/raftengine/etcd/peer_metadata_test.go` — 4 new tests covering v2 round-trip with mixed suffrage, v1 forward-compat, v2 always-write, unknown-version rejection.
- `internal/raftadmin/server_test.go` — `TestServerMapsEngineAdminMethods` extended for the two new RPCs.
- `cmd/raftadmin/main_test.go` — `add_learner` / `promote_learner` happy-path subcommand tests + extended usage assertions.
- All other suites (`internal/...`, `kv/...`, `store/...`, `monitoring/...`, `distribution/...`, `proxy/...`, `cmd/...`) pass.
- `golangci-lint`: 0 issues.

1. **Data loss** — Promotion never decrements voter count; quorum size after promote is `floor((N+1)/2)+1`, ≥ pre-promote. WAL purge tracker includes learners by construction. v2 peers file forward-compat is read-once-write-once.
2. **Concurrency** — Add/promote run on the single-threaded admin loop. `voterCount` / `isLearnerNode` cache is single-writer (apply loop) under existing `e.mu`. `recordQuorumAck` learner-ack reject reads `isLearnerNode` under the same lock as `e.peers`. Promote-during-leader-change handled via `min_applied_index` precondition.
3. **Performance** — Adding a learner adds outbound replication traffic but no extra Raft round-trips on the write path; no extra per-call cost on the lease-read fast path. v2 peers file overhead: 1 byte per peer (bounded at 1024 peers).
4. **Data consistency** — Learner does not serve `LinearizableRead` from local FSM (returns `ErrNotLeader`). Voter / learner suffrage is sourced from `ConfState`, not `e.peers`. Snapshot-restore path also refreshes `voterCount` / `isLearnerNode` and runs the alarm hook.
5. **Test coverage** — New / changed branches each have a unit test as listed above; conformance test confirms suffrage round-trips across restart; lease-read regression test pins the 1-voter+1-learner fast path.

## Caveats
- `proto/Makefile` pins `libprotoc 29.3`; this branch was regenerated with `libprotoc 34.0` (host) + `protoc-gen-go v1.36.11` + `protoc-gen-go-grpc 1.6.1` (matching). The diff is the standard "two new messages + msgTypes index shift" — re-run `make gen` with the pinned toolchain if byte-identical generation matters.
- Branch is **64 commits behind `origin/main`** as of writing. Rebase before merge.

### DIFF
--- a/cmd/raftadmin/main.go
+++ b/cmd/raftadmin/main.go
@@ -23,8 +23,13 @@ const (
 	defaultRPCTimeoutSeconds = 5
 	minCommandArgs           = 2
 	addVoterArgCount         = 2
+	addLearnerArgCount       = 2
+	promoteLearnerArgCount   = 1
 	transferArgCount         = 2
 	addVoterPrevIndex        = 2
+	addLearnerPrevIndex      = 2
+	promoteLearnerPrevIndex  = 1
+	promoteLearnerMinApplied = 2
 	removePrevIndex          = 1
 	timeoutEnv               = "RAFTADMIN_RPC_TIMEOUT_SECONDS"
 	allowInsecureEnv         = "RAFTADMIN_ALLOW_INSECURE"
@@ -179,16 +184,26 @@ func isLoopbackHost(host string) bool {
 	return ip != nil && ip.IsLoopback()
 }
 
+func readOnlyCommands() map[string]func(context.Context, pb.RaftAdminClient) error {
+	return map[string]func(context.Context, pb.RaftAdminClient) error{
+		"leader":        printLeader,
+		"state":         printState,
+		"configuration": printConfiguration,
+		"config":        printConfiguration,
+	}
+}
+
 func executeCommand(ctx context.Context, client pb.RaftAdminClient, command string, commandArgs []string) error {
+	if handler, ok := readOnlyCommands()[command]; ok {
+		return handler(ctx, client)
+	}
 	switch command {
-	case "leader":
-		return printLeader(ctx, client)
-	case "state":
-		return printState(ctx, client)
-	case "configuration", "config":
-		return printConfiguration(ctx, client)
 	case "add_voter":
 		return addVoter(ctx, client, commandArgs)
+	case "add_learner":
+		return addLearner(ctx, client, commandArgs)
+	case "promote_learner":
+		return promoteLearner(ctx, client, commandArgs)
 	case "remove_server":
 		return removeServer(ctx, client, commandArgs)
 	case "leadership_transfer":
@@ -253,6 +268,50 @@ func addVoter(ctx context.Context, client pb.RaftAdminClient, args []string) err
 	})
 	if err != nil {
 		return errors.Wrap(err, "add voter")
+	}
+	fmt.Printf("index: %d\n", resp.Index)
+	return nil
+}
+
+func addLearner(ctx context.Context, client pb.RaftAdminClient, args []string) error {
+	if len(args) < addLearnerArgCount || len(args) > addLearnerArgCount+1 {
+		return usageError("add_learner")
+	}
+	prevIndex, err := parseOptionalUint64(args, addLearnerPrevIndex)
+	if err != nil {
+		return err
+	}
+	resp, err := client.AddLearner(ctx, &pb.RaftAdminAddLearnerRequest{
+		Id:            args[0],
+		Address:       args[1],
+		PreviousIndex: prevIndex,
+	})
+	if err != nil {
+		return errors.Wrap(err, "add learner")
+	}
+	fmt.Printf("index: %d\n", resp.Index)
+	return nil
+}
+
+func promoteLearner(ctx context.Context, client pb.RaftAdminClient, args []string) error {
+	if len(args) < promoteLearnerArgCount || len(args) > promoteLearnerArgCount+2 {
+		return usageError("promote_learner")
+	}
+	prevIndex, err := parseOptionalUint64(args, promoteLearnerPrevIndex)
+	if err != nil {
+		return err
+	}
+	minApplied, err := parseOptionalUint64(args, promoteLearnerMinApplied)
+	if err != nil {
+		return err
+	}
+	resp, err := client.PromoteLearner(ctx, &pb.RaftAdminPromoteLearnerRequest{
+		Id:              args[0],
+		PreviousIndex:   prevIndex,
+		MinAppliedIndex: minApplied,
+	})
+	if err != nil {
+		return errors.Wrap(err, "promote learner")
 	}
 	fmt.Printf("index: %d\n", resp.Index)
 	return nil
@@ -336,23 +395,24 @@ func usageError(command string) error {
 	return fmt.Errorf("usage: %s", usage(command))
 }
 
+var usageStrings = map[string]string{
+	"leader":                        "raftadmin <addr> leader",
+	"state":                         "raftadmin <addr> state",
+	"configuration":                 "raftadmin <addr> configuration",
+	"config":                        "raftadmin <addr> configuration",
+	"add_voter":                     "raftadmin <addr> add_voter <id> <address> [previous_index]",
+	"add_learner":                   "raftadmin <addr> add_learner <id> <address> [previous_index]",
+	"promote_learner":               "raftadmin <addr> promote_learner <id> [previous_index] [min_applied_index]",
+	"remove_server":                 "raftadmin <addr> remove_server <id> [previous_index]",
+	"leadership_transfer":           "raftadmin <addr> leadership_transfer",
+	"leadership_transfer_to_server": "raftadmin <addr> leadership_transfer_to_server <id> <address>",
+}
+
+const usageFallback = "raftadmin <addr> <leader|state|configuration|add_voter|add_learner|promote_learner|remove_server|leadership_transfer|leadership_transfer_to_server> [args]"
+
 func usage(command string) string {
-	switch command {
-	case "leader":
-		return "raftadmin <addr> leader"
-	case "state":
-		return "raftadmin <addr> state"
-	case "configuration", "config":
-		return "raftadmin <addr> configuration"
-	case "add_voter":
-		return "raftadmin <addr> add_voter <id> <address> [previous_index]"
-	case "remove_server":
-		return "raftadmin <addr> remove_server <id> [previous_index]"
-	case "leadership_transfer":
-		return "raftadmin <addr> leadership_transfer"
-	case "leadership_transfer_to_server":
-		return "raftadmin <addr> leadership_transfer_to_server <id> <address>"
-	default:
-		return "raftadmin <addr> <leader|state|configuration|add_voter|remove_server|leadership_transfer|leadership_transfer_to_server> [args]"
+	if u, ok := usageStrings[command]; ok {
+		return u
 	}
+	return usageFallback
 }
--- a/cmd/raftadmin/main.go
+++ b/cmd/raftadmin/main.go
@@ -20,21 +20,22 @@ import (
 )
 
 const (
-	defaultRPCTimeoutSeconds = 5
-	minCommandArgs           = 2
-	addVoterArgCount         = 2
-	addLearnerArgCount       = 2
-	promoteLearnerArgCount   = 1
-	transferArgCount         = 2
-	addVoterPrevIndex        = 2
-	addLearnerPrevIndex      = 2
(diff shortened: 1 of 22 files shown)


bootjp:
@claude review

bootjp:
@claude review

Addressed the eleven points from the previous review pass — full breakdown in the commit message of 08f59c76. Key items:

- §4.6 enumerates all four `len(e.peers)` overloads and introduces an apply-loop-owned `voterCount` + `isLearnerNode` cache so the 1-voter+1-learner lease-read stall is fixed at the call site, not just at `recordQuorumAck`. `refreshStatus` and `removePeer` use the same cache.
- §4.2 splits `e.config.Servers` ownership: `upsertPeer` keeps only the address cache; the apply loop rebuilds `e.config.Servers` from `configurationFromConfState` after every conf change.
- §4.3 documents the two-phase recovery and pins `confStateForPeers` as voters-only / bootstrap-only. `Peer.Suffrage` is a v2-file artifact, not a runtime authority — closes the divergence concern.
- §4.5 gives a concrete enforcement model for `--raftJoinAsLearner` (post-apply local alarm, not pre-propose veto) — closes the underspecification.
- §3.3 enumerates the `len(e.peers)` overloads at design time so future implementers don't discover them. Milestone 1 explicitly fixes all four in one PR.
- §5.5 adds a snapshot-during-learner-catch-up test; lease-read regression test split into the two named scenarios.
- §8 records the `min_applied_index = 0` footgun as an open question with the safer `skip_min_applied_check` alternative.
- §3.2 swaps line numbers for function-name + excerpt references.

bootjp:
@claude review

Addressed both new issues from Round 2 (commit dcf13373):

- **High — `validateConfState` voter-count mismatch on restart with learners**: §4.2 edit 3 now documents that `validateConfState` builds its own expected voter set filtered by `Suffrage != "learner"`. `confStateForPeers` stays minimal as Round 1 documented. (Resolution (a) from your two options.)

- **Medium — `e.voterCount` / `e.isLearnerNode` not initialized at Open**: §4.6 now has an explicit "Initialization at Open time" subsection that wires both fields adjacent to the existing `e.config` init from `prepared.disk.LocalSnap.Metadata.ConfState`. Closes the false single-node-fast-path activation on 3-voter startup.

- **Minor**: §4.6 calls out that `isLearnerNode` is rebuilt from scratch on each conf change, never patched — addresses the stale-true-after-promotion footgun. §8 numbering fixed (2/2a → 2/3).

bootjp:
@claude review

Addressed all three Round-3 items (commit 34c84dc4):

- **Minor — `validateConfState` Resolution (a) learner ordering**: switched the learner check to set-based comparison. No prior convention pins learner ordering in persisted state (learners have never been persisted before), and a set comparison is strictly more robust than introducing a new ordering contract on the v2 writer. Voter check stays element-wise to match the existing well-established invariant.

- **Minor — `applyConfigChange` / `applyConfigChangeV2` signatures**: chose option (a). §4.2 edit 4 now specifies the extended signature `(..., confState raftpb.ConfState)` mirroring `nextPeersAfterConfigChange`. Includes an explicit ordering note: suffrage-aware updates run after the per-peer cleanup helpers so the apply-loop ordering is deterministic.

- **§8 renumber**: items were 1, 2, 3, 3, 4 → now 1, 2, 3, 4, 5.

bootjp:
@claude review

Both Round-4 issues were genuine bugs in my prose. Fixed in commit a9b918f0:

- **High — set-based formula `&&` → `||`**: §4.2 edit 3 now states the error condition correctly: `len mismatch || !allLearnersPresent`. Includes the equivalent valid-condition form and an explicit note on the two mismatch cases the `&&` form would have silently accepted (same-count member divergence; conf-learner missing from peers).

- **Medium — ordering contradiction**: §4.2 edit 4 now spells out the apply-loop sequence as four explicit steps. The two suffrage-aware updates have different orderings:
  1. `voterCount` / `isLearnerNode` update **before** `upsertPeer`/`removePeer` (so `removePeer` reads the post-change count for its ack-tracker threshold — matches §4.6 and §5.2).
  2. `e.config.Servers` rebuild **after** `upsertPeer`/`removePeer` (so it reads the post-cleanup address cache).

  Step-numbered sequence makes this unambiguous for an implementer.

bootjp:
@claude review

Addressed CodeRabbit Round 1 (commit 2f3bff23):

**Real bug**:
- `shouldCampaignSingleNode` counted all servers including learners. A 1-voter+N-learner cluster on restart would skip the fast-path Campaign() and wait for electionTimeout. Now filters by `Server.Suffrage == SuffrageVoter`. Same class as the four `len(e.peers)` overloads from §3.3 — one I missed in M1.

**`min_applied_index = 0` footgun (closes design §8 open question 3)**:
- Proto: `RaftAdminPromoteLearnerRequest` gains `skip_min_applied_check` bool.
- Engine REJECTS `min_applied_index == 0` with `errPromoteLearnerMinAppliedZero` unless skip flag is set.
- `Admin.PromoteLearner` API gains `skipMinAppliedCheck` parameter.
- `raftadmin promote_learner` gains optional `skip_min_applied_check` positional arg.
- New `TestPromoteLearnerRejectsZeroMinAppliedWithoutSkip`.
- Design doc §8 marks open question 3 Resolved; runbook updated.

**Test polish**:
- `TestPersistedPeersV2RewritesAfterV1Read` → `TestPersistedPeersWriterAlwaysEmitsV2` (body never read v1).
- New `TestPersistedPeersV2UnknownSuffrageRejected` pins the 0xFF-rejection path.

**Doc polish**:
- §2 / §3.2 / §7 line-number refs replaced with stable function-name + excerpt refs.

**Skipped**:
- The "dead code in raftadmin server_test.go" finding was stale — review predated commit 3ed3c897 which adds the `AddLearner` / `PromoteLearner` proto RPCs and corresponding Server methods.
- Table-driven the two PromoteLearner negative tests — reviewer marked optional and noted standalone tests are fine.

bootjp:
@claude review

Round 2 caught a real high-priority bug. Fixed in commit 536a90a6:

- **Root cause**: `nextPeersAfterConfigChange` ignored the `confState` parameter, so AddLearner wrote the v2 peers file with `Suffrage=""` for every peer (`persistedSuffrageByte("")` → voter byte). On restart, `validateConfState` saw a learner-as-voter peers list and rejected startup with `errClusterMismatch`.

- **Fix at three persist sites** (the missing path was wider than just `nextPeersAfterConfigChange`):
  1. `nextPeersAfterConfigChange` / `nextPeersAfterConfigChangeV2` now use the confState parameter (no more `_`); new `annotatePeerSuffrageFromConfState` helper stamps `Suffrage` on the per-peer clone.
  2. `prepareOpenState` annotates the operator-provided peer slice from `disk.LocalSnap.Metadata.ConfState` before `savePersistedPeers` writes the bootstrap-time file. Without this, a joiner that wrote its peers file at Open would never get the suffrage update.
  3. `applyReadySnapshot` writes the v2 peers file with suffrage from the snapshot's ConfState — the joiner-via-snapshot path bypasses the apply-loop `nextPeersAfterConfigChange` hook entirely (the conf change is already incorporated in the snapshot's ConfState; no separate entry to apply).

- **New regression test** `TestLearnerPeersFilePersistsSuffrageAcrossRestart`: add a learner, propagate a write, verify both nodes' v2 peers files carry `Suffrage="learner"` for the learner peer, close + reopen both engines, assert no startup failure. This is the conformance test §5.5 promised but didn't have.

- **Pre-existing tests updated** (`TestNextPeersAfterConfigChange{,V2}*`) — they asserted `Suffrage=""` which was the pre-fix behavior; now assert the annotated suffrage.

- **Non-blocking observations**: skipped per the reviewer's framing (CLI usage hint, adminError mapping, JoinRoleViolationCount safety — all confirmed correct or acknowledged in the review).

bootjp:
Addressed the Codex finding (commit 32e20b15):

`alarmIfJoinedAsVoter` previously only ran from `applyConfigChange` / `applyConfigChangeV2` / `applyReadySnapshot`. A node that previously mis-joined as voter and is then restarted with `--raftJoinAsLearner=true` hits none of those paths — the post-mis-join `ConfState` is loaded from disk into `prepared.disk.LocalSnap` and never replayed in steady state.

**Fix**: `Open` now calls `engine.alarmIfJoinedAsVoter(prepared.disk.LocalSnap.Metadata.ConfState)` once after `refreshStatus`, before `run()` spins up. The existing `joinAlarmFired` latch (atomic.Bool) ensures the alarm fires at most once per process even if a later snapshot or conf change reapplies the same role assignment.

**Regression test**: `TestJoinAsLearnerAlarmFiresAtStartupForPersistedVoterRole` opens the joiner with `joinAsLearner=false`, has the leader `AddVoter` it, propagates a write so the conf change lands on the joiner's disk, closes it, then reopens with `joinAsLearner=true` and asserts `JoinRoleViolationCount` incremented and the engine stays running.

bootjp:
@claude review

PR #666, opened 2026-04-26
perf(redis): event-driven BZPOPMIN wake-up

bootjp:
## Summary

- BZPOPMIN's busy-poll loop was the leader's #3 CPU consumer in production (32% of the 30s pprof, on top of XREAD's 45% which #663 addresses). Same `time.Sleep(redisBusyPollBackoff)` pattern as XREAD BLOCK.
- Replace with an in-process notification: `keyWaiterRegistry` tracks BZPOPMINs by zset key; `ZADD` / `ZINCRBY` Signal after `coordinator.Dispatch` returns. The wait loop selects on `signal | fallback timer (100ms) | handlerCtx | deadline`.
- The `keyWaiterRegistry` type is intentionally generic, not zset-specific. When #663 (streams) consolidates onto this same type, `RedisServer` will hold one instance per domain.

## Design

- Independent of #663 — this PR introduces a new `keyWaiterRegistry` type alongside #663's `streamWaiterRegistry`. A small follow-up will consolidate the two onto the shared type.
- 100 ms fallback timer covers Lua-driven ZADD and follower-side applies (same as #663's stream side). Phase 2 — FSM `ApplyObserver` — addresses both gaps; tracked in `docs/design/2026_04_26_proposed_fsm_apply_observer.md` (in #663).

## CPU savings

- Signal-driven wake (leader ZADD → leader BZPOPMIN): one wake per ZADD instead of 100 polls/sec per blocked reader. Cuts BZPOPMIN's share of leader CPU from 32% to a small constant proportional to the ZADD rate.
- Fallback poll (10 ms → 100 ms): 10× reduction for follower-side and Lua-driven writers.

## Self-review (CLAUDE.md 5 lenses)

1. **Data loss** — None. Notifications are advisory; data still flows through Raft → FSM → Pebble. A lost signal is bounded by the 100 ms fallback.
2. **Concurrency** — `keyWaiterRegistry.mu` is held briefly: Register inserts under the lock; Signal copies the waiter set out of the lock and then non-blocking-sends. Sends are non-blocking so ZADD never stalls on a slow waiter. Registration completes before the first `tryBZPopMin` so a signal that fires between check and wait is preserved in the size-1 channel buffer.
3. **Performance** — Wait-loop overhead is one timer + one select per iteration (was: one `time.Sleep`). Hot-path ZADD/ZINCRBY adds one non-blocking channel send per Signal; Signal short-circuits when no waiters are registered.
[...]

PR #677, opened 2026-04-26
perf(redis): empty-zset fast path for BZPOPMIN signal-driven wakes

bootjp:
## Summary

After #666 deployed, n1 leader CPU dropped from 1600% to 371% (-77%), but BZPOPMIN was still 47.76% of the (smaller) profile. Per-wake breakdown: each `tryBZPopMin` runs `keyTypeAtExpect`, the zset prefix probe misses on the empty queue, and the slow-path fallback re-issues all ~19 `keyTypeAt` seeks just to confirm "still empty". On a busy queue with ZADD signals firing at the producer rate, this baseline grows linearly with the wake rate.

Add a fast path for the **signal-driven** wake case:

- `keyTypeAtExpectFast`: probes only the expected type's prefixes (3 seeks for zset). On miss, returns `redisTypeNone` immediately — no `rawKeyTypeAt` slow-path, no `hasHigherPriorityStringEncoding` guard.
- `tryBZPopMinFast`: `tryBZPopMin` variant using `keyTypeAtExpectFast`.
- `waitForBlockedCommandUpdate` now returns `true` iff the wake came from a Signal (false on fallback timer / shutdown).
- `bzpopminWaitLoop` tracks the wake reason: first iteration is full; signal-driven wakes flip to fast; fallback-timer wakes flip back to full.

The wrongType invariant is preserved because `zsetWaiters.Signal` only fires on `ZADD` / `ZINCRBY` — the only writes that produce zset rows. A wrongType-introducing write (`HSET`, `SET`, `XADD` etc.) does not signal, so the fallback timer's full check is the safety net that detects it within ~one `redisBlockWaitFallback` (100 ms).

## Behavior change

| Scenario | Before this PR | After this PR |
|---|---|---|
| BZPOPMIN on a pre-existing string key | WRONGTYPE on first iteration | WRONGTYPE on first iteration (unchanged) |
| BLOCK timeout on a missing key | nil after BLOCK budget | nil after BLOCK budget (unchanged) |
| ZADD wakes BZPOPMIN | Returns the entry | Returns the entry (unchanged; less CPU per wake) |
| HSET / SET arrives mid-BLOCK | WRONGTYPE within ~100 ms | WRONGTYPE within ~100 ms (unchanged ceiling) |

No correctness regression. The mid-block wrongType ceiling is the same 100 ms set by #666's fallback timer.

1. **Data loss** — None. Read-only fast path; `persistBZPopMinResult` is unchanged.
2. **Concurrency** — `waitForBlockedCommandUpdate`'s new bool return is set inside the same select that already serialised the three wake sources; no new shared state.
3. **Performance** — 16 fewer Pebble seeks per signal-driven wake on empty queues. The number of wakes per BZPOPMIN is unchanged — this is per-wake cost reduction, not wake-rate reduction.
4. **Data consistency** — Signal-driven wake reads at a fresh `readTS`; the fast probe and the slow probe see the same MVCC snapshot. WrongType detection on the slow path is preserved on initial iteration and on every fallback wake.
5. **Test coverage** — two new lockdown tests for the wrongType paths that cannot regress silently if the fast / full mode flag is wired incorrectly.

## Test plan

- [x] `go test -race -run "TestRedis_BZPopMin|TestKeyWaiterRegistry_|TestRedis_BullMQ" ./adapter/...` — passes
- [ ] CI: full adapter test suite
- [ ] Production profile after deploy to confirm BZPOPMIN drops below 47% on n1 leader

@claude review

### DIFF
--- a/adapter/redis_bzpopmin_wake_test.go
+++ b/adapter/redis_bzpopmin_wake_test.go
@@ -119,6 +119,82 @@ func TestRedis_BZPopMinWakesOnZIncrBy(t *testing.T) {
 // regression in the wait-loop refactor where the new
 // waitForBlockedCommandUpdate timer or context-cancel branch could otherwise
 // leak a -ERR reply.
+// TestRedis_BZPopMinRejectsInitialWrongType locks down the
+// first-iteration full-check invariant: when BZPOPMIN is issued
+// against a key that already holds a wrongType encoding (e.g. a
+// string), the very first tryBZPopMin must surface WRONGTYPE. The
+// signal-driven fast path must never run before the initial full
+// check has confirmed the type.
+func TestRedis_BZPopMinRejectsInitialWrongType(t *testing.T) {
+	t.Parallel()
+	nodes, _, _ := createNode(t, 3)
+	defer shutdown(nodes)
+
+	rdb := redis.NewClient(&redis.Options{Addr: nodes[0].redisAddress})
+	defer func() { _ = rdb.Close() }()
+	ctx := context.Background()
+
+	// Pre-write a string at the BZPOPMIN target key.
+	require.NoError(t, rdb.Set(ctx, "bzpop-wrongtype", "I am a string", 0).Err())
+
+	// BZPOPMIN with a 1 s budget — full check on the first iteration
+	// must catch the wrongType immediately, well under the BLOCK
+	// timeout.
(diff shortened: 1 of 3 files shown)


bootjp:
@claude review

bootjp:
@claude review

PR #682, opened 2026-04-26
test(keyviz): Phase 2-A bench coverage + concurrent-burst exact-count test

bootjp:
## Summary

Closes the "Benchmark gate green" half of the Phase 2-A exit criteria in `docs/admin_ui_key_visualizer_design.md` §12. Pure-test addition; no production code changes.

Four new benchmarks pin the cost shape of the `keyviz` package:

| Benchmark | What it pins | Apple M1 Max @ 200 ms |
|---|---|---|
| `BenchmarkObserveParallel` | Hot-path contention — distinct slot per goroutine, only shared work is the table `atomic.Pointer.Load` | 20.6 ns/op, 0 allocs |
| `BenchmarkRegisterRoute` | COW route-table mutation cost; flat ns/op would signal a shift to a shared mutable map | 123.9 µs/op, 24 allocs |
| `BenchmarkFlush` | Per-step drain over 1024 live routes; catches O(N²) scan regressions | 96.4 µs/op |
| `BenchmarkSnapshot` | Read-side pivot over 1024 routes × 64 columns; dominant term in admin handler latency | 1.86 ms/op |

`TestObserveExactCountUnderConcurrentBurst` — 32 routes × 8 writers × 4000 ops gated on a `sync.WaitGroup` so every goroutine starts simultaneously. Pins the Phase 2-A invariant that no counts are lost under concurrent burst: `sampleRate = 1` today, so every `Observe` must be reflected exactly in the post-Flush Snapshot. When the design §5.2 sub-sampling estimator lands, this test must be updated alongside the new ±5% / 95%-CI assertion.

## Why not the ±5% accuracy SLO test?

Per design §5.2, the SLO is for the Horvitz–Thompson estimator that runs when the adaptive controller raises `sampleRate` above 1 under burst. **That controller is not yet implemented** — the current code path always counts every Observe — so a ±5% / 95%-CI assertion would be testing the absence of a feature. This PR pins the current "exact counting" invariant instead; the SLO test lands with the sub-sampling implementation in a follow-up that updates §5.2 to reflect the chosen controller shape.

## Five-lens self-review

1. **Data loss** — n/a; tests only.
2. **Concurrency / distributed** — the burst test is the concurrency check; with race detector on (`go test -race`) it's clean.
3. **Performance** — adds CI-runnable benchmarks; baseline ns/op recorded above.
4. **Data consistency** — pins the exact-counting invariant explicitly so a regression that loses counts on the hot path is caught.
5. **Test coverage** — the burst test plus four benchmarks are the new coverage. Existing tests unchanged.

## Test plan

- [x] `go test -race -count=1 ./keyviz/...` — clean
- [x] `go test -bench='Benchmark(ObserveParallel|RegisterRoute|Flush|Snapshot)$' -benchtime=200ms -run='^$' ./keyviz/...` — runs and reports the numbers above
- [x] `golangci-lint run ./keyviz/...` — clean

## Summary by CodeRabbit

* **Tests**
  * Added performance benchmarks for concurrent operations, route registration, flush operations, and snapshot functionality
  * Added concurrency correctness test to verify data integrity and accurate counting under concurrent load

### DIFF
--- a/keyviz/sampler_test.go
+++ b/keyviz/sampler_test.go
@@ -1095,44 +1095,74 @@ func BenchmarkObserveMiss(b *testing.B) {
 }
 
 // BenchmarkObserveParallel pins the contention profile of the hot
-// path. Each goroutine hits a distinct slot so the only shared work
-// is the atomic.Pointer load of the route table; per-slot atomic
-// adds do not contend across goroutines. A regression that adds a
-// shared mutex (or a global counter) on the hot path will show up
-// as a sharp drop in ns/op as the parallelism rises.
+// path. Each goroutine is shard-pinned to a disjoint range of route
+// IDs so per-slot atomic adds genuinely never contend across
+// goroutines and the only shared work is the atomic.Pointer load of
+// the route table. A regression that adds a shared mutex (or a
+// global counter) on the hot path will show up as a sharp drop in
+// ns/op as parallelism rises.
+//
+// The earlier draft used a per-worker `var i uint64` counter which
+// always started at zero, so all workers walked the same route
+// sequence in lockstep — that defeated the "distinct slot" claim
+// and the benchmark would have masked false-sharing regressions
+// (Codex P2 / Gemini medium on PR #682).
 func BenchmarkObserveParallel(b *testing.B) {
-	const numRoutes = 64
+	const (
+		numRoutes      = 64
+		routesPerShard = 4
+	)
 	s := NewMemSampler(MemSamplerOptions{Step: time.Second, HistoryColumns: 4, MaxTrackedRoutes: numRoutes})
 	for r := uint64(1); r <= numRoutes; r++ {
 		if !s.RegisterRoute(r, []byte{byte(r)}, []byte{byte(r) + 1}) {
 			b.Fatalf("RegisterRoute(%d) returned false", r)
 		}
 	}
+	var nextShard atomic.Uint64
 	b.ReportAllocs()
 	b.ResetTimer()
 	b.RunParallel(func(pb *testing.PB) {
+		shardIndex := nextShard.Add(1) - 1
+		shardBase := (shardIndex * routesPerShard) % numRoutes
 		var i uint64
 		for pb.Next() {
+			s.Observe(shardBase+(i%routesPerShard)+1, OpWrite, 16, 64)
 			i++
-			s.Observe((i%numRoutes)+1, OpWrite, 16, 64)
 		}
 	})
 }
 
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

Round-1 follow-ups (19abb415) addressing all 4 items from Gemini + Codex on PR #682:

- **Codex P1 / route presence** — `TestObserveExactCountUnderConcurrentBurst` now indexes rows by RouteID and asserts every registered route is present, so a future Flush regression that silently drops a route is caught instead of being masked by a loop that just iterates fewer times.
- **Codex P2 / Gemini medium / parallel shard** — `BenchmarkObserveParallel` assigns each worker a disjoint shard of route IDs via `atomic.Add`. The previous draft had every worker start its counter at zero, so workers walked the same routes in lockstep. Now they actually hit distinct slots.
- **Gemini medium / RegisterRoute O(N²)** — fixed pre-load of 1024 routes; benchmark now churns one ID in/out per iteration. b.N controls only the iteration count, so reported ns/op is constant per-call rather than growing with N.
- **Gemini medium / Flush timed reseed** — `b.StopTimer` / `b.StartTimer` bracket the reseed loop. Reported cost dropped from 96.4 µs/op to 82.3 µs/op once the 1024 Observe calls were excluded.

New M1 Max baselines (200 ms benchtime):

| Bench | Before | After |
|---|---|---|
| ObserveParallel | 20.6 ns/op | 4.4 ns/op |
| RegisterRoute | 124 µs/op (growing with N) | 162 µs/op (constant) |
| Flush | 96.4 µs/op | 82.3 µs/op |
| Snapshot | 1.86 ms/op | 1.92 ms/op |

`go test -race -count=1 ./keyviz/...` clean. `golangci-lint run ./keyviz/...` clean.

bootjp:
@claude review

Round-2 follow-ups (3310a1a6):

- **P1 / nolint:gosec** — `runConcurrentBurst` now takes `numRoutes uint64`. Iterating in that type from the start eliminates the `int → uint64` conversion that triggered gosec G115; the suppression is gone.
- **P2 / shard count** — `BenchmarkObserveParallel` numRoutes bumped 64 → 256. With `routesPerShard = 4` that's 64 disjoint shards, so RunParallel keeps the distinct-slot guarantee up to GOMAXPROCS = 64. New M1 Max baseline at 200 ms benchtime: **3.5 ns/op** (was 4.4 ns/op; the shard increase actually frees up a bit more parallelism).

`go test -race -count=1 ./keyviz/...` clean. `golangci-lint run ./keyviz/...` clean.

PR #680, opened 2026-04-26
feat(admin): KeyViz heatmap page in web/admin SPA (Phase 2-B)

bootjp:
## Summary

Phase 2-B of the Key Visualizer design (`docs/admin_ui_key_visualizer_design.md`): integrate the heatmap into the existing `web/admin/` SPA rather than building a parallel dashboard. Doc + implementation in one PR (doc commit first, per CLAUDE.md design-doc-first workflow).

- New page at `/keyviz` — canvas heatmap polling `/admin/api/v1/keyviz/matrix` with series picker (writes / reads / write_bytes / read_bytes), row-budget input (clamped at 1024), and off / 5 s / 30 s auto-refresh.
- Cold cells (value 0) render as the page background, not a faint blue — spotting actually-cold routes stays the dominant visual signal.
- Row-detail flyout on hover: bucket_id, start, end, aggregate, route_count, route_ids (with truncation indicator). `start` / `end` decode through a printable-or-hex preview so binary keys do not render as mojibake.
- Dependency-free: hand-rolled five-stop colour ramp in `lib/colorRamp.ts`. No d3, no ECharts. Bundle grew from ~155 kB to ~208 kB (raw), 64 kB gzipped.

Backend is unchanged. The handler at `/admin/api/v1/keyviz/matrix` and its sampler wiring already shipped under Phase 2-A (PRs #639 / #645 / #646 / #647 / #651 / #660 / #661 / #672).

## Design

`docs/design/2026_04_27_proposed_keyviz_spa_integration.md` — proposed status, lands with this PR. Parent design §12 phase table is split into 2-A (server, shipped) / 2-B (SPA, this PR) / 2-C (cluster fan-out, open).

## Five-lens self-review

1. **Data loss** — n/a; SPA is read-only against an existing handler.
2. **Concurrency / distributed** — n/a; single browser tab polling a single handler. Sampler concurrency was already covered by Phase 2-A tests.
3. **Performance** — Canvas + per-cell `fillRect` runs under §10 budget at 1024 × 500. Auto-refresh defaults off; 5-second cadence is the lower bound (sampler flush is 1 s).
4. **Data consistency** — SPA renders whatever the handler returns; consistency guarantees come from the existing leader-issued counters in the sampler.
5. **Test coverage** — `tsc -b --noEmit` clean; `vite build` clean; `go build ./internal/admin/...` clean (embed glob unaffected); `go test ./internal/admin/...` clean. Manual verification documented in the design doc §5.

## Test plan

[...]

PR #684, opened 2026-04-27
fix(admin): KeyViz heatmap honours devicePixelRatio

bootjp:
## Summary

Phase 2-B follow-up to PR #680. Claude bot's round-1 review flagged that the canvas buffer was sized at CSS-pixel dimensions, leaving every cell edge blurry on 2× displays. Fix:

- Scale the canvas buffer to physical pixels via `window.devicePixelRatio`, keep the CSS `style` at logical pixels.
- Reset the transform via `ctx.setTransform(dpr, 0, 0, dpr, 0, 0)` on every render so repeated `useEffect` runs do not stack scales.
- Clamp the DPR at 4 so a browser reporting an absurd ratio (e.g. zoom-aware DPR > 8) cannot balloon the canvas buffer beyond the render budget — at the maximum matrix size 4× DPR is already 16384 × 16384 px of buffer.

## Five-lens self-review

1. **Data loss** — n/a; SPA-only render change.
2. **Concurrency / distributed** — n/a; single render path.
3. **Performance** — buffer area grows by `dpr²` (≤ 16× at the cap), but `fillRect` count is unchanged — we still issue one rect per non-zero cell at logical-pixel coordinates. Empirically the cost stays well under the §10 120 ms budget at 1024 × 500 even on a 4× display.
4. **Data consistency** — render is purely cosmetic; no data semantics change.
5. **Test coverage** — type check + Vite build clean. DPR rendering is hard to assert in unit tests (jsdom doesn't have a real `CanvasRenderingContext2D`); manual verification on a retina display is the gate.

## Test plan

- [x] `tsc -b --noEmit` clean
- [x] `vite build` clean
- [ ] Manual: open `/keyviz` on a retina display; cell edges crisp instead of blurry
- [ ] Manual: switch between retina and external 1× display in the same session; canvas re-renders correctly without scale stacking

### DIFF
--- a/web/admin/src/pages/KeyViz.tsx
+++ b/web/admin/src/pages/KeyViz.tsx
@@ -111,10 +111,22 @@ function Heatmap({ matrix }: HeatmapProps) {
   useEffect(() => {
     const canvas = canvasRef.current;
     if (!canvas) return;
-    canvas.width = width;
-    canvas.height = height;
+    // Scale the backing buffer to physical pixels and keep CSS at
+    // logical pixels: on a 2× display every cell edge is otherwise
+    // rendered against a half-resolution buffer and reads as blurry.
+    // We clamp the ratio so a future browser quirk reporting an
+    // absurd value (e.g. Firefox's experimental zoom-aware DPR > 8)
+    // does not balloon canvas memory beyond reason — at the maximum
+    // matrix size 4 × dpr is already 16384 × 16384 px of buffer.
+    const dpr = Math.min(window.devicePixelRatio || 1, 4);
+    canvas.width = Math.max(1, Math.floor(width * dpr));
+    canvas.height = Math.max(1, Math.floor(height * dpr));
     const ctx = canvas.getContext("2d");
     if (!ctx) return;
+    // setTransform (not scale) so the matrix is reset cleanly on
+    // every render — repeated useEffect runs would otherwise stack
+    // scales on top of the previous one.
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

Round-1 follow-ups (b10a0aed) addressing both Gemini items on PR #684:

- **Fractional DPR transform** — `ctx.setTransform` now uses `canvas.width / width` (and `canvas.height / height`) instead of the raw `dpr`. On 1.25× / 1.5× displays the buffer dimensions are floored but the transform was raw, so the renderer drew at sub-pixel coordinates the buffer could not represent. Using the actual ratio eliminates the residual blur.
- **DPR change listener** — added a `matchMedia` listener that bumps a `dprTick` state. The canvas effect now depends on it, so dragging the window between displays of different densities or zooming the browser triggers an immediate re-render rather than waiting for the next data refresh.

`tsc -b --noEmit` and `vite build` clean.

bootjp:
@claude review

Round-2 cosmetic follow-up (7a77b578): hoisted the empty-matrix early return ahead of `setTransform`. The live path now collapses to a single `ctx.setTransform(canvas.width / width, 0, 0, canvas.height / height, 0, 0)` with no `width > 0` ternary. The empty-matrix branch resets the transform first (in case the canvas was reused) before clearing — makes the dead-state cleanup explicit.

`tsc -b --noEmit` and `vite build` clean.

PR #685, opened 2026-04-27
docs: KeyViz cluster fan-out proposal (Phase 2-C)

bootjp:
## Summary

Phase 2-C proposal — cluster-wide KeyViz heatmaps via fan-out.

Phase 2-A and 2-B ship a heatmap that only renders the local node's view. Per parent design §5.1, reads are recorded by the follower that serves them and writes by the leader, so a single-node SPA shows a partial picture by construction. This proposal scopes a minimum-viable fan-out that ships the operator-visible value without yet requiring the proto extension parent §9.1 calls for.

## Key design decisions worth a careful read

- **Static `--keyvizFanoutNodes` flag** instead of `GetClusterOverview` discovery for the MVP (§3). The membership cache lands when clusters grow beyond hand-configured size.
- **Conservative max-merge for writes** with a `conflict` flag instead of the canonical `(bucketID, raftGroupID, leaderTerm, windowStart)` rule (§4.2). The canonical rule needs proto/JSON extensions and lands in Phase 2-C+. The MVP rule never overcounts (operators react badly to phantom hotspots) but may undercount during a leadership flip — surfaced via `conflict=true`.
- **Embedded `internal/admin` aggregator** instead of a standalone `cmd/elastickv-admin` binary (§2.2). The standalone binary was deferred indefinitely; the embedded path is the go-forward.
- **Forward-compatible wire format** (§5). Old SPA against fan-out server keeps working — the new fields are additive.

## What this proposal does NOT do

Listed in §2.2 to keep the PR-1 scope reviewable: no membership discovery, no proto extension, no standalone binary, no response caching. Each gets its own follow-up if and when needed.

## Implementation plan (per §7)

- **PR 1** — server-side aggregator + flag plumbing + table-driven test
- **PR 2** — SPA degraded banner + conflict hatching + header counter
- **PR 3 (Phase 2-C+)** — proto/JSON extension; replace max-merge with the canonical rule; promote `conflict` from per-row to per-cell

## Test plan

- [ ] Design review: confirm the §4.2 conservative max-merge trade is acceptable, or push back with the alternative (e.g., always sum + always conflict-flag if multiple nodes report > 0)
- [ ] Confirm `--keyvizFanoutNodes` flag shape (host:port list, self-included) is the right contract for now
- [ ] Confirm the timeout shape in §9 open question 2 (proposed: fixed 2 s with operator override)

### DIFF
--- a/docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md
+++ b/docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md
@@ -182,10 +182,15 @@ respectively (sum and max). Same justification.
 ### 4.4 Row identity
 
 Two rows from different nodes belong to the same logical row when
-their `BucketID` matches. `Start` / `End` / `Aggregate` /
-`RouteCount` / `RouteIDs` are taken from the **first non-empty
-node response** (deterministic by node order in
-`--keyvizFanoutNodes`). If two nodes disagree on `Start`/`End`
+their `BucketID` matches. After the aggregator has collected **all**
+per-node responses (i.e. waited for every peer's deadline to
+resolve), `Start` / `End` / `Aggregate` / `RouteCount` / `RouteIDs`
+are taken from the **lowest-indexed node in `--keyvizFanoutNodes`
+whose response is non-empty for the row**. Selection by config-list
+position rather than wall-clock arrival order is the load-bearing
+property: peers respond in non-deterministic order, so picking on
+first arrival would let `Start`/`End` flap between polls for the
+same bucket. If two nodes disagree on `Start`/`End`
 for the same `BucketID`, that indicates a routing-catalog
 divergence the operator should investigate; we surface it as a
 warning in the per-node status payload but do not block the
@@ -306,7 +311,8 @@ PR 3 (Phase 2-C+):
    would invent traffic) and the conflict flag tells operators
    when the data is soft.
 2. **Concurrency / distributed** — Per-node calls are issued in
-   parallel with a 5-second ceiling per node. The aggregator
+   parallel with the configured per-call timeout (default 2 s,
+   override via `--keyvizFanoutTimeout` per §3). The aggregator
    itself is single-goroutine (waits for all peer responses then
    merges). No shared mutable state between concurrent fan-out
    requests.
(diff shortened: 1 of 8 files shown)


bootjp:
@claude review

Round-1 follow-ups (1cd6ed52):

**Codex P2** — added per-node `warnings` array with `catalog_divergence` code in §5; gives the §4.4 catalog-mismatch signal an actual wire field.

**Gemini medium / consistency** — `conflict` flag scope now stated up front in §2.1 as row-level (with the per-cell upgrade explicitly tied to Phase 2-C+); removes the §2.1 vs §5/§6 mismatch.

**Gemini security-medium / auth** — rewrote §3 with a clean two-tier story:
- Phase 2-C MVP: anonymous fan-out on a private network with an explicit "do NOT enable across an untrusted network" warning.
- Phase 2-C+: signed inter-node token derived from the existing admin session-signing key, **NOT** a replay of the browser cookie. The earlier draft conflated the two; this paragraph supersedes it.

**Gemini medium / column drift** — §4.5 now bucket-aligns column timestamps to `keyvizStep` boundaries before merging, so sub-step NTP jitter does not split each window into two adjacent merged columns.

**Gemini security-medium / body cap** — §8 lens 3 documents the 64 MiB `io.LimitReader` cap (matches what PR #686 implements).

**Gemini medium / DNS** — §9 Q1 resolved: resolution rides the stdlib resolver and nscd cache; no aggregator-local cache. Per-call deadline already protects against transient DNS hangs. Operators on flaky DNS colocate a caching resolver — same answer for every short-lived HTTP client, not unique to this feature.

**Gemini high / rollout strategy** — pushed back in §1 "Rollout notes". This feature:
- Is off by default (`--keyvizFanoutNodes` empty)
- Lives entirely on the admin path (no data-plane impact)
- Stores nothing on disk (no schema migration)
- Has a forwards-compatible wire format (old SPA + new server, and old node + new aggregator, both work)

A dual-write proxy / blue-green isn't a fit for this scope; mixed-version rollout is correct by construction. The worst-case partial deploy surfaces as a per-node `ok=false` in the response, not data loss.

Open to pushback on any of these calls if the trade-offs aren't right.

bootjp:
@claude review

Round-2 follow-ups (d012f4c2) addressing all 7 doc-level items:

1. **§8 lens 3 / cap math** — replaced `32 B/uint64 ≈ 128 MiB raw, ~32 MiB JSON` (inconsistent: JSON doesn't compress) with `1024 × 4096 × 8 B/uint64 ≈ 32 MiB raw; 64 MiB cap = 2× headroom`.
2. **§8 lens 3 / MaxHistoryColumns** — explicit reference to `MaxHistoryColumns = 100 000` and the ~800 MiB raw upper bound that trips the cap on extreme ring queries.
3. **§8 lens 2 / timeout** — replaced `5-second ceiling` with `configured per-call timeout (default 2 s, override via --keyvizFanoutTimeout per §3)`.
4. **§9 Q2 / mark resolved** — restated as **Resolved** with the same shape as Q1, citing §3.
5. **§4.4 / first-non-empty precision** — explicit "after all per-node responses have been collected" + selection by lowest-indexed-in-flag rather than wall-clock arrival. Notes why (peers respond out-of-order).
6. **§8 lens 5 / scenario count** — was "four scenarios" with no fourth listed; now lists the three §7 PR-1 scenarios plus the url-builder / per-node-order / over-cap tests.
7. **§8 lens 3 / over-cap behavior** — clarified that `io.LimitReader` causes a JSON decode error (EOF mid-stream), recorded as `ok=false` with the error in that node's status entry plus a WARN-level log. No partial data accepted.

Q3 (per-node `generated_at`) intentionally still open — deferred to PR 2 (SPA UI).

bootjp:
@claude review

Round-3 follow-ups (0b5bfa03):

1. **§3 self-detection** — documented the matching shape: exact host:port match vs `--adminAddress`, plus wildcard-bind/loopback handling (0.0.0.0:8080 listener with 127.0.0.1:8080 in the flag still self-skips). Reverse-proxy / DNS-aliased entries fall through to a harmless loopback call. The behavior matches the implementation in PR #686 (`isSelfFanoutNode`).
2. **§5 conflict field** — explicit "always present, defaults false in local-only mode" so an SPA that pattern-matches `"conflict" in row` rather than the value doesn't regress when `--keyvizFanoutNodes` flips.
3. **§4.5 keyvizStep_ms** — defined in-section as `time.Duration(--keyvizStep) / time.Millisecond` so the formula is self-contained.
4. **§8 lens 3 sizing prose** — replaced "similar size for typical small values" with the precise statement: decimal encoding is **smaller** than 8-byte binary for ~4-digit counters; cap only becomes a real ceiling for traffic-saturated heatmaps with ~20-digit counters.

bootjp:
@claude review

Round-4 follow-up (30351811): tightened the `conflict` predicate in §2.1 and mirrored the clarification in §4.2.

The predicate is now stated as **"≥ 2 nodes report a non-zero value for the same cell"**, not "any disagreement". Follower zeros against a leader's non-zero value do **not** count — that's the steady-state shape, not a conflict. This matches the implementation in PR #686 (`maxMerge` ignores zero on either side and returns `conflict=false`) and the §7 "stable-leader: max wins, no conflict" test expectation.

Both §2.1 (scope summary) and §4.2 (merge-rule detail) now state the rule explicitly so future readers can't reach the looser interpretation.

bootjp:
@claude review

PR #686, opened 2026-04-27
feat(admin): KeyViz cluster fan-out aggregator (Phase 2-C PR-1)

bootjp:
## Summary

Server-side half of Phase 2-C cluster fan-out, per `docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md` (PR #685). With `--keyvizFanoutNodes` set, the existing `/admin/api/v1/keyviz/matrix` handler aggregates the local matrix with peer responses; with the flag empty (default), behaviour is unchanged.

This is **PR-1** of the implementation plan in §7 of the design. PR-2 (SPA degraded banner + conflict hatching) and PR-3 (proto extension with `raftGroupID` + `leaderTerm` for the canonical merge) are separate.

**Depends on PR #685** (the design doc) for review context — branched off that branch.

## Aggregator (`internal/admin/keyviz_fanout.go`)

- Parallel HTTP GETs to each peer with a per-call timeout (default 2 s, `--keyvizFanoutTimeout` overrides).
- Forward-compatible request shape: the per-peer URL is rebuilt from the parsed parameters, so a peer running an older or newer server cannot trip on an unrecognised parameter.
- **Merge rules per design §4**:
  - `reads` / `read_bytes` → sum across nodes
  - `writes` / `write_bytes` → max with row-level `Conflict` flag when inputs disagree
- Degraded mode: a peer that times out, 5xx's, or fails to decode contributes `FanoutNodeStatus{OK: false, Error: ...}`. The local matrix still ships.

## Wire format additions (forward-compatible)

```jsonc
{
  "rows": [{
    "...": "...",
    "conflict": false        // NEW: row-level disagreement flag
  }],
  "fanout": {                // NEW: omitted when fan-out is disabled
    "nodes": [{"node":"...", "ok":true, "error":""}],
    "responded": 2,
    "expected": 3
  }
}
```

Old SPA against fan-out server keeps working — extra fields are additive.

## Flag plumbing

- `--keyvizFanoutNodes` accepts comma-separated `host:port` endpoints.
- `--keyvizFanoutTimeout` sets the per-peer ceiling (default 2 s).
- Self-filter handles the bind-vs-advertise asymmetry: `0.0.0.0`/`::` bound listener matched against `127.0.0.1`/`localhost`/`::1` entries; a symmetric `--keyvizFanoutNodes=node1,node2,node3` stamped on every host does not loop back over HTTP.

## Five-lens self-review

1. **Data loss** — Read-only against the existing sampler; conservative max-merge can under-count during a leadership flip but never over-counts. Conflict flag tells operators when the cell is soft.
2. **Concurrency / distributed** — Per-peer calls run in parallel with a context deadline; the aggregator itself waits for all responses synchronously then merges. Test `TestKeyVizFanoutRunPeerTimeout` confirms a hanging peer cannot extend the wall time.
3. **Performance** — Aggregator cost is O(N × M × C); at the 1024-row × 60-column × 3-node ceiling that's ~180k cells per request — well under the SPA's existing render budget. Hot path (`Observe`) is untouched.
4. **Data consistency** — Merge rules conservative under leadership transitions. Reads exact in steady state and during transitions. The §9.1 canonical rule is preserved as a Phase 2-C+ contract.
5. **Test coverage** — Table-driven over the four canonical scenarios in the design + URL-builder variants + per-node order. `go test -race -count=1 ./internal/admin/...` clean; `golangci-lint run ./...` clean.

## Test plan

- [x] `go test -race -count=1 ./internal/admin/...`
- [x] `go test -race -count=1 .` (main_admin tests with the new `keyVizFanoutConfig{}` arg)
- [x] `go build ./...`
- [x] `golangci-lint run ./...`
- [ ] Manual: 3-node cluster with `--keyvizFanoutNodes=...` on each, drive write traffic against one node, confirm the heatmap reflects writes regardless of which node the SPA hits
- [ ] Manual: kill one node, confirm the response carries `responded: 2, expected: 3` plus the failed node's error string

## What this PR does NOT do

- **SPA changes**. Wire-format additions are forwards-compatible; the SPA renders correctly without knowing about `Fanout` or `Conflict`. UI for those lands in PR-2.
- **`raftGroupID` / `leaderTerm` proto extension**. PR-3 of the plan.
- **Membership discovery** via `GetClusterOverview`. Static `--keyvizFanoutNodes` is the contract for Phase 2-C; cache + refresh-interval lands when clusters grow beyond hand-configured size.
- **Inter-node auth**. The fan-out path is currently anonymous; intra-cluster admin traffic is assumed to ride a private network. Auth follows in a follow-up before this hits production.

### DIFF
--- a/internal/admin/keyviz_fanout.go
+++ b/internal/admin/keyviz_fanout.go
@@ -1,0 +1,463 @@
+package admin
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"net/url"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	pkgerrors "github.com/cockroachdb/errors"
+	"github.com/goccy/go-json"
+)
+
+// errKeyVizPeer is the sentinel wrapped by every fan-out failure so
+// callers can errors.Is() against it without parsing strings. Each
+// concrete failure adds its own %w-wrapped detail.
+var errKeyVizPeer = errors.New("keyviz fan-out peer error")
+
+// keyVizFanoutDefaultTimeout matches the design 9 open-question 2
+// proposed default: 2 seconds per peer call. Operators on weird
+// networks can override via WithTimeout.
+const keyVizFanoutDefaultTimeout = 2 * time.Second
+
+// keyVizPeerErrorBodyLimit caps how many bytes of a peer's non-OK
+// response body we splice into the error message. 512 is enough to
+// surface a typical structured error envelope without letting a
+// misbehaving peer flood operator logs.
+const keyVizPeerErrorBodyLimit = 512
+
+// keyVizMergeBucketHint is a hand-tuned starting capacity for the
+// merge phase's bucket map / order slice. Most fan-out responses
+// are well under 1024 rows; 64 lets a small cluster avoid the
+// initial map grow while keeping the worst-case overhead trivial
+// against the 1024-row budget.
+const keyVizMergeBucketHint = 64
+
+// FanoutResult is the per-response fan-out summary attached to
+// KeyVizMatrix.Fanout when fan-out is enabled. Nodes is ordered by
+// the operator-supplied node list (self first) so the SPA can render
+// a stable row order; Responded counts ok=true entries; Expected is
+// the configured peer count plus self.
+//
+// See docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md 5.
+type FanoutResult struct {
+	Nodes     []FanoutNodeStatus `json:"nodes"`
+	Responded int                `json:"responded"`
+	Expected  int                `json:"expected"`
+}
+
+// FanoutNodeStatus is one node's contribution status for a single
+// fan-out request. OK=true means the node returned a parseable
+// matrix; OK=false carries the reason (timeout, refused, 5xx body,
+// JSON decode failure). The local node always reports OK=true: its
+// matrix is computed in-process and cannot fail in this layer.
+type FanoutNodeStatus struct {
+	Node  string `json:"node"`
+	OK    bool   `json:"ok"`
+	Error string `json:"error,omitempty"`
+}
+
+// KeyVizFanout aggregates this node's local matrix with matrices
+// fetched from a static peer list. The contract:
+//
+//   - peers must NOT include self; the handler computes the local
+//     matrix and passes it to Run alongside the peer set.
+//   - Each peer is queried in parallel via HTTP GET on the same
+//     /admin/api/v1/keyviz/matrix path. The query string is rebuilt
+//     from the parsed parameters so a peer running an older or newer
+//     server does not receive an unrecognised parameter we never
+//     intended to forward.
+//   - A peer that times out, errors, or returns a non-OK status
+//     contributes a FanoutNodeStatus{OK: false, Error: ...} but does
+//     not abort the request. Aggregation proceeds with whatever
+//     succeeded.
+//
+// The merge rules are documented in 4 of the design doc:
+//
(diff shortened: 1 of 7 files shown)


bootjp:
@claude review

Round-1 follow-ups (97e09a86) addressing both Gemini items on PR #686:

- **security-high (line 274)** — JSON decoder now reads through `io.LimitReader` at 64 MiB. Sizing per the design's worst case (1024 rows × 4096 columns × ~32 B per uint64 ≈ 128 MiB raw / ~32 MiB JSON-encoded). Over-cap responses log a warning via the existing `KeyVizFanout.logger` rather than silently truncating. New `TestKeyVizFanoutRunPeerOverlargeBody` pins the in-cap path.
- **medium (line 266)** — `resp.Body.Close()` errors now log at debug level on the same logger; the previously-unused `logger` field is wired into both the close path and the size-cap warning.

`go test -race -count=1 ./internal/admin/...` clean. `golangci-lint run ./internal/admin/...` clean.

bootjp:
@claude review

Round-2 follow-ups (2addffe9):

- **`peerResponseExceededLimit` false-negative** — replaced the LimitReader-plus-external-probe pattern with the `countingReader` you suggested. Increments on every Read (including the chunks `json.NewDecoder` buffers internally), so the post-decode `n > cap` check fires deterministically regardless of where the decoder happened to stop. Security bound unchanged (cap+1 max bytes off the wire); only the overshoot **detection** is now reliable.
- **Over-cap test gap** — added `TestKeyVizFanoutRunPeerExceedsBodyLimit` driven by a new `WithResponseBodyLimit` option. Test lowers the cap to 1 KiB, serves a body well over that, and asserts the call completes promptly with per-node status surfaced.
- **Sizing-comment inconsistency** — rewrote the `keyVizPeerResponseBodyLimit` comment to be honest about the worst case: 1024 × 4096 × 20 B (max-encoded uint64) ≈ 80 MiB, which exceeds the 64 MiB cap. The operator-visible failure mode is "warning logged, matrix may be truncated", not DoS. A future flag exposes the cap when an extreme-traffic deployment actually trips it.

`go test -race -count=1 ./internal/admin/...` clean. `golangci-lint run ./internal/admin/...` clean.

bootjp:
@claude review

Round-3 follow-ups (4538f1dd) addressing both round-2 minor notes:

- **`//nolint:wrapcheck` rationale** — kept the suppression but extended the comment to explain why refactoring is impossible here: `json.Decoder` uses `err == io.EOF` (pointer compare, not `errors.Is`) to decide whether to keep pulling, so wrapping the io.Reader error with `%w` would silently break the decoder's stop condition. The CLAUDE.md "avoid nolint, refactor instead" rule has a genuine exception here.
- **Warning-fires test gap** — added `TestKeyVizFanoutRunPeerNearCapSucceedsWithWarning`: small JSON envelope (~30 B) + 256 B of trailing whitespace against a 100 B cap. `json.Decoder` reads in bufio chunks so the LimitReader hands it cap+1 = 101 bytes; decoder sees the complete object and returns nil, leaving `cr.n == 101 > 100`. The warning condition is now reachable from a unit test. The log line itself is not asserted (no clean slog injection point on the package), but the byte-count reliability is pinned.

`go test -race -count=1 ./internal/admin/...` clean. `golangci-lint run ./internal/admin/...` clean.

bootjp:
@claude review

Round-3 follow-up (186b6e97 after rebase on updated #685): generalized the `//nolint:wrapcheck` comment per your imprecision note. Now states the constraint as the io.Reader contract (sentinels must be passed unwrapped so `err == io.EOF` pointer compares keep working), with `encoding/json` and `goccy/go-json` mentioned as illustrative consumers — the underlying rule holds regardless of which decoder is in scope.

`go test -race -count=1 ./internal/admin/...` clean. `golangci-lint run ./internal/admin/...` clean.

PR #689, opened 2026-04-27
feat(admin): /admin/healthz/leader for load-balancer leader probes

bootjp:
## Summary

- Add `/admin/healthz/leader` so an upstream load balancer (Caddy, HAProxy, ...) can route admin traffic only to the current Raft leader, skipping the follower→leader AdminForward proxy hop on writes.
- Mirrors the `/healthz/leader` endpoints already shipped on the S3 (`adapter/s3.go:serveS3LeaderHealthz`) and DynamoDB (`adapter/dynamodb.go:serveDynamoLeaderHealthz`) adapters — same semantics, same body shape, same method allowlist — so a multi-protocol load balancer probing any of the three sees identical responses.
- New `LeaderProbe` interface in `internal/admin`; `NewRouterWithLeaderProbe` constructor (`NewRouter` delegates with `nil` probe, preserving existing callers); `ServerDeps.LeaderProbe`; `newAdminLeaderProbe` in `main_admin.go` consults the local Raft group runtimes.

## Behaviour matrix

| Path                       | Probe state        | Response                   |
|----------------------------|--------------------|----------------------------|
| `/admin/healthz`           | (n/a)              | 200 `ok`                   |
| `/admin/healthz/leader`    | wired, leader      | 200 `ok`                   |
| `/admin/healthz/leader`    | wired, follower    | 503 `not leader`           |
| `/admin/healthz/leader`    | not wired          | 404 `not_found` (JSON)     |
| `/admin/healthz/leader` POST | (any)            | 405 `method_not_allowed`   |

1. **Data loss** — None. Read-only handler; no Raft proposals.
2. **Concurrency** — `VerifyLeader` is bounded at 2 s, called only on nodes that already claim leadership via `Status()`. No new shared state.
3. **Performance** — Cheap path: followers short-circuit via `Status()` (no Raft round-trip). Leader path: one ReadIndex per probe per group; load balancers typically probe every few seconds, so the overhead is bounded and well below the existing background HLC lease + ReadIndex traffic.
4. **Data consistency** — `VerifyLeader` is the same ReadIndex round-trip the lease-read path uses; a stale-leader follower mid-silent-leadership-change cannot return 200.
5. **Test coverage** — six router-level tests: 200-when-leader, 503-when-follower, HEAD on both branches, POST rejection, nil-probe → 404, classify ordering (path is not swallowed by the SPA fallback).

## Test plan

- [x] `go test -race ./internal/admin/...` — passes (1.3s)
- [x] `go test -race -run TestStartAdminServer .` — passes
- [x] `golangci-lint run ./...` — clean
- [ ] CI: full suite under `-race`

@claude review

### DIFF
--- a/internal/admin/router.go
+++ b/internal/admin/router.go
@@ -26,17 +26,48 @@ import (
 // versions (`/admin/api`, `/admin/api/v2`, ...) return a JSON 404
 // instead of being silently answered with the SPA HTML.
 const (
-	pathPrefixAdmin   = "/admin"
-	pathAPIRoot       = "/admin/api"
-	pathPrefixAPI     = pathAPIRoot + "/"
-	pathAPIv1Root     = "/admin/api/v1"
-	pathPrefixAPIv1   = pathAPIv1Root + "/"
-	pathHealthz       = "/admin/healthz"
+	pathPrefixAdmin = "/admin"
+	pathAPIRoot     = "/admin/api"
+	pathPrefixAPI   = pathAPIRoot + "/"
+	pathAPIv1Root   = "/admin/api/v1"
+	pathPrefixAPIv1 = pathAPIv1Root + "/"
+	pathHealthz     = "/admin/healthz"
+	// pathHealthzLeader is a sibling of pathHealthz for load balancers
+	// that want to route admin traffic only to the current Raft leader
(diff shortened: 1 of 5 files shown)


bootjp:
@claude review

bootjp:
@claude review

PR #692, opened 2026-04-27
fix(admin): forward session cookie on KeyViz fan-out so peers do not 401

bootjp:
## Summary

**Hotfix for the screenshot a user shared after enabling fan-out** — the heatmap collapsed to "1 of 6 nodes responded" with every peer returning `401 missing session cookie`.

**Root cause**: PR #686 shipped fan-out with the design assumption (#685 §3) that peers would accept anonymous calls on a private network. But the receiving side runs the same `SessionAuth` middleware as the browser path, so anonymous fan-out calls are rejected with 401. The "anonymous on a private network" framing was wrong: the implementation didn't put a bypass on the peer side, so anonymous calls never went through end-to-end.

**Fix**: forward the inbound user's cookies on every peer call. Cluster nodes already share `--adminSessionSigningKey` for HA, so a cookie minted on node A is verifiable on node B — no new infrastructure needed.

## What changed

- `KeyVizFanout.Run` gains a `cookies []*http.Cookie` parameter; nil preserves the legacy behaviour for tests / future non-browser callers.
- `KeyVizHandler` passes `r.Cookies()` so a SPA poll forwards `admin_session` (and `admin_csrf`, harmless on a GET) transparently.
- `TestKeyVizFanoutRunForwardsCookies` pins that both cookies reach the peer verbatim.
- Design doc §3 rewritten: documents the cookie-forwarding scheme + the operator-trust assumption that `--keyvizFanoutNodes` points at trusted hosts only. Phase 2-C+ will replace cookie forwarding with a dedicated inter-node token to decouple browser-session expiry from inter-node call validity.

## Five-lens self-review

1. **Data loss** — n/a; auth path change.
2. **Concurrency / distributed** — Cookies are passed by reference into per-peer goroutines but the slice is read-only inside `fetchPeer` (each goroutine `req.AddCookie`s independently). No shared mutation.
3. **Performance** — Adds N `req.AddCookie` calls per fan-out request. Trivial.
4. **Data consistency** — Auth-only change; merge semantics unchanged.
5. **Test coverage** — New `TestKeyVizFanoutRunForwardsCookies` asserts both cookies reach the peer with the original values. Existing fan-out tests pass `nil` and continue to exercise the unauthenticated path the existing code handled (peer 401s, request reports `ok=false`).

## Test plan

- [x] `go test -race -count=1 ./internal/admin/...` — clean
- [x] `golangci-lint run ./internal/admin/...` — clean
- [ ] Manual: 6-node cluster with shared `ELASTICKV_ADMIN_SESSION_SIGNING_KEY`, `--keyvizFanoutNodes` set on every node. Heatmap should now show "6 of 6 nodes" instead of "1 of 6".

## Trust / threat model note

Cookie forwarding is safe when (a) every peer is operator-configured and trusted, and (b) the network is private. **Do NOT point `--keyvizFanoutNodes` at an untrusted host** — the user's admin session would be replayed there. The design doc §3 update is explicit about this. Phase 2-C+ removes this footgun by switching to a dedicated inter-node token.

Closes the production-impacting symptom from screenshot review.

### DIFF
--- a/docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md
+++ b/docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md
@@ -132,12 +132,22 @@ elastickv \
   network); the HTTP client uses `http://`. A follow-up will
   introduce `--keyvizFanoutTLS` once the rest of the admin path
   has TLS too.
-- **Auth (Phase 2-C MVP)**: the aggregator forwards the inbound
-  user's `admin_session` cookie (and any other cookies the request
-  carries) on every peer call. The peer's `SessionAuth` middleware
-  verifies the cookie against its own `--adminSessionSigningKey`;
+- **Auth (Phase 2-C MVP)**: the aggregator forwards a whitelist of
+  the inbound user's cookies (`admin_session` + `admin_csrf` only)
+  on every peer call. The peer's `SessionAuth` middleware verifies
+  `admin_session` against its own `--adminSessionSigningKey`;
   cluster nodes already share that key for HA, so a cookie minted
   on node A is verifiable on node B without any new infrastructure.
+  Unrelated cookies the browser may carry (analytics, feature
+  flags, other-app sessions on the same domain) are dropped at the
+  fan-out boundary so they are not leaked across the internal
+  network (Gemini security-medium on PR #692).
+- **Recursion guard**: peer requests carry an `X-Admin-Fanout-Peer`
+  header. The receiving handler short-circuits its own fan-out
+  when this header is set; without the guard, a symmetric
+  configuration (every node lists every other node) would generate
+  O(N²) HTTP calls per browser poll as each peer recursively
+  fanned out. (Claude bot P1 on PR #692.)
   - The earlier draft of this paragraph said "anonymous on a
     private network" — that was wrong: the receiving side enforces
     session auth, so anonymous calls are rejected with 401 and the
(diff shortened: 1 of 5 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

Round-1 follow-ups (3c423787) addressing all 3 items:

- **Claude bot P1 / recursion storm** — Added `X-Admin-Fanout-Peer` marker header set by `fetchPeer`; `KeyVizHandler.ServeHTTP` short-circuits its own fan-out when present. Without this, a symmetric `--keyvizFanoutNodes` configuration would have produced O(N²) peer HTTP calls per browser poll once cookies started actually reaching peers. New `TestKeyVizHandlerSkipsFanoutForPeerCall` pins the guard with a peer stub that fails the test if dialled.
- **Gemini security-medium / cookie whitelist** — `attachAdminCookies` helper centralises the whitelist: only `admin_session` and `admin_csrf` are forwarded; unrelated cookies (analytics, feature flags, other-app sessions on the same domain) are dropped at the fan-out boundary. `TestKeyVizFanoutRunForwardsCookies` now also stages an `unrelated_app_session` cookie and asserts it does **not** reach the peer.
- **Claude bot P2 / nil-guard** — `TestKeyVizFanoutRunForwardsCookies` now `require.NotNil(merged.Fanout)` and `require.Len(Nodes, 2)` before indexing, so a regression surfaces with a precise message rather than a panic.

Also extended the test to assert `keyVizFanoutPeerHeader == "1"` on the peer-side, locking the recursion-guard contract end-to-end.

Design doc §3 updated to document both the whitelist and the recursion-guard behavior so the Phase 2-C MVP description matches the code.

`go test -race -count=1 ./internal/admin/...` clean. `golangci-lint run ./internal/admin/...` clean.

PR #699, opened 2026-04-28
ci: remove Claude Code OAuth diagnostic steps

bootjp:
## Summary

Remove the temporary OAuth diagnostic steps added in #698. They isolated the failure as intended — the `Run Claude Code` step is fine on this side and the bug is upstream.

## Diagnostic outcome (run [25057882769](https://github.com/bootjp/elastickv/actions/runs/25057882769))

| Step | Result | Meaning |
| --- | --- | --- |
| Install Claude CLI for diag | ✅ | runner can fetch / install the CLI |
| Verify OAuth on the runner | ✅ | on the same runner, with the same token via env, `claude --print "ok"` returns a normal Claude response |
| Run Claude Code (action / agent SDK) | ❌ | same token, same env, throws `Could not resolve [authentication]` |

So the failing path is **only** through `anthropics/claude-code-action` → `@anthropic-ai/claude-agent-sdk`'s child-process auth forwarding. The token, the Anthropic OAuth backend, the runner network, and step-level env propagation are all healthy.

## Changes

- Drop both diag steps (`Install Claude CLI for diag` + `Verify OAuth on the runner`).
- Replace the comment block on `Run Claude Code` to record the diagnostic conclusion and the workarounds that did NOT help, so the next person doesn't redo the same investigation.

## Kept as-is (deliberately)

- `uses: ...@<v1.0.107 SHA>` pin
- duplicated `env: CLAUDE_CODE_OAUTH_TOKEN` on the step

Neither fixes the issue, but reverting them now is churn we would undo again when upstream ships a fix. Lift both in a single follow-up at that point.

## Operational follow-ups (out of scope for this PR)

- Rotate `CLAUDE_CODE_OAUTH_TOKEN` (insurance, since the diag step did briefly carry the token through `pbpaste | docker stdin` in local testing too — the run logs themselves did not leak it).
- File an upstream issue against `anthropics/claude-code-action` with run 25057882769 as the minimal repro. Not done in this PR; can be a separate task.
- If `@claude review` recovery is urgent before upstream fixes things, switch to `ANTHROPIC_API_KEY` auth (separate billing). Also out of scope here.

## Test plan

- [ ] After merge, `@claude` on a PR will continue to fail until upstream fixes the SDK auth forwarding. That is the expected state — the workflow file is back to its pre-diagnostic shape and waiting.

## Self-review

CI workflow only. 5-lens collapses to: no data-loss / concurrency / consistency / perf surface. Test coverage N/A (this is the workflow that runs review tooling, not code). Reverts a known-good prior state (apart from the comment refresh).

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -30,60 +30,30 @@ jobs:
         with:
           fetch-depth: 1
 
-      # TEMPORARY diagnostic. Verifies whether the OAuth token works against
-      # `claude --print` directly on the runner (i.e., bypassing the action /
-      # agent SDK entirely), so we can localize the failure to either:
-      #   - SDK auth-forwarding (diag passes, main fails), or
-      #   - runner ↔ Anthropic OAuth backend (diag fails, main fails).
-      #
-      # Split into two steps so the OAuth token is never present in the env
-      # of `curl | bash` or its subprocesses (Codex P2): the installer runs
-      # without the secret; only the verify step has it, and only for the
-      # single claude invocation. Both diagnostic steps are
-      # `continue-on-error: true` so neither a transient install failure nor
-      # a runner-side OAuth failure can short-circuit `Run Claude Code` —
-      # both step results need to be observed for the isolation logic to
-      # work, and temporary diagnostic infrastructure must not block normal
-      # `@claude` handling (Codex P1, both rounds).
-      #
-      # Token-leak hardening:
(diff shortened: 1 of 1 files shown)


bootjp:
Superseded by #700, which installed the Claude Code GitHub App and replaced this workflow on main. The original failure during 04-28 was actually intermittent App install/uninstall on the repo (not an upstream regression as this PR's body theorized). Closing.

Issue #12, opened 2023-12-05
DynamoDB API Impl


bootjp:
https://docs.aws.amazon.com/ja_jp/amazondynamodb/latest/developerguide/Programming.LowLevelAPI.html

PR #706, opened 2026-04-29
docs(design): propose per-adapter logical backup format

bootjp:
## Summary

Adds `docs/design/2026_04_29_proposed_logical_backup.md`, a proposal to expose elastickv backups as a per-adapter, decomposed directory tree instead of the opaque FSM snapshot stream. Backups become recoverable **without elastickv** — `cat`, `jq`, `find`, `aws s3 sync`, `redis-cli --pipe` are enough to extract any single record.

## Layout (top level)

- `dynamodb/<table>/_schema.json` + `items/<pk>/<sk>.json` (one wire-format DynamoDB item per file).
- `s3/<bucket>/<original/object/path>` — the raw assembled object body at its natural path, with a `<obj>.metadata.json` sidecar (`Content-Type`, ETag, user metadata, …). `aws s3 sync` works on the dump directory directly.
- `redis/db_<n>/{strings,hashes,lists,sets,zsets,streams,hll}/<key>.{bin,json,jsonl}` — strings stay raw bytes (binary-safe), collections are JSON, streams are JSONL in stream order.
- `sqs/<queue>/_queue.json` + `messages.jsonl` — one record per line, ordered by `(send_ts, sequence_number, message_id)`.
- `MANIFEST.json` and `CHECKSUMS` at the root; `MANIFEST.json` is the only file a restore tool must read first.

## Key design choices

- **Strips internal magic prefixes** (`storedDynamoItemProtoPrefix` `0x00 'D' 'I' 0x01`, `storedSQSMsgPrefix` `0x00 'S' 'M' 0x01`, `EKVPBBL1`) and the wide-column key encodings (`!hs|fld|<keyLen>...`, `!s3|blob|<bucket><gen><obj>...`, `!ddb|item|...`). The dump format is decoupled from internal layouts that have changed before and will change again.
- **Cluster-wide consistent `read_ts`** via the lease-read path; held open via `kv/active_timestamp_tracker.go` so the compactor cannot retire MVCC versions the dump still depends on.
- **Internal-state policy** is explicit: re-derivable indexes (DynamoDB GSI, SQS visibility/age/dedup/group, Redis TTL index, S3 route catalog) are excluded by default and rebuilt by replaying user data through public adapter endpoints on restore. HLC ceiling, Raft term/index, txn intents, in-flight multipart uploads are never dumped.
- **Filename encoding** is documented (RFC3986 unreserved + `%HH`, `b64.` prefix for binary DynamoDB keys, SHA-256 fallback for >240-byte segments) with a `KEYMAP` sidecar that round-trips originals.

## Phasing

1. Phase 1 — `cmd/elastickv-backup` producer + `BeginBackup` / `EndBackup` admin RPCs.
2. Phase 2 — `cmd/elastickv-restore` consumer driving public adapter endpoints (PutItem, PutObject, SET/HSET/…, SendMessage). Restore exercises the same code path real clients do, so format/version drift never silently bricks an old backup.
3. Phase 3 — incremental / scoped backups (out of scope here, sketched at the bottom).

## Self-review

- **Data loss** — Doc only; no code changes. The proposal explicitly leaves the FSM snapshot path untouched (see `2026_04_14_implemented_etcd_snapshot_disk_offload.md`) and registers `read_ts` with the active timestamp tracker so dumps cannot lose data to compaction.
- **Concurrency / distributed failures** — Cross-shard consistency uses the same lease-read primitive every adapter already uses; producer crash leaves no `MANIFEST.json`, so restore refuses partial trees.
- **Performance** — Dump is 1.5×–3× larger and slower than the physical FSM snapshot; trade-off is documented. No hot-path impact.
- **Data consistency** — Single `read_ts` across all adapters; GSIs/secondary indexes intentionally re-derived on restore so they cannot diverge from the base data they shadow.
- **Test coverage** — P0/P1/P2 test plan included in the doc (filename round-trip, multipart S3 reassembly, FIFO order, cross-adapter consistency, external-tool replay against MinIO / local DynamoDB / real Redis).

## Test plan

- [ ] No code changed — `make lint` / `make test` are no-ops for this PR.
- [ ] Reviewers: confirm the layout is decomposable enough to recover *one* record without elastickv, and that the internal-state exclusion list is complete (anything that would resurrect aborted state on restore must be excluded).

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive design specification for a logical backup system featuring admin controls, cluster-wide consistency, and automatic expiration management
  * Added specification for a backup file decoder tool enabling users to inspect and restore from logical backups

### DIFF
--- a/docs/design/2026_04_29_proposed_logical_backup.md
+++ b/docs/design/2026_04_29_proposed_logical_backup.md
@@ -312,9 +312,10 @@ item table emits 50 million inodes. On most modern filesystems
 slow for live filesystem operations.
 
 For tables where the inode count is the binding constraint, the
-producer accepts an opt-in `--dynamodb-bundle-mode jsonl[:size=64MiB]`
-that emits items as `items/data-<part-id>.jsonl` instead, packed up
-to a configurable per-file size budget:
+producer accepts an opt-in `--dynamodb-bundle-mode jsonl` (paired
+with `--dynamodb-bundle-size 64MiB`, defaulting to that value) that
+emits items as `items/data-<part-id>.jsonl` instead, packed up to the
+configurable per-file size budget:
 
 ```
 dynamodb/orders/
@@ -435,8 +436,10 @@ redis/
     │   └── tags%3Apost1.json
     ├── zsets/
     │   └── leaderboard.json
-    └── streams/
-        └── events.jsonl
+    ├── streams/
+    │   └── events.jsonl
+    └── hll/
+        └── pfcount%3Auniques.bin
 ```
 
 Redis values are encoded so that `redis-cli --pipe` (or the equivalent in
@@ -611,7 +614,7 @@ the proposal does add is a thin **iterator wrapper** so a multi-million
 key range does not need to be materialized in one call:
 
 ```go
-// kv/backup_scan.go (new)
+// kv/shard_store.go (new method on *ShardStore — same file as ScanAt)
 type BackupScanner interface {
     // Next advances the iterator. Returns (nil, false, nil) at end-of-range.
     // The returned KVPair is owned by the caller until the next call to Next.
@@ -620,11 +623,35 @@ type BackupScanner interface {
 }
 
 // NewBackupScanner returns a forward iterator over [start, end) at ts.
-// Internally calls ShardStore.ScanAt in pages of pageSize, chaining
+// Internally calls ScanAt in pages of pageSize, chaining
 // `start = lastReturnedKey + \x00` between pages so no lock is held
 // across pages and the compactor's MVCC retention budget is bounded.
-func (c *ShardedCoordinator) NewBackupScanner(start, end []byte, ts uint64, pageSize int) BackupScanner
+func (s *ShardStore) NewBackupScanner(start, end []byte, ts uint64, pageSize int) BackupScanner
 ```
+
+`NewBackupScanner` lives on `*ShardStore` (its natural home, sharing a
+file with `ScanAt`) rather than `*ShardedCoordinator`. The coordinator
+type today (`kv/sharded_coordinator.go:124`) has no `*ShardStore`
+field — only `engine`, `router`, `groups`, `clock`, and a raw
+`store.MVCCStore` — so placing the scanner on the coordinator would
+force either an awkward extra field plus `main.go` wiring or a
+roundabout coordinator → router → ShardStore lookup. Putting it on
+`*ShardStore` keeps the scan primitive cohesive with `ScanAt` and
+lets the admin server reach it directly via the existing
+`AdminServer` → `ShardStore` wiring used by `GetRaftGroups`.
+
+> **Phase 1 also updates `ScanAt`'s doc comment**
+> (`kv/shard_store.go:101–105`) which today warns _"this method is
+> NOT a globally consistent snapshot; the caller must implement a
+> cross-shard snapshot fence"_. Under the backup contract the caller
+> *is* fencing — `BeginBackup` waits for `applied_index ≥ f(ts)` on
+> every group before any `Next` call. The new comment must say
+> something like: _"consistent across groups when the caller has
+> fenced `applied_index ≥ f(ts)` cluster-wide before calling, as
+> `BeginBackup` does. Without that fence, results are eventually-
+> consistent across groups."_ Without this comment update, code
+> review of the Phase 1 producer will flag the `ScanAt` use as
+> contradicting the warning.
 
 `pageSize` is the same `ScanAt` `limit` parameter; the producer's
 default is 1024 and is exposed as a `--scan-page-size` CLI flag.
@@ -660,17 +687,29 @@ message BeginBackupResponse {
 // dump calls this every ttl_ms/3 (producer default) so a multi-hour
 // scan never relies on a single TTL window. The read_ts is preserved
 // across renewals; only the deadline shifts.
-message RenewBackupRequest  { bytes pin_token = 1; uint64 ttl_ms = 2; }
+message RenewBackupRequest {
+  bytes  pin_token = 1;
+  // Same range constraint as BeginBackupRequest.ttl_ms:
+  // 60s–24h, bounded above by backup_max_ttl_ms. Out-of-range values
+  // are rejected with InvalidArgument.
+  uint64 ttl_ms    = 2;
+}
 message RenewBackupResponse { uint64 ttl_ms_effective = 1; }
 
 message EndBackupRequest  { bytes pin_token = 1; }
 message EndBackupResponse {}
 
+// ListAdaptersAndScopes runs the per-adapter metadata-prefix scan at
+// the read_ts associated with pin_token, so the returned scope list
+// is a precise match for what BackupScanner will surface. A new scope
+// created by a concurrent client between BeginBackup and this call is
+// invisible at the pinned read_ts and therefore not listed.
+message ListAdaptersAndScopesRequest  { bytes pin_token = 1; }
 message ListAdaptersAndScopesResponse {
-  repeated string dynamodb_tables = 1;  // from !ddb|meta|table| scan
-  repeated string s3_buckets       = 2;  // from !s3|bucket|meta| scan
+  repeated string dynamodb_tables = 1;  // from !ddb|meta|table| scan at read_ts
+  repeated string s3_buckets       = 2;  // from !s3|bucket|meta| scan at read_ts
   repeated uint32 redis_databases  = 3;  // {0} until multi-db lands
-  repeated string sqs_queues       = 4;  // from !sqs|queue|meta| scan
+  repeated string sqs_queues       = 4;  // from !sqs|queue|meta| scan at read_ts
 }
 ```
 
@@ -833,7 +872,10 @@ elastickv-restore apply \
   [--adapter dynamodb,s3,redis,sqs] \
   [--scope    dynamodb=orders] \
   [--mode replace|merge|skip-existing] \
-  [--rate-limit 5000ops/s]
+  [--rate-limit 5000ops/s] \
+  [--preserve-ttl] \
+  [--preserve-visibility] \
+  [--stream-merge-strategy reject|auto-id]
 ```
 
 `replace` deletes the target scope before re-importing; `merge`
@@ -1031,6 +1073,8 @@ Scope: out of this proposal; mentioned only to draw the boundary.
 | `TestS3PathFileVsDirectoryCollision` | Bucket holds both `path/to` (object) and `path/to/obj`; producer renames the shorter key to `path/to.elastickv-leaf-data` and records it in `KEYMAP.jsonl`; restore tool reverses it via `MANIFEST.s3_collision_strategy` |
 | `TestBeginBackupTooManyActiveBackups` | Reaching `max_active_backup_pins` returns `ResourceExhausted`; releasing one pin frees a slot for the next request |
 | `TestRenewBackupExtendsDeadline` | `RenewBackup` shifts the deadline; producer's failed-renewal path aborts the dump with a critical log line rather than continuing past the TTL |
+| `TestRenewBackupTTLRangeValidation` | `RenewBackup` with `ttl_ms < 60s` or `ttl_ms > backup_max_ttl_ms` returns `InvalidArgument`; in-range values succeed |
+| `TestListAdaptersAndScopesAtPinTS` | A scope created (e.g. CreateTable) after `BeginBackup` is not surfaced by `ListAdaptersAndScopes(pin_token)`; pre-existing scopes are |
 
 ### P1
 
--- a/docs/design/2026_04_29_proposed_logical_backup.md
+++ b/docs/design/2026_04_29_proposed_logical_backup.md
@@ -121,7 +121,7 @@ or migrate the data **without elastickv being part of the recovery path**.
 ```
 backup-<utc-timestamp>-<cluster-id>-<commit-ts>/
 ├── MANIFEST.json
-├── CHECKSUMS                         # SHA-256 of every regular file under the root
+├── CHECKSUMS                         # sha256sum(1)-compatible: "<64-hex>  <relative-path>" per line
 ├── dynamodb/
 │   └── <table-name>/
 │       ├── _schema.json
@@ -130,20 +130,34 @@ backup-<utc-timestamp>-<cluster-id>-<commit-ts>/
 ├── s3/
 │   └── <bucket-name>/
 │       ├── _bucket.json
-│       └── <object-key-path>          # original object bytes, original hierarchy
+│       ├── <object-key-path>                       # original object bytes
+│       └── <object-key-path>.elastickv-meta.json   # sidecar (reserved suffix)
 ├── redis/
 │   └── db_<n>/
 │       ├── strings/<key>.bin
+│       ├── strings_ttl.json
 │       ├── hashes/<key>.json
 │       ├── lists/<key>.json
 │       ├── sets/<key>.json
 │       ├── zsets/<key>.json
-│       └── streams/<key>.jsonl
+│       ├── streams/<key>.jsonl
+│       └── hll/<key>.bin                           # HyperLogLog opaque sketch (!redis|hll|<key>)
 └── sqs/
     └── <queue-name>/
         ├── _queue.json
         └── messages.jsonl
 ```
+
+`CHECKSUMS` is exact `sha256sum(1)` output so verification works without
+elastickv: `sha256sum -c CHECKSUMS` from the dump root succeeds on a
+clean dump and fails (with file-level diagnostics) on tampering.
+
+The `.elastickv-meta.json` suffix is **reserved**: a user S3 object
+whose key happens to end in `.elastickv-meta.json` is rejected at dump
+time with a typed error rather than silently colliding with its own
+sidecar. Operators who hold such keys pass `--rename-collisions` to have
+them emitted as `<obj>.elastickv-meta.json.user-data` (the rename is
+recorded in `KEYMAP`).
 
 The directory **is** the index. There is no other side-table that must be
 parsed before the user can find a record.
@@ -190,15 +204,21 @@ convenience, not a correctness dependency.
 
 ```
 dynamodb/
-└── orders/
+└── orders/                                        # composite-key table (hash + range)
     ├── _schema.json
     └── items/
-        ├── customer-7421/
-        │   ├── 2026-04-29T12:00:00Z.json
+        ├── customer-7421/                          # <pk>/
+        │   ├── 2026-04-29T12:00:00Z.json           # <sk>.json
         │   └── 2026-04-29T13:15:42Z.json
         ├── customer-7422/
         │   └── 2026-04-29T09:00:00Z.json
-        └── b64.AAECAw../single.json     # binary partition key
+        └── b64.AAECAw../                           # binary partition key (B attribute)
+            └── 2026-04-29T10:00:00Z.json
+└── sessions/                                      # hash-only table
+    ├── _schema.json
+    └── items/
+        ├── sess-abc123.json                        # <pk>.json directly under items/
+        └── sess-def456.json
 ```
 
 `_schema.json`:
@@ -272,22 +292,26 @@ s3/
     │   └── 04/
     │       └── 29/
     │           ├── img.jpg
-    │           ├── img.jpg.metadata.json
+    │           ├── img.jpg.elastickv-meta.json
     │           ├── thumbnails/
-    │           │   └── img-128x128.jpg
-    │           └── thumbnails/img-128x128.jpg.metadata.json
+    │           │   ├── img-128x128.jpg
+    │           │   └── img-128x128.jpg.elastickv-meta.json
     └── archive/
-        └── manifest.csv
+        ├── manifest.csv
+        └── manifest.csv.elastickv-meta.json
 ```
 
 The S3 object body sits at its natural path — every byte that the
 elastickv S3 adapter would have streamed through `streamObjectChunks` for a
 GET is reassembled in order and written out as a single regular file.
 
-A sidecar `<object>.metadata.json` carries the parts of `s3ObjectManifest`
-that S3 itself exposes via headers (`Content-Type`, `Content-Encoding`,
-`Cache-Control`, `Content-Disposition`, user-defined `x-amz-meta-*`,
-`ETag`, `LastModified`):
+A sidecar `<object>.elastickv-meta.json` carries the parts of
+`s3ObjectManifest` that S3 itself exposes via headers (`Content-Type`,
+`Content-Encoding`, `Cache-Control`, `Content-Disposition`, user-defined
+`x-amz-meta-*`, `ETag`, `LastModified`). The suffix is reserved (see
+"Top-Level Layout" above): a user S3 object whose key ends in
+`.elastickv-meta.json` is rejected at dump time unless
+`--rename-collisions` is passed.
 
 ```json
 {
@@ -477,8 +501,11 @@ emits an `_internals/` subdirectory of newline-delimited records.
     "include_sqs_side_records":   false
   },
   "checksum_algorithm": "sha256",
+  "checksum_format":    "sha256sum",
   "encoded_filename_charset": "rfc3986-unreserved-plus-percent",
-  "key_segment_max_bytes": 240
+  "key_segment_max_bytes": 240,
+  "backup_ts_ttl_ms":   1800000,
+  "s3_meta_suffix":     ".elastickv-meta.json"
 }
 ```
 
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
Claude-bot review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340164148) reflected in 57191281. Summary of responses:

**Critical**
- #1 `ScanRange` doesn't exist — replaced with `ShardStore.ScanAt` (which does, `kv/shard_store.go:106`) plus a new `BackupScanner` iterator wrapper for paged scans.
- #2 `ActiveTimestampTracker` no TTL — added `PinWithDeadline` + per-second sweeper goroutine spec.

**Important**
- #3 `hll/` added to the layout tree.
- #4 sidecar collision — switched to reserved `.elastickv-meta.json` suffix, with `--rename-collisions` for the legitimately-named-that-way case.
- #5 cross-shard wait — `BeginBackup` polls each group's `applied_index` until it covers `read_ts` (5s default deadline); the 50ms skew alone was only a liveness assertion.
- #6 `CHECKSUMS` is sha256sum(1)-compatible (`sha256sum -c CHECKSUMS` from the dump root).
- #7 `ListAdaptersAndScopes` RPC sketched alongside BeginBackup/EndBackup, including `ttl_ms` / `pin_token`.

**Minor**
- #8 Redis TTL: default `skip-expired` on restore, `--preserve-ttl` opt-in.
- #9 `aws s3 sync` example now `--exclude '*.elastickv-meta.json'`.
- #10 DynamoDB example clarified — composite-key vs hash-only layouts shown side by side.
- #11 SQS dedup window moved into trade-offs as an explicit callout (with `--include-sqs-side-records` mention).
- #12 Redis stream merge: `merge` refuses non-empty target; `--stream-merge-strategy=auto-id` falls back to `XADD *` with `_id_remap.jsonl`.

**Nits**
- `backup_ts_ttl` / `s3_meta_suffix` / `checksum_format` added to `MANIFEST.json`; `ttl_ms` is on the `BeginBackup` RPC sketch.

New tests: `TestPinWithDeadlineExpiry`, `TestBeginBackupWaitsForLaggingShard`, `TestBackupScannerPaging`, `TestS3SidecarSuffixCollision`, `TestRedisTTLExpiredKeySkippedByDefault`, `TestRedisStreamMergeRejectsNonEmpty`.

@claude review

bootjp:
Gemini review (https://github.com/bootjp/elastickv/pull/706#pullrequestreview-3498977717) addressed in `bbd1bb8b`. Summary:

**HIGH (line 515) — TTL refresh for large backups**
- Added `RenewBackup` RPC; producer calls it every `ttl_ms/3` and aborts the dump on renewal failure (continuing past TTL would silently corrupt the artifact, since the compactor would already have retired versions the in-flight scan depends on).
- `backup_max_ttl_ms` (default 1h) caps a single window; multi-hour dumps work via repeated renewals.
- `pin_renewals_total` metric for operator visibility.

**MEDIUM (line 129) — DynamoDB per-item file scalability**
- Default stays per-item (`items/<pk>/<sk>.json`). This is the user-stated requirement that motivates the entire proposal — "DynamoDBならテーブルディレクトリ配下にPKのファイル名.json でレコードがある" — and what makes one-line restore scripts trivial.
- Added opt-in `--dynamodb-bundle-mode jsonl[:size=64MiB]` for tables where inode count is the binding constraint (millions of items). Structured warning emitted when an unbundled scope exceeds 1M items so operators know to switch.
- `MANIFEST.json` records `dynamodb_layout: "per-item"|"jsonl"` so restore dispatches the right reader.

**MEDIUM (line 163) — S3 file-vs-directory POSIX collision**
- Producer detects keys like `path/to` and `path/to/obj` coexisting in the same bucket; the shorter key is renamed to `path/to.elastickv-leaf-data` with the rename in `KEYMAP.jsonl` and `s3_collision_strategy: "leaf-data-suffix"` in `MANIFEST.json`. Sidecar follows the renamed base.

**MEDIUM (line 181) — KEYMAP scalability**
- `KEYMAP` → `KEYMAP.jsonl` (line-streamable, never loaded fully into memory).

**MEDIUM (line 341) — Redis TTL file scalability**
- `strings_ttl.json` → `strings_ttl.jsonl`, one record per line.

**MEDIUM (line 503) — DoS bound on concurrent backup pins**
- `max_active_backup_pins` (default 4) caps concurrent `BeginBackup` registrations; tracker returns `ErrTooManyActiveBackups`, RPC surfaces `ResourceExhausted`. Prevents a misbehaving caller from holding the compactor open across the MVCC retention horizon.

New CLI flags: `--ttl-ms`, `--scan-page-size`, `--dynamodb-bundle-mode`, `--dynamodb-bundle-size`, `--rename-collisions`.

New tests: `TestS3PathFileVsDirectoryCollision`, `TestBeginBackupTooManyActiveBackups`, `TestRenewBackupExtendsDeadline`.

@claude review

bootjp:
Round-3 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340209496) addressed in `e3b29360`. Summary:

**Carry-overs from round 2 (5 items)**
1. **`ScanAt` comment** — Phase-1 note added: the comment must be updated to reflect that `BeginBackup`'s `applied_index` fence makes cross-shard `ScanAt` globally consistent.
2. **`NewBackupScanner` placement** — moved from `*ShardedCoordinator` to `*ShardStore`. Verified: `ShardedCoordinator` (`kv/sharded_coordinator.go:124`) has no `*ShardStore` field, so the natural home is `*ShardStore` next to `ScanAt`.
3. **`ListAdaptersAndScopesRequest`** — `pin_token` field added; the scope scan is now bounded to the pinned `read_ts` so concurrently-created tables/buckets are not surfaced.
4. **`hll/`** — added to the Redis `db_0` per-adapter example tree (top-level layout already had it).
5. **`--stream-merge-strategy`** — added to the `elastickv-restore apply` CLI block alongside `--preserve-ttl` and `--preserve-visibility`.

**New from Gemini changes (2 items)**
6. **`--dynamodb-bundle-mode` syntax** — two-flag form (`--dynamodb-bundle-mode jsonl` + `--dynamodb-bundle-size 64MiB`) is now used throughout; the inline `jsonl:size=64MiB` form is gone.
7. **`RenewBackupRequest.ttl_ms`** — documents the same `60s–24h` constraint as `BeginBackupRequest`, bounded above by `backup_max_ttl_ms`; out-of-range returns `InvalidArgument`.

New tests: `TestRenewBackupTTLRangeValidation`, `TestListAdaptersAndScopesAtPinTS`.

@claude review

bootjp:
Round-4 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340233249) addressed in `797d5711`. Both data-loss gaps verified against `adapter/redis.go`.

**Important (data-loss gaps)**

1. **Stream TTLs silently dropped** — verified at `adapter/redis.go:2471` (`buildTTLElems`). Streams can have TTL via `PEXPIREAT`, kept in `!redis|ttl|`. Added `expire_at_ms` to the `_meta` line:
   ```
   {"_meta": true, "length": 2, "last_ms": 1714400000001, "last_seq": 0, "expire_at_ms": null}
   ```
2. **HLL TTLs silently dropped** — verified at `adapter/redis.go:2319-2320` comment. Added `hll_ttl.jsonl` sidecar at the `db_<n>` root with the same shape as `strings_ttl.jsonl`. Omitted when no HLL key has a TTL.
3. **`XADD NOMKSTREAM` semantic error** — replaced with the correct mechanism: attempt `XADD <id>` and treat `ERR The ID specified in XADD is equal or smaller` as "already present, skip." `NOMKSTREAM` does not skip-by-ID — it only suppresses stream creation. Documented O(N) cost.

**Minor**

4. **`--begin-backup-deadline 5s`** added to the `elastickv-backup dump` CLI block (the prose mentioned it but the flag list lagged).

New tests: `TestRedisStreamTTLRoundTrip`, `TestRedisHLLTTLRoundTrip`, `TestRedisStreamSkipExistingHandlesERR`.

@claude review

bootjp:
Round-5 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340258363) addressed in `2c00d560`.

**Important (correctness gap)**

1. **`BeginBackup` pin was node-local** — verified at `main.go:294,335`. Each node owns its own `ActiveTimestampTracker` and only its compactor consults it; a pin on the receiving node does not protect group leaders on other nodes from compacting past `read_ts`.

   **Fix**: pins propagate through each Raft group's log as `BackupPin{pin_id, read_ts, deadline}` / `BackupExtend` / `BackupRelease` FSM commands. Every replica applies them to its local tracker, so the pin set is replicated and durable across leader changes (a new leader inherits the pin from the same log entry it applied — no separate handoff). `BeginBackup` fans the propose out per group and waits for commit; if any group fails to commit within `--begin-backup-deadline`, `BackupRelease` is proposed on groups that did commit and `Unavailable` is returned so retry is clean. `RenewBackup` / `EndBackup` fan out the same way; `BackupRelease` is idempotent so partial state is tolerated.

**Minor**

2. **`_id_remap.jsonl` path** — specified as `redis/db_<n>/streams/<key>._id_remap.jsonl` (next to the source) so tools can find it on disk.
3. **Test description** — `TestRedisStreamSkipExistingHandlesERR`: "the dump" → "the restore".
4. **`KEYMAP.jsonl`** — added to every per-adapter scope in the layout tree (omitted when empty).

New tests: `TestBeginBackupPinFanOutAllNodes`, `TestBeginBackupPinSurvivesLeaderChange`, `TestBeginBackupGroupUnreachable`.

@claude review

bootjp:
Round-6 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340289225) addressed in `020e0c4c`. Both important items verified against `kv/fsm.go`.

**Important**

1. **FSM plumbing missing from Phase 1 scope** — added the four concrete items the implementor needs:
   - New byte tags `0x03/0x04/0x05` in `kv/fsm.go` alongside `raftEncodeHLCLease = 0x02`.
   - New `applyBackupPin / applyBackupExtend / applyBackupRelease` handlers, dispatched from `kvFSM.Apply` in the same shape as `applyHLCLease`.
   - New `*ActiveTimestampTracker` field on `kvFSM` wired via `NewKvFSMWithHLCAndTracker(store, hlc, tracker)` — analogous to the `*HLC` sharing invariant in `CLAUDE.md`. Coordinator and FSM share one tracker.
   - New `kv/backup_codec.go` with fixed 33 / 25 / 17-byte binary layouts for the three entry types (matching the HLC-lease style, not proto, to keep the apply path light).

2. **`kvFSM.Restore` wipes in-memory pins** — added an explicit safe-bound invariant:

   > `backup_duration + worst_case_compaction_lag < mvcc_retention_horizon`

   `BeginBackup` enforces a soft form via `--snapshot-headroom-entries` (default 10000): if any group's `appliedIndex - firstIndex` is below this margin, `BeginBackup` returns `FailedPrecondition` rather than starting a dump that risks the snapshot-installation path. The producer's per-scope `expected_keys` heuristic catches mid-dump snapshot-induced shortfalls. Heavier mitigations (snapshotting tracker state, sibling-pin queries on `Restore`) are explicitly **Phase 3 only** — Phase 1's contract is "your dump completes within the retention horizon."

**Minor**

3. **`--begin-backup-deadline` binding-constraint** clarified: the `applied_index` poll for lagging shards is the binding wait; `BackupPin` commit is sub-second on a healthy cluster. Operators tuning the flag are adjusting tolerance for shard lag.
4. **`RenewBackup` retries** `BackupExtend` per-group up to 3 times with 500ms backoff before aborting. Transient leader elections (<1s under etcd/raft defaults) no longer kill otherwise-healthy dumps.

New tests: `TestBackupPinFSMCodecRoundTrip`, `TestRestoreWipesLocalPins`, `TestBeginBackupRefusesNearSnapshotThreshold`, `TestRenewBackupRetriesLeaderElection`.

@claude review

bootjp:
Round-7 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340315378) addressed in `b1669803`. Both important items are bugs caught by reading `internal/raftengine/engine.go` directly.

**Important**

1. **`--snapshot-headroom-entries` check was inverted**. Real formula:
   ```
   entries_since_last_snapshot = AppliedIndex - LastSnapshotIndex
   remaining_headroom          = SnapshotEvery - entries_since_last_snapshot
   refuse if: remaining_headroom < --snapshot-headroom-entries
   ```
   `Status` has `LastSnapshotIndex` (not `FirstIndex`), and `defaultSnapshotEvery = 10000` (verified at `internal/raftengine/etcd/engine.go:92`). A freshly-snapshotted cluster has the *largest* remaining headroom and is now correctly allowed; the check refuses backups that would trigger the snapshot-installation corner case mid-dump. Default lowered from 10000 to 1000 (one-tenth of `SnapshotEvery`).
2. **`expected_keys` heuristic was undefined**. Replaced with a concrete **expected-keys baseline**:
   - `BeginBackup` runs a count-only `ScanAt` at `read_ts` *before* the pin is installed, returns per-scope `(key_count, applied_index)` in `BeginBackupResponse.expected_keys`.
   - Producer compares actual count against baseline; fails with `ErrCompactionDuringDump` if below `99% × baseline ± sqrt(baseline)` (binomial-noise tolerance for routine TTL expiry).
   - `MVCCStore.ScanAt` / `ShardStore.ScanAt` extended with a `keys_only bool` parameter so the baseline pass skips value materialization (Pebble iterator with payload trimmed at the block layer; ~10× cheaper).

**Minor**

3. `_long_keys.jsonl` → `KEYMAP.jsonl` on line 192 (consistent with the rest of the spec).
4. `firstIndex` → `LastSnapshotIndex` (item #1 already covers it).
5. `BackupPin` fan-out is now explicitly **concurrent** (one goroutine per group). A 100-shard cluster does not pay 100× serial Raft round-trips at the start of every backup.

New tests: `TestExpectedKeysBaselineToleratesTTLExpiry`, `TestExpectedKeysBaselineCountOnlyScan`. `TestRestoreWipesLocalPins` and `TestBeginBackupRefusesNearSnapshotThreshold` updated to match the corrected formulas.

@claude review

bootjp:
Round-8 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340356184) addressed in `7ca1ae51`.

**Important**

1. **Proto definitions filled in.** Added `ShardApplied {group_id, applied_index}` and `ScopeKeyCount {scope_id, key_count, applied_index_at_count}`; added `repeated ScopeKeyCount expected_keys = 5` to `BeginBackupResponse`.

2. **Expected-keys baseline ordering bug fixed.** It said "during step 1" but step 1 is "pick `read_ts`" and step 2 is shard catch-up. A lagging shard would return a partial count for `ScanAt` at `read_ts` before applying through it, producing an undercount and silently disabling `ErrCompactionDuringDump` on exactly the shards most at risk of snapshot-induced pin loss. Moved to "**after step 2 (every shard has applied through `read_ts`) and before step 3 (pin)**".

**Minor**

3. `--snapshot-headroom-entries` CLI default `10000` → `1000` (matches the one-tenth-of-`SnapshotEvery` default documented in the prose).

4. `ScanAt` `keys_only` parameter would be a source-incompatible Go change touching ~10 callers in `shard_store.go` alone. Replaced with a **new `ScanKeysAt` overload** on `MVCCStore` / `ShardStore` (returns `[][]byte` rather than `[]*KVPair`). Existing `ScanAt` callers untouched; the new path delegates to a private helper that shares route resolution and merge logic.

Tests: `TestExpectedKeysBaselineCountOnlyScan` now verifies both the new path and absence of regression on existing `ScanAt` callers; new `TestExpectedKeysBaselineRunsAfterShardCatchup` guards against the ordering bug.

@claude review

bootjp:
Round-9 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340384011) addressed in `5bda8a8b`. Both important items verified against `internal/raftengine/{etcd/,}engine.go` and `kv/fsm.go`.

**Important**

1. **`SnapshotEvery` not on `raftengine.Engine` interface, and the wrong env var name in the doc.** Phase 1 adds `StatusReader.SnapshotEvery() uint64` implemented on the etcd backend by returning the value cached at `Open` from `snapshotEveryFromEnv()`. Also corrected: there is no `--raftSnapshotEvery` flag; the override is `ELASTICKV_RAFT_SNAPSHOT_COUNT` (`internal/raftengine/etcd/engine.go:93`).

2. **Rolling-upgrade safety.** Confirmed: an unknown leading tag falls through `decodeRaftRequests` (`kv/fsm.go:140`) → `decodeLegacyRaftRequest` (`kv/fsm.go:145`) → `proto.Unmarshal` → `Apply` error → apply-loop stall on un-upgraded replicas. A single-release introduction of `BackupPin` would corrupt apply on every un-upgraded replica the moment the first `BeginBackup` runs.

   **Two-release rollout** specified in Phase 1 scope:
   - **Release N**: `kvFSM.Apply` treats unknown tags in `[0x03, 0x0F]` as no-ops with an `unknown_fsm_tag` warning (forward-compat handler only — no other backup machinery active). Reserved range bounded so future tags (CDC, etc.) inherit the path without reopening the question.
   - **Release N+1**: `BackupPin` / `Extend` / `Release` handlers ship; `BeginBackup` gates on every group member reporting `version ≥ N` via `Admin.GetRaftGroups`, returning `FailedPrecondition` otherwise.

   Operators may collapse N and N+1 into one calendar release if they enforce a fleet-wide upgrade window before any backup runs; the gate check still short-circuits.

**Minor**

3. **`ScopeKeyCount.scope_id`** format now documented for all four adapters, including the `redis_databases` `uint32` → `"redis/db_<n>"` mapping for Redis.

New tests: `TestForwardCompatUnknownFSMTag`, `TestBeginBackupGatesOnMinVersion`, `TestSnapshotEveryReadsFromEngine`.

@claude review

bootjp:
Round-10 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340416675) addressed in `9bf18e1c`. The important item was a real bug — verified against `proto/admin.proto:38-49`.

**Important**

1. **`RaftGroupState` does not carry `node_version`.** My round-9 claim that `Admin.GetRaftGroups` "already carries per-node version metadata" was wrong — `RaftGroupState` has only `raft_group_id`, `leader_node_id`, `leader_term`, `commit_index`, `applied_index`, `last_contact_unix_ms`. Phase 1 therefore explicitly **adds `string node_version = 7;` to `RaftGroupState`**, populated by each admin handler from the build-time version constant. Without this, the rolling-upgrade gate is unimplementable. Removed the false "already carries" claim.

**Minor**

2. **`MaxEntryBytes` doesn't exist in the codebase.** Replaced with the real limit: `MaxSizePerMsg` (`defaultMaxSizePerMsg = 1<<20`, `internal/raftengine/etcd/engine.go:55`).

3. **Forward-compat handler position specified.** In `kvFSM.Apply` immediately after the `raftEncodeHLCLease` guard and **before** `decodeRaftRequests`, so all leading-byte dispatch lives in one place. Tags `≥ 0x10` still fall through and error, bounding the forward-compat door.

`TestBeginBackupGatesOnMinVersion` updated to assert against the new `RaftGroupState.node_version` field.

@claude review

bootjp:
Round-11 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340445638) addressed in `a0292c95`. Both important items verified against `adapter/admin_grpc.go`.

**Important**

1. **Rolling-upgrade gate redesigned.** `GetRaftGroups` (`adapter/admin_grpc.go:382-416`) iterates the local `AdminServer`'s groups and reports each row from the *local* engine's `Status()`. Adding `node_version` to `RaftGroupState` would only ever surface the admin server's own version — followers on other nodes would not be reflected, and the gate would falsely pass on a partially-upgraded cluster.

   Phase 1 instead adds a dedicated `GetNodeVersion` RPC fanned out by the `BeginBackup` handler to every peer in `AdminServer.members`. Each peer returns its compiled-in version constant; the handler refuses `BeginBackup` if any peer is below N or unreachable within `--begin-backup-deadline`. `RaftGroupState` still gains `leader_node_version = 7` for UI visibility, but it is **not used for the gate decision**.

2. **`AdminServer` has no `*kv.ShardStore` field** (confirmed at `adapter/admin_grpc.go:62-82`). The "existing `AdminServer` → `ShardStore` wiring used by `GetRaftGroups`" claim was wrong — `GetRaftGroups` only reaches `AdminGroup.Status()` and `Configuration()`. Phase 1 adds the field explicitly and wires it in `main.go`.

**Minor**

3. **`AdminGroup` interface** (`adapter/admin_grpc.go:41-44`) does not embed `StatusReader`, so the `SnapshotEvery()` addition on `StatusReader` does not reach `BeginBackup` through `s.groups[id]`. Phase 1 extends `AdminGroup` itself with `SnapshotEvery() uint64`; mocks in `adapter/admin_grpc_test.go` gain one extra method that returns `defaultSnapshotEvery` for parity with production.

`TestBeginBackupGatesOnMinVersion` updated to match the new `GetNodeVersion` fan-out path.

@claude review

bootjp:
Round-12 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340484069) addressed in `3efd68d2`. Important item verified against `adapter/admin_grpc.go`.

**Important**

1. **`AdminServer.members` is a static bootstrap seed** (`adapter/admin_grpc.go:64,89` — set once in `NewAdminServer`, never updated). A node added later via `AddVoter` would not appear, silently missing exactly the nodes most likely to be at an older version during a rolling upgrade.

   **Fix**: `BeginBackup` reuses `snapshotMembers` (`adapter/admin_grpc.go:158`) — the same path `GetClusterOverview` uses — which unions the seed with members discovered from each group's `Configuration()`. Fan-out target is the live cluster.

**Minor**

2. **`GetRaftGroups` async-dial latency contract specified.** Adding `leader_node_version` would otherwise turn a synchronous local read into a fan-out handler with unbounded latency. Per-leader dial is async with 500ms timeout (failures yield `""`), and a 10s response cache prevents repeated fan-outs across rapid polls.

3. **`GetNodeVersion` failure-mode disambiguation.** Error message distinguishes "old version" (`FailedPrecondition: node X reports version v0.9.0; minimum is v1.0.0`) from "unreachable" (`FailedPrecondition: node X did not respond within 5s`) so operators can diagnose without out-of-band investigation. `--begin-backup-deadline` is **per-peer** (fan-out is concurrent), so worst-case wait is one deadline period regardless of cluster size.

New tests: `TestBeginBackupGatesIncludesDynamicallyAddedNodes`, `TestGetRaftGroupsLeaderVersionAsync`. `TestBeginBackupGatesOnMinVersion` updated to assert the new error shapes and the snapshotMembers-based fan-out.

@claude review

bootjp:
Round-13 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340510611) addressed in `6e62c7e7`. Important item verified against `adapter/admin_grpc.go` and `kv/grpc_conn_cache.go`.

**Important**

1. **`AdminServer` has no outbound gRPC connection mechanism today.** `GetNodeVersion` fan-out (in `BeginBackup`) and async leader probes (in `GetRaftGroups`) both need to dial peer admin servers. Without specifying the dialer, two implementors would make incompatible credential / lifecycle choices.

   **Fix**: Phase 1 adds a `*kv.GRPCConnCache` field to `AdminServer`, dialed with `internal.GRPCDialOptions()` — matching how `ShardStore` and the etcd-raft gRPC transport (`internal/raftengine/etcd/grpc_transport.go:469`) obtain peer connections today. Kept **separate from** `ShardStore.connCache` because admin and data-plane addresses occupy different address spaces (admin cache keyed on `NodeIdentity.GRPCAddress`; data-plane cache keyed on Raft-group-leader addresses that migrate on leader change).

**Minor**

2. **10s `leader_node_version` cache** now an explicit `AdminServer` field (`versionCache map[string]struct{ version string; fetchedAt time.Time }`) listed in Phase 1 scope rather than implicit prose.

New test: `TestAdminGRPCConnCacheReuse`.

@claude review

bootjp:
Round-14 review (https://github.com/bootjp/elastickv/pull/706#issuecomment-4340540845) addressed in `486675d4`. Important item verified against `adapter/admin_grpc.go:455-505`.

**Important**

1. **`GetNodeVersion` is itself an `Admin` method** and would be auth-gated on every cluster running `--adminToken`. The fan-out through `*kv.GRPCConnCache` makes outbound calls with no metadata, so every dial would return `Unauthenticated` and the gate would misreport "node did not respond" on every auth-enabled cluster.

   **Fix**: `BeginBackup` propagates the inbound bearer token to outbound `GetNodeVersion` calls via `metadata.NewOutgoingContext(context.Background(), inMD)`. Same pattern for async `GetNodeVersion(leader)` probes in `GetRaftGroups`. Forwarding the inbound caller's token (not a server-stored value) preserves the operator's authorization decision — no privilege escalation by the admin server.

**Minor**

2. **`versionCache` committed to `sync.Map`** with key/value types spelled out, removing the "(or sync.Map equivalent)" ambiguity that left room for a plain map + `groupsMu` approach that would either contend or race.

New tests: `TestBeginBackupPropagatesAdminAuthToken`, `TestVersionCacheRaceUnderLoad` (under `-race`).

@claude review

bootjp:
### Proposal split into Phase 0 + Phase 1

After review feedback exposed how much complexity the live-cluster PIT machinery had accumulated (replicated `BackupPin` FSM commands, `GetNodeVersion` fan-out, version gate, expected-keys baseline, admin-side connection cache, auth-token propagation), the design has been split so the **vendor-independent recovery** goal — the actual user-facing requirement — can ship without dragging the entire live-PIT pipeline along.

**Phase 0 — `docs/design/2026_04_29_proposed_snapshot_logical_decoder.md` (new)**
- Offline `.fsm` ↔ logical-format directory tree converter.
- `cmd/elastickv-snapshot-decode` / `cmd/elastickv-snapshot-encode`.
- **Owns the format definition** (per-adapter layout, filename encoding, `MANIFEST.json`, `KEYMAP.jsonl`, internal-state policy).
- No admin RPCs, no FSM tag, no Raft involvement, no proto changes.
- Restore via stop-replace-restart of the `.fsm` + `.snap` files.
- Sufficient for single-shard clusters, one-time exports off elastickv, and any case where the latest available snapshot is a good-enough recovery point.

**Phase 1 — `docs/design/2026_04_29_proposed_logical_backup.md` (this PR's existing doc, retitled)**
- Title updated to "Live Cluster, PIT-Consistent Extraction (Phase 1)".
- New Phasing section near the top cross-references Phase 0 as the format owner.
- Body retains the live-cluster machinery: `read_ts` pinning, cross-shard catch-up, expected-keys baseline, version-gated rolling-upgrade safety, FSM-level pin propagation.
- `MANIFEST.phase` field discriminates dumps from the two phases (`"phase0-snapshot-decode"` vs. `"phase1-live-pinned"`).
- Required only when cross-shard PIT consistency or "snapshot now" cadence is needed.

Both phases produce the **same on-disk format**, so external recovery tools (`aws s3 sync`, `aws dynamodb put-item`, `redis-cli --pipe`, …) do not care which phase produced a given dump.

Commit: `4ee0f3fa`

bootjp:
@claude review

bootjp:
CodeRabbit review (https://github.com/bootjp/elastickv/pull/706#pullrequestreview-3499053884) addressed in `605598fc`.

**Major (2)**

1. **Filename reversibility claim was overstated.** Reworded both Phase 0 and Phase 1 docs:
   - `%HH` percent-encoded segments and `b64.<base64url>` segments → filename alone is reversible; `KEYMAP.jsonl` is **convenience only**.
   - **SHA-fallback segments** (`<sha256-prefix-32>__<truncated-original>` for keys >240 bytes) and **S3 path-collision renames** (`<obj>.elastickv-leaf-data`) → `KEYMAP.jsonl` is a **correctness dependency**; exact-byte recovery is impossible without it.

**Minor (1)**

2. **Internal-state markdown table** had unescaped pipes inside inline code spans (`` `!s3route|` ``, `` `!redis|ttl|` ``, `` `!txn|` ``) which broke the three-column rendering (MD056). Pipes now escaped as `` `!s3route\|` `` etc.

**Nit (2)**

3. **Untyped fenced code blocks** — added `text` language identifier to all unlabeled fences (12 in Phase 1, 7 in Phase 0) so MD040 passes.

@claude review

PR #705, opened 2026-04-29
docs(design): propose per-adapter logical backup format

bootjp:
Per-adapter directory hierarchy (dynamodb/<table>/items/<pk>.json, s3/<bucket>/<original/path>, redis/db_<n>/<type>/<key>.{bin,json,jsonl}, sqs/<queue>/messages.jsonl) so a backup is recoverable without elastickv — cat / jq / find / aws s3 sync / redis-cli are sufficient. Strips internal magic prefixes and wide-column key encodings; pins read_ts via the lease- read path and the active timestamp tracker so the dump is consistent and the compactor cannot retire its versions mid-dump.

## Documentation
* Added design documentation for Phase 0 snapshot converter, specifying offline conversion between snapshot files and vendor-independent logical format with encoding rules and restore procedures.
* Added design documentation for Phase 1 logical backup system, outlining cluster-wide point-in-time consistent backup lifecycle management and forward-compatibility approach.

### DIFF
--- a/docs/design/2026_04_29_proposed_logical_backup.md
+++ b/docs/design/2026_04_29_proposed_logical_backup.md
@@ -312,9 +312,10 @@ item table emits 50 million inodes. On most modern filesystems
 slow for live filesystem operations.
 
 For tables where the inode count is the binding constraint, the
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
@claude review

PR #709, opened 2026-04-29
feat(keyviz): sampler GroupID + LeaderTerm stamping (Phase 2-C+ PR-3a)

bootjp:
## Summary

- Add `GroupID` field to `routeSlot` and `RaftGroupID + LeaderTerm` to `MatrixRow` (Phase 2-C+ §9.1 dedupe key)
- Add `MemSampler.SetLeaderTerm(groupID, term)` API and `groupTerms` snapshot at Flush time
- Extend `RegisterRoute` with a `groupID uint64` parameter; update all call sites (`main.go` reads `r.GroupID` from `distribution.Engine.Stats()`)

## Why

Per `docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md` §9.1, the cluster fan-out aggregator dedupes write samples by `(routeID, raftGroupID, leaderTerm, columnAt)` instead of the conservative max-merge. Max-merge undercounts during a leadership flip when the new leader observes a window the old leader was halfway through. Carrying `RaftGroupID + LeaderTerm` on every row gives the merge enough information to dedupe per-term and sum across terms.

## Scope (PR-3a — sampler API only)

This PR ships the sampler-side API. Wiring lands in follow-up PRs:

- **PR-3a (this)**: `routeSlot.GroupID`, `MatrixRow.RaftGroupID + LeaderTerm`, `SetLeaderTerm` API, `RegisterRoute` signature, `Flush` term snapshot.
- **PR-3b (follow-up)**: periodic ticker in `main.go` that polls engine Status and calls `SetLeaderTerm`; proto + JSON wire-format additions.
- **PR-3c (follow-up)**: aggregator-side `(group, term)`-keyed dedupe in `internal/admin/keyviz_fanout.go`.

With `SetLeaderTerm` never called (this PR alone), every row emits `LeaderTerm=0` and the fan-out merge falls back to today's max-merge → no behavior change for legacy deployments.

## Test plan

- [x] `go test -race -count=1 ./keyviz/...` — passes
- [x] `go test -race -count=1 ./internal/admin/...` — passes (uses MemSampler)
- [x] `go test -race -count=1 .` (root, `main_keyviz_test.go`) — passes
- [x] `golangci-lint run` — 0 issues
- [ ] Jepsen — N/A (sampler-internal change, no replication/MVCC impact)

## Five-lens self-review

1. **Data loss** — no on-disk format change; legacy max-merge fallback preserves today's behavior when `SetLeaderTerm` is never called.
2. **Concurrency** — `groupTermsMu` is a fine-grained `RWMutex`; `Observe` never touches it; `snapshotGroupTerms` clones at the top of `Flush` so every row in a column observes a stable view, even if `SetLeaderTerm` fires concurrently.
3. **Performance** — `Observe` hot path is unchanged (no new map lookup, no new lock); `Flush` gains one `RLock` + map clone per column, bounded by `len(groupTerms)` (typically ≤ a few groups).
4. **Data consistency** — `(routeID, raftGroupID, leaderTerm)` is a strict superset of today's `routeID` dedupe key; rows with `LeaderTerm=0` collapse to the legacy max-merge.
5. **Test coverage** — new `SetLeaderTerm` publishing + `Flush` snapshot tests under `keyviz/sampler_test.go`; existing `RegisterRoute` test signatures updated; race detector clean.

### DIFF
--- a/keyviz/flusher_test.go
+++ b/keyviz/flusher_test.go
@@ -9,7 +9,7 @@ import (
 func TestRunFlusherTicksUntilCancel(t *testing.T) {
 	t.Parallel()
 	s := NewMemSampler(MemSamplerOptions{Step: 5 * time.Millisecond, HistoryColumns: 16})
-	if !s.RegisterRoute(1, []byte("a"), []byte("b")) {
+	if !s.RegisterRoute(1, []byte("a"), []byte("b"), 0) {
 		t.Fatal("Register failed")
 	}
 	ctx, cancel := context.WithCancel(context.Background())
--- a/keyviz/sampler.go
+++ b/keyviz/sampler.go
@@ -160,6 +160,13 @@ type MemSampler struct {
 	// twice in the same column — once as an aggregate row, once as
 	// an individual row — even under register/remove churn.
 	virtualIDCounter atomic.Uint64
+
+	// groupTermsMu guards groupTerms. SetLeaderTerm and Flush both
+	// touch the map; Observe never does. The lock is fine-grained
+	// enough that contention is bounded by leader-term flips, which
+	// happen at most a few times per second across the whole cluster.
+	groupTermsMu sync.RWMutex
+	groupTerms   map[uint64]uint64
 }
 
 // retiredSlot tracks a removed slot through its post-removal grace
@@ -237,6 +244,14 @@ type routeTable struct {
 type routeSlot struct {
 	metaMu  sync.RWMutex
 	RouteID uint64
+	// GroupID is the Raft group this route belongs to. Phase 2-C+
+	// stamps this on every emitted MatrixRow so the cluster fan-out
+	// aggregator can dedupe write samples by (raftGroupID,
+	// leaderTerm) instead of the conservative max-merge that may
+	// undercount during a leadership flip. 0 means "no group
+	// attached" (legacy single-group deployments and the synthetic
+	// virtual-bucket slots both use 0).
+	GroupID uint64
 	Start   []byte
 	End     []byte
 	// Aggregate marks virtual buckets that fold multiple coarsened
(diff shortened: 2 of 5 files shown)


bootjp:
@claude review

bootjp:
@claude review

Round-1 follow-ups (`eab91c9a`) addressing 4 items from your initial review:

- **Medium / `SetLeaderTerm(0, nonZeroTerm)` silently stamps virtual-bucket rows** — added a `groupID == 0` early-return guard in `SetLeaderTerm` plus an explicit doc note that "groupID == 0 is reserved for virtual aggregate buckets". Reproducer test `TestSetLeaderTermZeroGroupIDDoesNotPolluteVirtualBucket` fails on the previous code and passes on the fix.
- **Medium / no test pins virtual-bucket `RaftGroupID=0`** — added `TestVirtualBucketRaftGroupIDIsZero` that registers two routes (`MaxTrackedRoutes=1`, both groupID=42), forces the second to fold into a virtual bucket, and asserts the aggregate row's `RaftGroupID == 0` (member's groupID=42 must NOT propagate).
- **Minor / live-slot idempotency undocumented by tests** — added `TestRegisterRouteSecondCallIgnoresGroupID` that calls `RegisterRoute(1, ..., 10)` then `RegisterRoute(1, ..., 99)` and asserts the emitted `RaftGroupID == 10`. Locks the documented "second call with different groupID is silently ignored" contract.
- **Nit / `snapshotGroupTerms` allocates on every Flush before PR-3b wires `SetLeaderTerm`** — added an `if len(s.groupTerms) == 0 { return nil }` early return; `nil[k]` returns zero value so the row-builder hot path needs no nil guard. Saves one allocation per `Flush` for current callers.

Cleanup: removed a stale comment leftover ("// TestNonPositiveOptionsFallBackToDefaults pins Codex round-8 P2: a") that was orphaned when `TestRegisterRouteCarriesGroupID` was inserted in the original PR.

`go test -race -count=1 ./keyviz/... ./internal/admin/... .` — pass; `golangci-lint run` — 0 issues.

PR #711, opened 2026-04-29
backup: filename encoding (Phase 0a foundation)

bootjp:
## Summary

First piece of the **Phase 0a logical-backup decoder** described in `docs/design/2026_04_29_proposed_snapshot_logical_decoder.md`. Adds `internal/backup/filename.go` + tests — the filename encoding/decoding primitive every per-adapter encoder will depend on.

Subsequent PRs will add: `KEYMAP.jsonl` writer, per-adapter encoders (DynamoDB / S3 / Redis / SQS), the main decoder pipeline, and the `cmd/elastickv-snapshot-decode` CLI.

## What this PR does

- **Encode**: bytes in `[A-Za-z0-9._-]` pass through; every other byte becomes `%HH`. Long segments (>240 bytes after expansion) take a SHA-256-prefix + truncated-original fallback.
- **Decode**: reverses percent and `b64.` segments; SHA-fallback inputs return `ErrShaFallbackNeedsKeymap` so callers cannot fabricate the original bytes from the filename alone.
- **Binary path**: DynamoDB B-attribute keys take a separate `b64.<base64url>` form so binary keys never collide with hex-shaped string keys.

## Test plan

- [x] `go test -race ./internal/backup/...` — all tests pass.
- [x] Property tests via `pgregory.net/rapid` covering round-trip on both encoding paths and SHA-fallback post-condition.
- [x] Negative tests: truncated `%HH`, non-hex digits, raw reserved bytes, malformed `b64.` segments.

## Self-review

- **Data loss** — N/A; pure encoding/decoding library. SHA-fallback explicitly returns a typed error rather than synthesizing a wrong key.
- **Concurrency** — All functions are pure; no shared state. `-race` clean.
- **Performance** — Single-pass encoders; pre-grown builder; no allocations beyond the output. SHA-256 only on the fallback path.
- **Data consistency** — Encoder is deterministic given the same input. Decode is the inverse on non-fallback inputs (verified by rapid). SHA-fallback is documented as requiring `KEYMAP.jsonl` for full reverse.
- **Test coverage** — Table-driven cases for the documented rules + rapid property tests for round-trip + targeted negative tests. New non-test code lines added: 222.

### DIFF
--- a/internal/backup/filename.go
+++ b/internal/backup/filename.go
@@ -80,30 +80,68 @@ var ErrShaFallbackNeedsKeymap = errors.New("backup: filename uses SHA fallback; 
 // filename component. It is the inverse of DecodeSegment for non-fallback
 // inputs.
 //
-// The encoding is deterministic and idempotent given the same input.
+// The encoding is deterministic given the same input.
 //
-// Two short-circuits ensure the encoder never trips its own invariants:
+// Three structural short-circuits ensure DecodeSegment cannot
+// misclassify a legitimate key:
 //
-//   - If raw is so large that percent-encoding it would always overflow
-//     maxSegmentBytes (3*len(raw) > maxSegmentBytes), we go straight to
-//     shaFallback without allocating the full expansion. Without this an
-//     adversarial caller could force a very large transient allocation
-//     just to discard it.
-//   - If the percent-encoded form happens to match the SHA-fallback shape
-//     (32 hex chars followed by "__"), we promote it to a real
+//   - If `raw` is longer than maxSegmentBytes, even a fully-unreserved
+//     encoding (1:1) cannot fit, so we go straight to shaFallback.
+//     This also caps the percent-encode allocation at
+//     ~maxSegmentBytes, preventing OOM on adversarial input.
+//   - If the percent-encoded form happens to match the SHA-fallback
+//     shape (32 hex chars followed by "__"), we promote it to a real
 //     SHA-fallback so DecodeSegment's structural detection cannot
-//     misclassify a legitimate key. Both isShaFallback and shaFallback
-//     are true on the resulting output, so KEYMAP.jsonl carries the
-//     original bytes for exact-byte recovery.
+//     fabricate a wrong original.
+//   - If the percent-encoded form starts with the binary "b64."
+//     prefix, we promote to SHA-fallback for the same reason: a
+//     plain string key like "b64.foo" would otherwise be decoded as
+//     base64 and produce different bytes on round-trip.
+//
+// Both promoted-fallback paths leave the original in KEYMAP.jsonl
+// (a correctness dependency, per the package doc), so exact-byte
+// recovery is preserved.
 func EncodeSegment(raw []byte) string {
-	if len(raw)*percentEncodeMaxExpansion > maxSegmentBytes {
+	if len(raw) > maxSegmentBytes {
+		// 1:1 lower bound on encoded length; cannot fit.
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
Review feedback addressed in `e7a84eba`:

**HIGH (Gemini line 90, Codex P1 line 207) — SHA-fallback collision**
Confirmed bug. A user key like `0123456789abcdef0123456789abcdef__suffix` (35 unreserved bytes) passed through `EncodeSegment` unchanged, then `DecodeSegment`'s structural detector misclassified it as a SHA fallback and returned `ErrShaFallbackNeedsKeymap`. Round-trip broke for legitimate keys that should not have required `KEYMAP.jsonl`. `EncodeSegment` now promotes such inputs to a real SHA fallback so `KEYMAP.jsonl` carries the original bytes.

**HIGH (Gemini lines 90, 104) — huge-input transient allocation**
`percentEncode` pre-grew its builder to `3 * len(raw)`; a 1 GiB raw input would allocate 3 GiB just to be discarded by the length check. Encoders now short-circuit to `shaFallback` when the worst-case expansion would always overflow the ceiling — `3*len` for percent, `4/3*len + 4` for `b64.<base64url>`. No allocation of the full expansion happens for adversarially large inputs.

**MEDIUM (Gemini line 111) — DecodeSegment defensive length check**
`DecodeSegment` now rejects inputs longer than `maxSegmentBytes` outright with `ErrInvalidEncodedSegment`. `EncodeSegment` never produces such inputs; any caller passing one is reading a corrupted dump or has a bug.

New tests:
- `TestEncodeSegment_KeyMatchingShaFallbackShapeIsPromotedToFallback` — the collision-shaped input
- `TestEncodeSegment_HugeInputDoesNotMaterialiseFullExpansion` — 1 MiB raw, asserts SHA fallback + bounded output
- `TestDecodeSegment_RejectsOversizedInput`
- `TestEncodeBinarySegment_HugeInputTakesShaFallbackWithoutEncoding`

@claude review

bootjp:
2 Codex P1 follow-ups addressed in `1e282382`:

**#100 — Overbroad fast path.** `len(raw)*3 > 240` falsely tripped for any input longer than 80 bytes, including pure-ASCII unreserved keys whose 1:1 encoding would have fit. Replaced with:
1. `len(raw) > maxSegmentBytes` → SHA fallback (1:1 lower bound, can't fit anyway).
2. `percentEncodeBounded` streams the encode and bails on overflow, capping the partial allocation at `maxSegmentBytes`. Restores the documented round-trip for 81–240 byte unreserved keys while preserving the OOM guard for adversarial inputs that DO need escaping.

`TestEncodeSegment_LongUnreservedASCIIEncodesAsIs` (200-byte ASCII round-trip) locks it in. The existing `TestEncodeSegment_HugeInputDoesNotMaterialiseFullExpansion` still passes because `percentEncodeBounded` never allocates more than `maxSegmentBytes + 3`.

**#146 — `b64.` prefix collision.** A user STRING key like `b64.foo` was returned as-is (all unreserved) and then misclassified by `DecodeSegment` as binary, decoding the base64 to `foo` instead of round-tripping. `EncodeSegment` now promotes any input whose encoded form starts with `b64.` to a real SHA fallback (parallel to the existing SHA-shape collision check). KEYMAP.jsonl carries the original bytes.

`TestEncodeSegment_KeyStartingWithBinaryPrefixIsPromotedToFallback`.

@claude review

PR #712, opened 2026-04-29
backup: KEYMAP.jsonl writer/reader and MANIFEST.json schema (Phase 0a)

bootjp:
## Summary

Stacked on top of #711 (filename encoding). Adds two more foundation pieces of the Phase 0 logical-backup decoder.

**KEYMAP.jsonl** (`internal/backup/keymap.go`)
- Append-only JSONL stream of `{encoded, original (b64url), kind}` records.
- Records exist only when the original bytes are NOT recoverable from the encoded filename alone:
  - `KindSHAFallback` — segments rendered as `<sha32>__<truncated>`
  - `KindS3LeafData` — S3 path collisions renamed to `.elastickv-leaf-data`
  - `KindMetaCollision` — user object key ending in `.elastickv-meta.json`
- `KeymapWriter`: streaming append, JSON encoder configured to skip HTML escapes so user-key bytes round-trip cleanly. Refuses empty `encoded` or `kind` so producer bugs surface loudly. `Count()` exposed for the "omit empty file" decision.
- `KeymapReader`: line-by-line scanner with bounded buffer (1 MiB); blank lines surface as `ErrInvalidKeymapRecord` rather than being silently skipped so truncated dumps are recognised.
- `LoadKeymap`: convenience helper that materialises the file as a map (last-wins on duplicates).

**MANIFEST.json** (`internal/backup/manifest.go`)
- Structs matching the schema in `docs/design/2026_04_29_proposed_snapshot_logical_decoder.md`.
- `CurrentFormatVersion = 1`; `ReadManifest` refuses `format_version > current` and `format_version == 0` (`ErrUnsupportedFormatVersion`).
- Phase discriminator: Phase 0 must not set `Live`, Phase 1 must not set `Source` — both validated at write and read time.
- `DisallowUnknownFields` on read so format drift surfaces loudly.
- Pretty-printed output (2-space indent, no HTML escapes) since `MANIFEST.json` is operator-facing.
- `NewPhase0SnapshotManifest` seeds policy fields with the documented defaults.

## Test plan

- [x] Tests cover round-trip, sticky-error semantics, last-wins dedup, HTML-escape suppression, future-version refusal, unknown-field refusal, unknown-phase refusal, cross-phase `Source`/`Live` exclusion.

## Self-review

- **Data loss** — N/A (read/write helpers). `KeymapReader` returns sticky errors so partial reads cannot be silently treated as success.
- **Concurrency** — `KeymapWriter`/`KeymapReader` are not goroutine-safe (per-scope use); manifest helpers are pure. `-race` clean.
- **Performance** — `bufio.Writer`(64 KiB) for the JSONL stream; bounded scanner buffer (1 MiB) on read.
- **Data consistency** — `DisallowUnknownFields` + format-version gate prevent silent drift. The phase discriminator's structural rules are enforced symmetrically at write and read.
- **Test coverage** — 7 keymap tests + 10 manifest tests covering the documented happy/sad paths.

## Stacking

Base branch is `feat/backup-phase0a-filename` (PR #711). When that lands, this PR's base will switch to `main` automatically.

### DIFF
--- a/internal/backup/keymap.go
+++ b/internal/backup/keymap.go
@@ -155,6 +155,11 @@ func NewKeymapReader(r io.Reader) *KeymapReader {
 // (zero, false, nil) at end of stream, and (zero, false, err) on parse
 // failure or I/O error. Once an error is returned the reader is sticky:
 // subsequent calls return the same error.
+//
+// The base64-encoded `original` field is validated at parse time rather
+// than lazily: a malformed dump must surface on the first read of the
+// affected line, not propagate silently until a much later
+// rec.Original() call. Same error class either way.
 func (r *KeymapReader) Next() (KeymapRecord, bool, error) {
 	if r.err != nil {
 		return KeymapRecord{}, false, r.err
@@ -174,6 +179,10 @@ func (r *KeymapReader) Next() (KeymapRecord, bool, error) {
 	}
 	if rec.Encoded == "" || rec.Kind == "" {
 		r.err = errors.Wrap(ErrInvalidKeymapRecord, "missing encoded or kind")
+		return KeymapRecord{}, false, r.err
+	}
+	if _, err := base64.RawURLEncoding.DecodeString(rec.OriginalB64); err != nil {
+		r.err = errors.Wrap(ErrInvalidKeymapRecord, err.Error())
 		return KeymapRecord{}, false, r.err
 	}
 	return rec, true, nil
--- a/internal/backup/keymap.go
+++ b/internal/backup/keymap.go
@@ -172,20 +172,47 @@ func (r *KeymapReader) Next() (KeymapRecord, bool, error) {
 		return KeymapRecord{}, false, nil
 	}
 	line := r.sc.Bytes()
-	var rec KeymapRecord
-	if err := json.Unmarshal(line, &rec); err != nil {
-		r.err = errors.Wrap(ErrInvalidKeymapRecord, err.Error())
-		return KeymapRecord{}, false, r.err
-	}
-	if rec.Encoded == "" || rec.Kind == "" {
-		r.err = errors.Wrap(ErrInvalidKeymapRecord, "missing encoded or kind")
-		return KeymapRecord{}, false, r.err
-	}
-	if _, err := base64.RawURLEncoding.DecodeString(rec.OriginalB64); err != nil {
-		r.err = errors.Wrap(ErrInvalidKeymapRecord, err.Error())
+	rec, err := decodeKeymapLine(line)
+	if err != nil {
+		r.err = err
 		return KeymapRecord{}, false, r.err
 	}
 	return rec, true, nil
+}
+
+// decodeKeymapLine parses one JSONL record. It enforces three properties:
+//
+//  1. The record must contain `encoded`, `original`, and `kind` fields —
+//     a missing `original` would otherwise be silently rewritten to empty
+//     bytes by base64.RawURLEncoding.DecodeString(""). Codex P2 round 5.
+//  2. `encoded` and `kind` must be non-empty strings.
+//  3. `original` (the base64) must be parseable at parse time so a
+//     corrupted dump fails on first read rather than at later
+//     Original() call. Codex P1 #179.
+func decodeKeymapLine(line []byte) (KeymapRecord, error) {
+	// Two-phase decode: first into a presence-aware map so we can
+	// distinguish "field absent" from "field present and empty
+	// string"; then into the typed struct for value extraction.
+	var fields map[string]json.RawMessage
+	if err := json.Unmarshal(line, &fields); err != nil {
+		return KeymapRecord{}, errors.Wrap(ErrInvalidKeymapRecord, err.Error())
+	}
+	for _, name := range [...]string{"encoded", "original", "kind"} {
+		if _, ok := fields[name]; !ok {
+			return KeymapRecord{}, errors.Wrapf(ErrInvalidKeymapRecord, "missing field %q", name)
+		}
+	}
+	var rec KeymapRecord
+	if err := json.Unmarshal(line, &rec); err != nil {
+		return KeymapRecord{}, errors.Wrap(ErrInvalidKeymapRecord, err.Error())
+	}
+	if rec.Encoded == "" || rec.Kind == "" {
+		return KeymapRecord{}, errors.Wrap(ErrInvalidKeymapRecord, "missing encoded or kind")
+	}
+	if _, err := base64.RawURLEncoding.DecodeString(rec.OriginalB64); err != nil {
+		return KeymapRecord{}, errors.Wrap(ErrInvalidKeymapRecord, err.Error())
+	}
+	return rec, nil
 }
 
 // LoadKeymap reads every record from r into an in-memory map keyed by
--- a/internal/backup/keymap.go
+++ b/internal/backup/keymap.go
@@ -41,13 +41,18 @@ const (
 // without holding pathological amounts of memory.
 const keymapBufSizeWriter = 64 << 10
 
-// keymapBufSizeReader bounds bufio.Scanner's per-line buffer. KEYMAP records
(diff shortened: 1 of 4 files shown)


bootjp:
Review feedback addressed in `c3d1e6fd`:

**Gemini MEDIUM #98 — Adapters nil-vs-empty must be distinguishable on disk**
The non-pointer `Adapter` fields collapsed "excluded by --adapter filter" and "included with no scopes" into the same on-disk shape. Switched to `*Adapter` pointers + `omitempty`. Now: `nil` → excluded (key absent), non-nil empty `&Adapter{}` → included with no scopes (`"dynamodb":{}`), non-nil populated → normal. New `TestAdaptersStruct_NilVsEmptyDistinguishedOnDisk` covers it.

**Codex P1 #179 — Validate base64 at parse time**
`KeymapReader.Next` previously deferred base64 validation to `rec.Original()`. A malformed dump would parse fine and only fail on a much later read. Now the base64 is decoded inside `Next()` so corruption surfaces on first ingest. New `TestKeymapReader_RejectsMalformedBase64AtParseTime`.

**Codex P2 #194 — Reject trailing bytes after MANIFEST JSON**
`ReadManifest` now uses `dec.More()` after the single decode and rejects with `ErrInvalidManifest` if anything decodable remains. Catches concatenation bugs, partial writes, and double-write corruption. New `TestReadManifest_RejectsTrailingBytes` and `TestReadManifest_RejectsTrailingNonWhitespace`.

**Gemini MEDIUM #160 — Test name fix**
`TestKeymapReader_AcceptsBlankLinesByPolicy` → `TestKeymapReader_RejectsBlankLines`. The test was already asserting rejection; only the name lied.

@claude review

bootjp:
**Codex P1 #295** addressed in `e4cd7d8e`. Phase 1 manifests now require `Live != nil` AND `Live.ReadTS != 0`; both `WriteManifest` and `ReadManifest` reject a phase1 dump that omits its cluster-wide pin record. New tests: `TestManifest_Phase1MustSetLive`, `TestManifest_Phase1RejectsZeroReadTS`, `TestManifest_Phase1WithLiveAndNonZeroReadTSIsValid`.

@claude review

bootjp:
**Codex P1 #205** (round 2) addressed in `6f231fe7`. The earlier `DisallowUnknownFields` was too strict — it broke the documented same-major minor-evolution path. A newer producer that adds an optional field at the same `format_version` would turn older readers into hard read failures during mixed-version operation.

`format_version` IS the breaking-change signal: major bumps are gated by `ErrUnsupportedFormatVersion`, same-major minor evolution must silently tolerate optional new fields. Removed `dec.DisallowUnknownFields()`. The trailing-bytes guard (Codex P2 #194 from round 1) is preserved — it catches concatenation/corruption, which is a different concern.

Test `TestReadManifest_RejectsUnknownFields` replaced by `TestReadManifest_AcceptsUnknownFieldsForSameMajorMinorEvolution`.

@claude review

bootjp:
**Codex P2 #146** (round 3) addressed in `73641330`. `ReadManifest` previously accepted manifests where `adapters` and/or `exclusions` sections were omitted because they decoded into non-pointer struct fields and validate() only checked scalar policy fields. The empty zero-value sections passed silently, dropping provenance and masking truncation.

`Manifest.Adapters` / `Manifest.Exclusions` are now pointer types. `validateRequiredFields` rejects `nil` for either with `ErrInvalidManifest`. `NewPhase0SnapshotManifest` seeds both to non-nil zero values so producers satisfy the contract by default.

Tests: `TestReadManifest_RejectsMissingAdapters`, `TestReadManifest_RejectsMissingExclusions`.

@claude review

bootjp:
@claude review

Round 5 fixes for Codex P2 findings on commit 736413302f:

1. `KeymapReader.Next` (keymap.go) — now decodes into a presence-aware `map[string]json.RawMessage` first and rejects records missing any of {encoded, original, kind}, before unmarshaling into the typed struct. Previously a missing `original` field passed through because `base64.DecodeString("")` succeeds, silently rewriting the mapping to empty bytes. New regression tests `TestKeymapReader_RejectsMissingOriginalField` (rejection) and `TestKeymapReader_AcceptsExplicitEmptyOriginal` (explicit empty stays valid).

2. `ReadManifest` (manifest.go) — split into a relaxed format_version probe (only `format_version uint32`) followed by the strict struct decode. A future-major manifest that also changes the JSON type of a known field (e.g. `phase` string → int) now surfaces as `ErrUnsupportedFormatVersion` instead of `ErrInvalidManifest`, restoring the advertised version-branching contract. New regression test `TestReadManifest_FutureMajorVersionTakesPrecedenceOverTypeMismatch`.

Verified locally: `go test -race ./internal/backup/...` and `golangci-lint run` clean.

bootjp:
@claude review

Round 6 fix for Codex P1 on commit 2cd58a93:

The source store caps a single snapshot key at 1 MiB (`store/mvcc_store.go maxSnapshotKeySize`), but `keymapBufSizeReader` was also 1 MiB. KEYMAP records carry the key as base64url which expands ~4/3 — so a maximum-sized valid key produced a ~1.33 MiB record that `bufio.Scanner` refused with "token too long", and the dump could not be loaded back via `LoadKeymap` / `KeymapReader`. Bumped the scanner cap to 4 MiB (2× margin over the theoretical worst case).

Test: `TestKeymapReader_AcceptsMaxSizedOriginal` round-trips a 1 MiB original key through writer→reader. `go test -race` and `golangci-lint` clean. Fix merged forward into #713 / #714 / #716 / #718.

bootjp:
@claude review

Round 7 fixes for Codex round 7 on commit 0afb5e3a (commit `6b2dea00`):

1. **P1 — null KEYMAP fields.** `decodeKeymapLine` was rejecting missing fields but accepting `"original": null` because Go's `json.Unmarshal` collapses null and "" for string targets, and `base64.DecodeString("")` succeeds. Added an explicit `json.RawMessage` compare against the literal `null` token; the same guard applies to `encoded` and `kind` for symmetry. Test: `TestKeymapReader_RejectsExplicitNullField` (3 sub-cases).

2. **P2 — missing exclusion flags.** `validateRequiredFields` only checked `Exclusions != nil`. Each bool field defaulted to `false` on absence, so a manifest dropping `preserve_sqs_visibility` (or any of the four) silently passed with altered exclusion semantics — losing the producer-side provenance the section is meant to capture. Added `validateExclusionsFieldsPresent` that walks the raw JSON and rejects manifests missing or null'ing any flag. Test: `TestReadManifest_RejectsMissingExclusionFlag` (5 sub-cases).

`go test -race` and `golangci-lint` clean. Fix merged forward into #713 / #714 / #716 / #718.

bootjp:
@claude review

Round 8 fixes for Codex P2 round 8 on commit 6b2dea00 (commit `f13cd1bf`):

1. **Missing `format_version` misclassified as upgrade-required.** The probe decoded directly into `uint32`, so an absent field collapsed silently to 0 and the gate returned `ErrUnsupportedFormatVersion` ("format_version is zero"). A truncated/malformed manifest belongs in `ErrInvalidManifest` so callers branching on `errors.Is` can distinguish "this file is broken" from "newer producer". Now a presence-and-null probe runs on the raw JSON `format_version` token before the typed unmarshal; the whole probe is extracted into `probeManifestFormatVersion` to keep `ReadManifest`'s cyclomatic complexity under the linter ceiling. Tests: `TestReadManifest_RejectsMissingFormatVersion` (also asserts the error is *not* `ErrUnsupportedFormatVersion`) and `TestReadManifest_RejectsNullFormatVersion`.

2. **`WriteManifest` accepted future format_version.** A caller mutating `m.FormatVersion = CurrentFormatVersion + 1` could write a manifest that the same package's `ReadManifest` then refuses — self-incompatible backup metadata for a one-line mutation. `validateRequiredFields` now rejects that case with `ErrInvalidManifest`. Test: `TestWriteManifest_RejectsFutureFormatVersion`.

`go test -race` and `golangci-lint` clean. Fix merged forward into #713 / #714 / #716 / #718.

PR #713, opened 2026-04-29
backup: Redis simple-type encoders (strings, HLL, TTL routing)

bootjp:
## Summary

Stacked on top of #712. Adds the first per-adapter encoder of the Phase 0 logical-backup decoder — covers the three Redis prefixes that always map ONE snapshot record to one user key:

- `!redis|str|<userKey>` → `strings/<encoded>.bin` (+ inline-TTL stripped into `strings_ttl.jsonl`). Both new magic-prefix format and legacy raw bytes accepted.
- `!redis|hll|<userKey>` → `hll/<encoded>.bin` (raw HLL sketch).
- `!redis|ttl|<userKey>` → routed by prior type observation:
  - `redisKindString` → `strings_ttl.jsonl` (legacy strings)
  - `redisKindHLL` → `hll_ttl.jsonl`
  - `redisKindUnknown` → buffered for wide-column types; orphan warning at `Finalize`.

Wide-column types (hash / list / set / zset / stream) ship in follow-up PRs.

## How TTL routing works

Pebble lex order guarantees `!redis|hll|*` < `!redis|str|*` < `!redis|ttl|*` (h < s < t). So when `HandleTTL` is called, the type tracker already knows what kind of key the user-key belongs to. No two-pass needed.

## Safety

- Filenames go through `EncodeSegment` (the encoding contract from #711).
- Blob writes are atomic (tmp + rename); `writeFileAtomic` refuses to overwrite a symlink so a malicious or accidental link cannot redirect output bytes.
- Truncated magic-prefix headers and bad-length TTL records surface typed errors (`ErrRedisInvalidStringValue`, `ErrRedisInvalidTTLValue`) rather than silent data loss.
- The package keeps the `0xFF 0x01` magic constants local (not imported from `adapter/redis_compat_types.go`) because the decoder is intentionally adapter-independent — must run as an offline tool with no live cluster. A doc comment points back to the live source so a future renamer updates both.

## Test plan

- [x] 11 tests covering new/legacy/inline-TTL strings, HLL raw round-trip, TTL routing for both kinds, orphan-TTL warning, truncated header/TTL rejection, EncodeSegment integration, symlink-overwrite refusal, `HasInlineTTL` helper.

## Stacking

Base: `feat/backup-phase0a-keymap-manifest` (PR #712).

## Summary by CodeRabbit

* **New Features**
  * Per-Redis-database output routing so backups are written into per-DB locations.
  * TTL handling improved: inline TTL deduplication and orphan TTL counting reported at finalize.

* **Bug Fixes**
  * Stronger protections when creating sidecar files: refuses symlinks, hardlink clobbering, and non-regular files (e.g., FIFOs).
  * More reliable disk-full detection across platforms.

* **Tests**
  * Added extensive regression tests covering safety, routing, TTL, and out-of-space scenarios.

### DIFF
--- a/internal/backup/disk_full_other.go
+++ b/internal/backup/disk_full_other.go
@@ -1,0 +1,15 @@
+//go:build !unix && !windows
+
+package backup
+
+// isDiskFullError is the fallback for non-unix/non-windows targets
+// (js, wasip1, plan9). Those runtimes either do not surface a
+// disk-full errno through Go's syscall package or do not have a
+// meaningful disk concept (wasm with no host filesystem, plan9 with
+// its own error vocabulary). Returning false matches the documented
+// helper contract: callers treat unrecognised errors as
+// non-retriable, which is the safe default. Codex P2 round 10.
+func isDiskFullError(err error) bool {
+	_ = err
+	return false
+}
--- a/internal/backup/disk_full_unix.go
+++ b/internal/backup/disk_full_unix.go
@@ -1,0 +1,15 @@
+//go:build unix
+
+package backup
+
+import (
+	"errors"
+	"syscall"
+)
+
+// isDiskFullError reports whether err is a POSIX ENOSPC anywhere in
+// the wrap chain. os.File.Write surfaces ENOSPC inside an
+// os.PathError which errors.Is unwraps natively.
+func isDiskFullError(err error) bool {
+	return errors.Is(err, syscall.ENOSPC)
+}
--- a/internal/backup/disk_full_windows.go
+++ b/internal/backup/disk_full_windows.go
@@ -1,0 +1,27 @@
+//go:build windows
+
+package backup
+
+import (
+	"errors"
+	"syscall"
+
+	"golang.org/x/sys/windows"
+)
+
+// isDiskFullError reports whether err is a Windows disk-full failure.
+// We accept both ERROR_DISK_FULL (Win32 errno 112) and
+// ERROR_HANDLE_DISK_FULL (39): the kernel returns 112 for write
+// failures driven by free-space exhaustion and 39 for the legacy
+// handle-level variant some FS drivers still surface. We also keep
+// syscall.ENOSPC in the chain because Go's os package occasionally
+// translates platform errors into the POSIX value before returning.
+func isDiskFullError(err error) bool {
+	switch {
+	case errors.Is(err, windows.ERROR_DISK_FULL),
+		errors.Is(err, windows.ERROR_HANDLE_DISK_FULL),
+		errors.Is(err, syscall.ENOSPC):
+		return true
+	}
+	return false
+}
--- a/internal/backup/open_nofollow_other.go
+++ b/internal/backup/open_nofollow_other.go
@@ -21,5 +21,9 @@ func openSidecarFile(path string) (*os.File, error) {
 		return nil, cockroachdberr.WithStack(cockroachdberr.Newf(
 			"backup: refusing to overwrite symlink at %s", path))
 	}
-	return os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600) //nolint:gosec,mnd // path is composed from output-root + fixed file name; 0600 is the standard owner-only mode
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600) //nolint:gosec,mnd // path is composed from output-root + fixed file name; 0600 is the standard owner-only mode
+	if err != nil {
+		return nil, cockroachdberr.WithStack(err)
+	}
+	return f, nil
 }
--- a/internal/backup/open_nofollow_other.go
+++ b/internal/backup/open_nofollow_other.go
@@ -1,0 +1,25 @@
+//go:build !unix && !windows
+
+package backup
+
+import (
+	"os"
+
+	cockroachdberr "github.com/cockroachdb/errors"
+)
+
+// openSidecarFile is the fallback for non-unix/non-windows targets
+// (js, wasip1, plan9). syscall.O_NOFOLLOW and the unix nlink-check
+// path are unavailable; we keep a Lstat-then-OpenFile guard to at
+// least refuse pre-existing symlinks. The remaining TOCTOU window
+// is acceptable here because dump tooling on those targets is
+// offline / sandboxed and the threat model that motivated the unix
+// hardening (a local adversary swapping the path between syscalls)
+// does not apply. Codex P2 round 10.
+func openSidecarFile(path string) (*os.File, error) {
+	if info, err := os.Lstat(path); err == nil && info.Mode()&os.ModeSymlink != 0 {
+		return nil, cockroachdberr.WithStack(cockroachdberr.Newf(
+			"backup: refusing to overwrite symlink at %s", path))
+	}
+	return os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600) //nolint:gosec,mnd // path is composed from output-root + fixed file name; 0600 is the standard owner-only mode
+}
--- a/internal/backup/open_nofollow_unix.go
+++ b/internal/backup/open_nofollow_unix.go
@@ -10,20 +10,47 @@ import (
 	cockroachdberr "github.com/cockroachdb/errors"
 )
 
-// openSidecarFile opens path for write with truncate+create semantics
-// and atomically refuses symlinks. The unix build leans on
-// syscall.O_NOFOLLOW so the kernel returns ELOOP if path is a
-// symlink, eliminating the TOCTOU race a separate Lstat-then-Create
-// pattern would have. The Windows build (open_nofollow_windows.go)
-// implements this with the best alternative available there.
+// openSidecarFile opens path for write while refusing both symlink and
+// hard-link clobber attacks.
+//
+//   - O_NOFOLLOW makes the kernel return ELOOP atomically if the path
+//     is a symbolic link — closing the TOCTOU race a separate
+//     Lstat-then-Create pattern would have.
+//   - To also refuse hard links to files outside the dump tree, we
+//     open WITHOUT O_TRUNC, fstat() the descriptor to check the
+//     link count, and only call Truncate(0) if Nlink == 1. An
+//     adversary that pre-created strings_ttl.jsonl as a hard link
+//     to /etc/passwd (or any other writable file outside the dump
+//     tree) would otherwise see the inode truncated on
+//     openSidecarFile despite the symlink guard. Codex P2 round 9.
+//
+// The Windows build (open_nofollow_windows.go) keeps the simpler
+// Lstat-then-OpenFile guard because Windows's
+// SeCreateSymbolicLinkPrivilege already raises the bar for the
+// equivalent attack.
 func openSidecarFile(path string) (*os.File, error) {
-	const flag = os.O_WRONLY | os.O_CREATE | os.O_TRUNC | syscall.O_NOFOLLOW
+	// Note: NO O_TRUNC here — we truncate after the link-count check.
+	const flag = os.O_WRONLY | os.O_CREATE | syscall.O_NOFOLLOW
 	f, err := os.OpenFile(path, flag, 0o600) //nolint:gosec,mnd // path is composed from output-root + fixed file name; 0600 is the standard owner-only mode
 	if err != nil {
 		if errors.Is(err, syscall.ELOOP) {
 			return nil, cockroachdberr.WithStack(cockroachdberr.Wrapf(err,
 				"backup: refusing to overwrite symlink at %s", path))
 		}
+		return nil, cockroachdberr.WithStack(err)
+	}
+	info, err := f.Stat()
+	if err != nil {
+		_ = f.Close()
+		return nil, cockroachdberr.WithStack(err)
+	}
+	if sysStat, ok := info.Sys().(*syscall.Stat_t); ok && sysStat.Nlink > 1 {
+		_ = f.Close()
+		return nil, cockroachdberr.WithStack(cockroachdberr.Newf(
+			"backup: refusing to overwrite hard-linked file at %s (nlink=%d)", path, sysStat.Nlink))
+	}
+	if err := f.Truncate(0); err != nil {
+		_ = f.Close()
 		return nil, cockroachdberr.WithStack(err)
 	}
 	return f, nil
--- a/internal/backup/open_nofollow_unix.go
+++ b/internal/backup/open_nofollow_unix.go
@@ -1,0 +1,30 @@
+//go:build unix
+
+package backup
+
+import (
+	"errors"
+	"os"
+	"syscall"
+
+	cockroachdberr "github.com/cockroachdb/errors"
+)
+
+// openSidecarFile opens path for write with truncate+create semantics
+// and atomically refuses symlinks. The unix build leans on
+// syscall.O_NOFOLLOW so the kernel returns ELOOP if path is a
+// symlink, eliminating the TOCTOU race a separate Lstat-then-Create
+// pattern would have. The Windows build (open_nofollow_windows.go)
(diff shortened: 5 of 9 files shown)


bootjp:
All four Gemini MEDIUM items addressed in `945d73c9`:

**#189 — `pendingWideColumnTTL` unbounded**
Capped at `maxPendingWideColumnTTL = 1,000,000` (~50 MiB). Records past the cap are dropped and counted via the warn sink at `Finalize`. The cap is well above any plausible real workload (wide-column encoders will claim every TTL anyway when they ship in a follow-up), but small enough to protect against a malformed snapshot.

**#217 — Hardcoded `db_0`**
`NewRedisDB` now takes a `dbIndex` parameter; the per-encoder root `"<outRoot>/redis/db_<idx>/"` is computed from it. Two encoders sharing an `outRoot` with different db indices no longer collide. `TestRedisDB_PerDBIndexRoutesIntoOwnDirectory` covers it.

**#218 — `MkdirAll` per blob**
`dirsCreated map[string]struct{}` on `RedisDB`. `ensureDir()` short-circuits on a map lookup; for a 10M-key dump this saves ~10M `stat + mkdir(EEXIST)` round-trips. `TestRedisDB_DirsCreatedCachesMkdirAll` asserts each subdir is populated exactly once.

**#382 — ENOSPC**
Added `IsBlobAtomicWriteOutOfSpace` as the explicit ENOSPC probe (via `errors.Is(err, syscall.ENOSPC)`). `IsBlobAtomicWriteRetriable` continues to report only `io.ErrShortWrite` as retriable — ENOSPC is **deliberately not** retriable (a backup against a full disk should surface to the operator, not spin). The two-function split lets the master pipeline pick the right alarm wording.

@claude review

bootjp:
@claude review

Round 2 fixes for Codex round 5 on commit 84a1c5a3:

1. **P1: deduplicate string TTL records.** The live Redis encoder commits both `!redis|str|<k>` (with TTL embedded inline in the magic-prefix header) and the scan-index `!redis|ttl|<k>` for every expiring string (`adapter/redis_lua_context.go` `stringCommitElems`). `HandleString` was already writing `strings_ttl.jsonl` from the inline TTL and then `HandleTTL` was routing the redundant `!redis|ttl|` record to the same sidecar — duplicating the entry and breaking the one-record-per-key contract. Now `RedisDB` tracks inline-emitted keys in a new `inlineTTLEmitted` set and `HandleTTL` drops the scan-index TTL when the same key was already covered inline. Legacy (no-magic-prefix) strings still flow through the `!redis|ttl|` path because their TTL only lives there.

2. **P2: refuse symlink overwrite in openJSONL.** `writeFileAtomic` already `Lstat`-checks before overwriting; `openJSONL` was using `os.Create` directly, which follows symlinks. A `strings_ttl.jsonl` or `hll_ttl.jsonl` symlink in the output tree would have truncated a target outside the dump tree. Mirror the same defence in `openJSONL`.

Tests: `TestRedisDB_NewFormatStringTTLNotDuplicatedByScanIndex`, `TestRedisDB_OpenJSONLRefusesSymlinkOverwrite`. `go test -race` and `golangci-lint` clean. Fix merged forward into #714/#716/#718.

bootjp:
@claude review

Round 3 fixes for Codex round 6 on commit 1c70e83d:

1. **P1 — TOCTOU race on JSONL symlink check.** The Lstat-then-os.Create pattern from round 5 had a window where a writable output directory could be swapped to a symlink between the two syscalls, defeating the guard. Replaced with a single `os.OpenFile(path, O_WRONLY|O_CREATE|O_TRUNC|syscall.O_NOFOLLOW, 0o600)` call so the kernel refuses symlinks atomically (ELOOP on Linux/macOS/BSD; `syscall.O_NOFOLLOW == 0` on Windows where the filesystem-permission model is different anyway). The existing `TestRedisDB_OpenJSONLRefusesSymlinkOverwrite` test still asserts the same substring on the wrapped error.

2. **P2 — orphan TTL keys buffered but unused.** `HandleTTL` was `bytes.Clone`-ing each orphan user key into `pendingWideColumnTTL`, but `Finalize` only read `len(...)` for the warning sink. With user keys up to 1 MiB (`maxSnapshotKeySize`), this could allocate gigabytes for no benefit. Replaced the slice with a plain `orphanTTLCount int` and removed the now-unused `redisTTLPending` struct + `maxPendingWideColumnTTL` cap. The wide-column-encoder PR will re-introduce per-key tracking when those encoders actually consume the keys. Test `TestRedisDB_PendingWideColumnTTLBounded` replaced by `TestRedisDB_OrphanTTLCountedNotBuffered`.

`go test -race` and `golangci-lint` clean. Fix merged forward into #714 / #716 / #718.

bootjp:
@claude review

Round 4 fixes for Codex round 7 on commit d965b659:

1. **P1 — SHA-fallback KEYMAP entries (commit `4cb6d837`).** `writeBlob` and `appendTTL` now call `recordIfFallback(encoded, userKey)` which lazily opens a per-DB `KEYMAP.jsonl` (with the same atomic `O_NOFOLLOW` open as the TTL sidecars) and writes a `KindSHAFallback` record whenever `EncodeSegment` took its non-reversible path. Without this, restore tools cannot recover Redis user keys longer than 240 bytes (or those promoted to the fallback by `b64.` / SHA-shape collisions). `closeKeymap` (called from Finalize) removes the file if no fallback records were ever emitted, matching the s3 encoder's empty-keymap policy. Tests: `TestRedisDB_SHAFallbackKeymapped` (round-trips a 300-byte key) and `TestRedisDB_NoKeymapWhenAllReversible`.

2. **P1 — `isNewRedisStrFormat` 0xFF 0x01 ambiguity (deferred).** This is a pre-existing limitation of the live encoder/decoder (`adapter/redis_compat_types.go` shares the exact same two-byte discriminator). The backup decoder must mirror the live decoder's parse, otherwise a backup→restore round-trip would diverge from the value the live cluster was already returning to clients for that key. Strengthening the discriminator requires a coordinated wire-format change to the live encoder (and a migration plan for existing 0xFF-0x01-prefixed legacy values), which is out of scope for the Phase 0a backup decoder PR. Filed as a follow-up; happy to write the design doc separately if that's the right next step.

`go test -race` and `golangci-lint` clean. SHA-fallback fix merged forward into #714 / #716 / #718.

bootjp:
@claude review

Round 5 fix for Codex P2 round 7 on commit d965b659 (commit `6f354464`):

`syscall.O_NOFOLLOW` is not defined on Windows, so the round-3 fix that used it directly broke the cross-platform build. Reproduced with `GOOS=windows GOARCH=amd64 go build ./internal/backup` failing at `undefined: syscall.O_NOFOLLOW`.

Extracted the per-platform open into a build-tagged helper `openSidecarFile`:
- `open_nofollow_unix.go` (`//go:build unix`) keeps the atomic O_NOFOLLOW open with ELOOP wrapping.
- `open_nofollow_windows.go` (`//go:build windows`) uses Lstat-then-OpenFile. The residual TOCTOU window is acceptable on Windows because mounting a successful attack already requires output-directory write access plus `SeCreateSymbolicLinkPrivilege` — a much higher bar than the unix case.

Both `openJSONL` and the keymap-creation path in `recordIfFallback` now go through the helper. `go test -race` passes on Linux/macOS; `GOOS=windows GOARCH=amd64 go build ./internal/backup/...` now succeeds (was the Codex repro). Lint clean. Fix merged forward into #714 / #716 / #718.

bootjp:
@claude review

Round 6 fix for Codex P2 round 9 on commit 87971aef (commit `ae3f0d9f`):

The round-5 unix opener kept `O_TRUNC` on the same call as `O_NOFOLLOW`. `O_NOFOLLOW` only blocks symlinks — an adversary who can write the output directory could pre-create `strings_ttl.jsonl` / `hll_ttl.jsonl` / `KEYMAP.jsonl` as a hard link to a writable file outside the dump tree (e.g. `/etc/passwd`) and `O_TRUNC` would zero that inode despite the symlink guard.

Restructured the unix open:
1. Open with `O_WRONLY|O_CREATE|O_NOFOLLOW` (no `O_TRUNC`).
2. `fstat()` the descriptor.
3. If `Nlink > 1`, refuse with a wrapped error and close.
4. Otherwise call `f.Truncate(0)` — atomic with the verified-single-link descriptor.

The Windows build is unchanged because the equivalent attack model is materially different there (`SeCreateSymbolicLinkPrivilege` required, ACLs gate the directory) and link-count semantics under fstat are nontrivial.

Test: `TestRedisDB_OpenJSONLRefusesHardLinkClobber` pre-creates the sidecar as a hard link to a bait file, drives `HandleString`, and asserts the bait file is untouched.

`go test -race`, `GOOS=windows GOARCH=amd64 go build`, and `golangci-lint` clean. Fix merged forward into #714 / #716 / #718.

bootjp:
@claude review

Round 7 fixes for Codex round 9 on commit 87971aef (commit `e061b6d0`):

1. **P1 — symlinked ancestors bypassed the guard.** `O_NOFOLLOW` and `Lstat(path)` only check the trailing component. A pre-existing directory symlink anywhere up the path (e.g. `<outRoot>/redis/db_0/strings -> /tmp/outside`) let `MkdirAll` silently honor it and steered subsequent writes outside `outRoot`. `ensureDir` now Lstat-walks each ancestor under `outRoot` via `assertNoSymlinkAncestors` and refuses if any is a symlink. Best-effort against TOCTOU (full race-free traversal would need `os.Root` / openat); closes the much more common stale-symlink-in-output case. Test: `TestRedisDB_RefusesSymlinkedAncestor`.

2. **P2 — Windows disk-full not detected.** `IsBlobAtomicWriteOutOfSpace` only checked `syscall.ENOSPC`, missing `ERROR_DISK_FULL` (112) / `ERROR_HANDLE_DISK_FULL` (39). Moved the platform-specific check into `disk_full_unix.go` (ENOSPC) and `disk_full_windows.go` (both Win32 codes + ENOSPC for the Go-translated case via `golang.org/x/sys/windows`).

`go test -race`, `GOOS=windows GOARCH=amd64 go build`, and `golangci-lint` clean. Fix merged forward into #714 / #716 / #718.

bootjp:
@claude review

Round 8 fix for Codex P2 round 10 on commit e061b6d0 (commit `20236ac4`):

The unix/windows split for `openSidecarFile` and `isDiskFullError` did not cover `GOOS=js GOARCH=wasm` or `GOOS=wasip1 GOARCH=wasm`, so cross-compilation failed with `undefined: openSidecarFile` / `undefined: isDiskFullError`. Codex reproduced via `GOOS=js go test -c ./internal/backup`.

Added `open_nofollow_other.go` and `disk_full_other.go` behind `//go:build !unix && !windows`:
- `openSidecarFile` falls back to Lstat-then-OpenFile (the wasm/plan9 threat model doesn't include the local-adversary scenario the unix nlink+atomic-open guard targets).
- `isDiskFullError` returns false (wasm has no disk; plan9 has a different error vocabulary; the helper contract treats unrecognised errors as non-retriable).

Verified: `GOOS=js GOARCH=wasm go build ./internal/backup/...`, `GOOS=wasip1 GOARCH=wasm go build ./internal/backup/...`, `GOOS=windows go build`, and `go test -race ./internal/backup/...` all clean. Lint clean. Fix merged forward into #714 / #716 / #718.

bootjp:
@claude review

Round 9 fix for Codex P2 round 11 on commit 20236ac4 (commit `1a5165e2`):

`openSidecarFile` opened the TTL sidecar with `O_WRONLY|O_CREATE|O_NOFOLLOW` and only checked the descriptor's link count after open. A pre-existing FIFO at `strings_ttl.jsonl` / `hll_ttl.jsonl` would block the first write indefinitely — POSIX: opening a reader-less FIFO with `O_WRONLY` hangs until a reader attaches. The symlink and hard-link guards don't catch this: `mkfifo` produces `nlink=1` and is not a symlink.

Added `O_NONBLOCK` to the unix open so a reader-less FIFO surfaces ENXIO immediately instead of hanging, then enforce `info.Mode().IsRegular()` after fstat. Reader-attached FIFOs, sockets, and character/block devices all refuse with a stable "non-regular file" error. Windows path unchanged (no FIFO concept there).

Test: `TestRedisDB_OpenJSONLRefusesFIFO` lives in `redis_string_unix_test.go` behind `//go:build unix` because `syscall.Mkfifo` is undefined on Windows and js/wasm.

Verified clean: `go test -race`, `GOOS=windows GOARCH=amd64 go build`, `GOOS=js GOARCH=wasm go build`, `golangci-lint`. Fix merged forward into #714 / #716 / #718.

bootjp:
@claude review

Round 13 fix for Codex P1 round 13 on commit 48727046 (commit `0383924b`):

`writeBlob` already routed through `ensureDir` + `assertNoSymlinkAncestors` (round 7), but TTL sidecars (`appendTTL` → `openJSONL`) bypassed that guard. `openJSONL` ran `os.MkdirAll(filepath.Dir(path))` then `openSidecarFile` — which protects only the final path element. A symlinked ancestor like `<outRoot>/redis/db_0 -> /tmp/outside` would still redirect `strings_ttl.jsonl` writes outside the dump root.

`appendTTL` now calls `r.ensureDir(r.dbDir())` before `openJSONL` so the same ancestor-walk protection applies.

Caller audit (per loop's audit-semantics rule): `openJSONL` has two callers — `appendTTL` (this fix) and `sqs.go`'s `flushQueue`/`flushInternals`. The SQS path bypasses RedisDB.ensureDir because SQS is a separate encoder; tracking the same hardening for SQS is out-of-scope for this PR's redis fix.

Test: `TestRedisDB_RefusesSymlinkedAncestorOnTTLSidecar` pre-places `<outRoot>/redis/db_0 -> /tmp/bait` and asserts `HandleString` (which drives appendTTL) refuses with the standard "refusing to traverse symlinked ancestor" error. `go test -race`, `GOOS=windows go build`, lint clean.

bootjp:
@claude review

Round 14 fix for Codex P1 round 13 follow-up on commit 0383924b (commit `8e8f8dda`):

`ensureDir`'s `dirsCreated` cache short-circuited BOTH `MkdirAll` AND `assertNoSymlinkAncestors`. After one successful write, an adversary that can write the output directory could swap the cached real dir for a symlink to outside the dump root, and subsequent `writeBlob` / `appendTTL` calls would follow it without re-validation — reintroducing the path-escape vector that the round-9 ancestor check was meant to close.

Restructured `ensureDir` so the cache only short-circuits `MkdirAll`; `assertNoSymlinkAncestors` runs on every call. Repeated symlink walks are an acceptable cost on the hot path because they touch ~5 ancestor entries per dump-root depth and are dominated by write I/O.

Caller audit (per loop's audit-semantics rule for behavior change): three callers in redis_string.go (`writeBlob`, `appendTTL`, `recordIfFallback`) all already check the error return and propagate. No API change.

Test: `TestRedisDB_EnsureDirRevalidatesAfterCachedSuccess` seeds the cache via a first `HandleString`, then swaps the cached real dir for a symlink targeting `<bait>/`, then drives a second `HandleString` and asserts the standard "refusing to traverse symlinked ancestor" refusal plus that the bait dir is untouched. `go test -race` and `golangci-lint` clean.

bootjp:
@claude review

Round 15 fix for Codex P2 round 14 on commit 8e8f8dda (commit `011b0318`):

The round-9 FIFO test used `syscall.Mkfifo` behind `//go:build unix`, but the standard library's syscall package does not define Mkfifo on every `unix`-tagged target — Solaris is the documented gap. `GOOS=solaris go test -c ./internal/backup` errored at compile time on the symbol, breaking cross-Unix portability claimed by the build tag.

Switched to `golang.org/x/sys/unix.Mkfifo` which is defined consistently across every `unix` target. The dependency is already in go.mod (used by `disk_full_windows.go` for ERROR_DISK_FULL constants).

Verified:
- `GOOS=solaris GOARCH=amd64 go test -c ./internal/backup` → clean (was the Codex repro).
- `go test -race ./internal/backup/...` on linux → clean.
- `golangci-lint run` → clean.

Test-only file; no production change. Fix merged forward into #714 / #716 / #725.

PR #715, opened 2026-04-30
feat(sqs): partition resolver for HT-FIFO routing (Phase 3.D PR 4-B-2)

bootjp:
## Summary

Routing-layer half of PR 4-B. Adds a `PartitionResolver` that `ShardRouter` consults BEFORE falling through to the byte-range engine. SQS HT-FIFO needs partition-aware dispatch, but the engine's non-overlapping-cover model can't express overlay routes — the resolver-first dispatch sidesteps this cleanly.

Stacks on top of #708 (PR 4-B-1, capability JSON). Next is PR 4-B-3 (leadership-refusal + catalog polling + flip `htfifoCapabilityAdvertised` to `true`).

## What's added

- `kv.PartitionResolver` interface — `ResolveGroup([]byte) (uint64, bool)`.
- `kv.ShardRouter.WithPartitionResolver(...)` — fluent option, nil-safe.
- `kv.ShardRouter.resolveGroup(...)` — unified dispatch path: resolver first, engine fallback. Both `groupRequests` (Commit/Abort) and `Get` route through it.
- `kv.ShardedCoordinator.WithPartitionResolver(...)` — delegates to the router so `main.go` can install via the existing fluent-construction style.
- `adapter.SQSPartitionResolver` — parses `(queue, partition)` from the partitioned key shape, looks up the operator-chosen group. Defensive copy at construction, nil-safe `ResolveGroup`, returns `(0, false)` for legacy / non-matching keys.
- `main.go` — builds the resolver from `runtimeConfig.sqsFifoPartitionMap` and installs it. Resolver is `nil` on a non-partitioned cluster — hot path stays engine-only.

## What's NOT added (deferred to PR 4-B-3)

- §8 leadership-refusal hook in `kv` (refuses leadership for an SQS Raft group hosting a partitioned queue when the binary lacks `htfifo`).
- Catalog-polling helper for the CreateQueue capability gate (PR 5 starts using it).
- Flipping `htfifoCapabilityAdvertised` from `false` to `true`.

The design's "advertise htfifo only when both routing AND leadership-refusal are in place" rule keeps the constant `false` in this PR — PR 4-B-3 flips it.

## Test plan

- [x] `adapter/sqs_partition_resolver_test.go` — 9 top-level tests: nil-on-empty, defensive-copy, partition dispatch across all 5 families, queue-name prefix isolation (`"queue"` vs `"queue1"`), legacy fall-through (8 sub-cases), unknown queue, out-of-range partition, nil receiver, prefix alignment with `sqs_keys.go` constants.
- [x] `kv/shard_router_partition_test.go` — 4 tests: resolver wins over engine, engine fallthrough on resolver-miss, nil resolver no-op, `Get` path also routes through the resolver.
- [x] `go test -race ./kv/...` pass.
- [x] `go test -race ./adapter/...` pass.
- [x] `golangci-lint ./kv/... ./adapter/... .` clean.

1. **Data loss** — routing layer only; no FSM/Pebble/retention path. No issue.
2. **Concurrency** — `partitionResolver` is set once at startup before any request. `ResolveGroup` reads a constructor-time defensive copy, so a future hot-reload of `--sqsFifoPartitionMap` cannot perturb in-flight requests. No issue.
3. **Performance** — one map lookup + 4-byte BigEndian decode per resolver hit (only on partitioned-prefix matches). Engine-only path adds a single `if s.partitionResolver != nil` branch. No issue.
4. **Data consistency** — resolver output strictly OVERRIDES the engine for partitioned keys; legacy keys flow through unchanged. "queue not found" / "partition out of range" branches return `(0, false)` so the router surfaces an explicit error rather than silently mis-routing. No issue.
5. **Test coverage** — 13 tests across two new files; existing `TestShardRouter*` tests unchanged. Both override and fall-through paths pinned, plus the queue-name prefix-isolation invariant from PR #703.

## Summary by CodeRabbit

* **New Features**
  * Added partition-based routing for distributed coordination. Requests are now directed to specific nodes based on configured partition mappings.
  * Enhanced SQS queue handling with optional partition-aware routing. When partition information is unavailable or unconfigured, the system automatically falls back to the existing routing mechanism.

### DIFF
--- a/adapter/sqs_partition_resolver.go
+++ b/adapter/sqs_partition_resolver.go
@@ -68,6 +68,14 @@ var sqsResolverFamilyPrefixes = [][]byte{
 // family prefix (legacy SQS, KV, S3, DynamoDB, queue-meta records,
 // …) so kv.ShardRouter falls through to its byte-range engine for
 // default routing.
+//
+// Returns (0, false) for a partitioned-shaped key whose queue is
+// not in the routes map or whose partition index is beyond
+// len(routes[queue]). The router pairs this with
+// RecognisesPartitionedKey to fail closed instead of falling
+// through — silently routing through the engine's
+// !sqs|route|global default would mis-route HT-FIFO traffic during
+// partition-map drift (codex P1 round 2 on PR #715).
 func (r *SQSPartitionResolver) ResolveGroup(key []byte) (uint64, bool) {
 	if r == nil || len(key) == 0 {
 		return 0, false
@@ -88,6 +96,26 @@ func (r *SQSPartitionResolver) ResolveGroup(key []byte) (uint64, bool) {
 		return 0, false
 	}
 	return groups[partition], true
+}
+
+// RecognisesPartitionedKey reports whether key has the structural
+// shape of a partitioned-SQS key — a partitioned family prefix
+// followed by an encoded queue segment, a '|' terminator, and a
+// fixed-width uint32 partition index. Implementations of
+// kv.PartitionResolver must answer purely on shape so the router
+// can use the predicate to decide between fall-through (not
+// partitioned) and fail-closed (partitioned but unresolved); see
+// kv.PartitionResolver doc and codex P1 round 2 on PR #715.
+//
+// A nil receiver returns false so kv.ShardRouter's typed-nil case
+// (ResolveGroup(nil) == (0, false)) pairs with an honest "I don't
+// recognise anything" answer instead of falsely claiming a shape.
+func (r *SQSPartitionResolver) RecognisesPartitionedKey(key []byte) bool {
+	if r == nil || len(key) == 0 {
+		return false
+	}
+	_, _, ok := parsePartitionedSQSKey(key)
+	return ok
 }
 
 // parsePartitionedSQSKey extracts the (queue, partition) pair from
--- a/adapter/sqs_partition_resolver.go
+++ b/adapter/sqs_partition_resolver.go
@@ -99,13 +99,14 @@ func (r *SQSPartitionResolver) ResolveGroup(key []byte) (uint64, bool) {
 }
 
 // RecognisesPartitionedKey reports whether key has the structural
-// shape of a partitioned-SQS key — a partitioned family prefix
-// followed by an encoded queue segment, a '|' terminator, and a
-// fixed-width uint32 partition index. Implementations of
-// kv.PartitionResolver must answer purely on shape so the router
-// can use the predicate to decide between fall-through (not
-// partitioned) and fail-closed (partitioned but unresolved); see
-// kv.PartitionResolver doc and codex P1 round 2 on PR #715.
+// shape of a partitioned-SQS key — i.e. starts with one of the
+// partitioned family prefixes. The check is PREFIX-ONLY, not a
+// full parse: a key with a partitioned prefix followed by a
+// malformed queue / partition segment still answers true, so the
+// router fails closed via kv.PartitionResolver semantics instead
+// of falling through to the engine and silently routing to the
+// SQS catalog default group via routeKey's !sqs|route|global
+// collapse (round 5 review nit on PR #715).
 //
 // A nil receiver returns false so kv.ShardRouter's typed-nil case
 // (ResolveGroup(nil) == (0, false)) pairs with an honest "I don't
@@ -114,7 +115,7 @@ func (r *SQSPartitionResolver) RecognisesPartitionedKey(key []byte) bool {
 	if r == nil || len(key) == 0 {
 		return false
 	}
-	_, _, ok := parsePartitionedSQSKey(key)
+	_, ok := stripPartitionedFamilyPrefix(key)
 	return ok
 }
 
--- a/adapter/sqs_partition_resolver.go
+++ b/adapter/sqs_partition_resolver.go
@@ -1,0 +1,134 @@
+package adapter
+
+import (
+	"bytes"
+	"encoding/binary"
+)
+
+// SQSPartitionResolver maps a partitioned-SQS key to the operator-
+// chosen Raft group for the (queue, partition) tuple. Implements
+// kv.PartitionResolver via duck typing — see the integration in
+// main.go where the resolver is installed on ShardedCoordinator.
+//
+// The byte-range engine cannot route partitioned queues because
+// adding per-partition routes would break its non-overlapping-cover
+// invariant (a partition route for partition K of one queue would
+// leave a gap for legacy keys that fall lexicographically between
+// partitions K and K+1). The resolver-first dispatch path avoids
+// this — it answers only for keys that match a partitioned family
+// prefix and otherwise lets the engine handle dispatch.
+type SQSPartitionResolver struct {
+	routes map[string][]uint64
+}
+
+// NewSQSPartitionResolver builds a resolver from the operator-
+// supplied partition map. routes[queue][k] is the Raft group ID
+// that owns partition k of queue, with len(routes[queue]) equal to
+// the queue's PartitionCount.
+//
+// Returns nil when routes is empty so callers can keep the resolver
+// out of the request path entirely on a non-partitioned cluster
+// (kv.ShardRouter.WithPartitionResolver(nil) is a documented no-op).
+//
+// The constructor takes a defensive copy so a later caller mutation
+// to the input map does not leak into the resolver's view at
+// runtime.
+func NewSQSPartitionResolver(routes map[string][]uint64) *SQSPartitionResolver {
+	if len(routes) == 0 {
+		return nil
+	}
+	cp := make(map[string][]uint64, len(routes))
+	for queue, groups := range routes {
+		ids := make([]uint64, len(groups))
+		copy(ids, groups)
+		cp[queue] = ids
+	}
+	return &SQSPartitionResolver{routes: cp}
+}
+
+// sqsResolverFamilyPrefixes is the set of partitioned-SQS family
+// prefixes ResolveGroup recognises. Kept package-internal so any
+// future renamed prefix touches both this list and the constant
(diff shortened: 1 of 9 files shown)


bootjp:
@claude review

Phase 3.D PR 4-B-2: routing-layer resolver for HT-FIFO. `htfifoCapabilityAdvertised` stays `false`; PR 4-B-3 will add the leadership-refusal hook + catalog-polling helper and flip the flag. PR description has the full self-review and test plan.

bootjp:
@claude review

Round 2 — addresses all four review findings:

1. **P1 (Codex) — resolver vs routeKey collapse**: `routeKey` collapses every `!sqs|...` key to `!sqs|route|global`, so the resolver never saw a partitioned-prefix key and the partition path was a no-op for Commit/Abort/Get. Fix: `ResolveGroup` now runs on the RAW key; the engine fallback keeps using `routeKey`-normalized input so legacy routing is unchanged. New `TestShardRouter_ResolverSeesRawKeyNotNormalized` uses a recording resolver to assert the raw key reaches it.

2. **HIGH (Gemini) — coordinator helpers bypass resolver**: `groupMutations`, `engineGroupIDForKey`, `groupForKey`, `routeAndGroupForKey` were calling `c.engine.GetRoute` directly — 2PC + read paths missed the resolver. Each now consults `c.router.ResolveGroup`. RouteID for keyviz observation is read from the engine separately (partition-aware heatmap is a follow-up).

3. **MEDIUM (Gemini, perf)** — `sqsResolverFamilyPrefixes` converted from `[]string` to `[][]byte` so the hot-path `bytes.HasPrefix` skips the per-call `[]byte(prefix)` allocation.

4. **MEDIUM (Gemini, doc)** — `WithPartitionResolver` comment corrected: interface assignment is not atomic in Go; the method is intended for startup wiring only.

`ResolveGroup` is exported so coordinator helpers and router internals share one dispatch path.

`go test -race ./kv/...` and `golangci-lint` clean.

bootjp:
@claude review

Round 3 — addresses Round 1 Claude's typed-nil interface bug:

**MEDIUM**: `buildSQSPartitionResolver` returned the concrete `*adapter.SQSPartitionResolver`. On a non-partitioned cluster the typed-nil pointer wrapped into a non-nil `kv.PartitionResolver` interface, so `s.partitionResolver != nil` was always true and every request paid an extra ResolveGroup call (safe via the nil-receiver guard but not free, contradicting the PR description's "hot path stays engine-only" claim).

Fix:
- Return type → `kv.PartitionResolver` interface so untyped `nil` propagates correctly.
- Defensive nil guard after `NewSQSPartitionResolver` for the empty-after-canonicalisation case.
- New `TestBuildSQSPartitionResolver_NilOnEmpty` uses `requireNilInterface` (a function-parameter forcing the interface conversion at the call boundary) to catch the regression — a plain `require.Nil` on the concrete pointer would pass even with the bug present, only the interface wrap exposes the failure.

`go test -race` and `golangci-lint` clean. Round 2's correctness fixes (raw-key resolver, coordinator helper plumbing, `[][]byte` perf, doc) all stand from the previous push.

bootjp:
@claude review

Round 4 — addresses Round 2 Claude's missing-regression-test note:

**CLAUDE.md gap**: the Gemini HIGH fix (ShardedCoordinator's helpers route through `c.router.ResolveGroup`) had no coordinator-level regression. Existing `TestShardRouter_*` tests pin the router-layer dispatch but don't exercise `Dispatch → groupMutations → c.router.ResolveGroup`.

New tests in `kv/sharded_coordinator_partition_test.go`:
- `TestShardedCoordinator_DispatchHonoursPartitionResolver`: engine routes everything to group 1, resolver claims a specific key for group 42; Dispatch on that key MUST hit group 42 (proves `groupMutations` no longer bypasses the resolver). Also asserts the resolver receives the RAW partitioned key — pins the codex-P1 fix at the coordinator boundary.
- `TestShardedCoordinator_DispatchFallsThroughForUnclaimedKeys`: keys NOT claimed by the resolver continue to route via the byte-range engine.

`stubResolver` is kv-internal so the tests don't pull in `adapter`. Race-safe defensive copy of each call key.

`go test -race ./kv/...` and `golangci-lint` clean.

Round 3's typed-nil fix (return type `kv.PartitionResolver`) and earlier rounds' fixes all stand from previous pushes.

bootjp:
@claude review

Round 5 — addresses Codex round 2 P1:

**P1 (Codex)**: when the resolver returns `(0, false)` for a partitioned-shaped key (unknown queue / OOR partition), the engine fall-through silently routed via routeKey's `!sqs|route|global` collapse to the SQS catalog default group. During partition-map drift or partial rollout this would mis-route HT-FIFO traffic instead of failing closed.

Fix:
- `kv.PartitionResolver` gains `RecognisesPartitionedKey(key) bool` — a shape-only predicate.
- `ShardRouter.ResolveGroup`: after `ResolveGroup` returns ok=false, consult `RecognisesPartitionedKey`. If the shape is recognised, fail closed; otherwise fall through to the engine as before.
- `adapter.SQSPartitionResolver` implements the predicate via `parsePartitionedSQSKey` (same parser as `ResolveGroup`).

Tests:
- `TestShardRouter_FailClosedOnRecognisedButUnresolved` — pins both branches.
- `TestSQSPartitionResolver_RecognisesPartitionedKey` — 8 shape cases (data/vis/byage families × known/unknown queues × OOR partition × legacy SQS × queue meta × non-SQS × empty × nil).
- Existing unknown-queue / OOR-partition tests renamed and now also assert `RecognisesPartitionedKey == true` to pin the router-side fail-closed contract.

`go test -race ./kv/...` and `golangci-lint` clean. All previous rounds' fixes (raw-key resolver, coordinator helper plumbing, [][]byte perf, doc, typed-nil interface, coordinator regression test) stand from earlier pushes.

bootjp:
@claude review

Round 6 — addresses Round 3 + Round 4 test gaps:

**Round 3 catch (real bug)**: `requireNilInterface` used `require.Nil` which testify reflects through to the underlying pointer — a nil pointer wrapped in a non-nil interface still passes. So if the typed-nil fix were reverted, the regression test would NOT catch it. Switched to `require.True(t, r == nil, msg)` which uses Go's native interface comparison (both type AND value must be nil tagged) — that is the exact invariant the typed-nil fix produces.

**Round 4 catch (real bug)**: `TestShardedCoordinator_DispatchHonoursPartitionResolver` does NOT regress the `groupMutations` bypass for single-mutation batches — `router.Commit`'s `groupRequests` re-routes by raw key and rescues the mis-routing.

Added `TestShardedCoordinator_DispatchSplitsMutationsByResolverGroup`: dispatches TWO mutations claimed by two different resolver groups (42, 43). Pre-fix, `groupMutations` bundles both under the engine default; `rawLogs` produces one request; `router.Commit` routes by `Mutations[0].Key` → group 43 receives zero. Post-fix: `groupMutations` splits via `c.router.ResolveGroup`, two separate requests, both groups receive theirs. Asserts both groups got exactly one request — only the post-fix path passes.

Updated the original test's comment to reflect what it actually pins (WithPartitionResolver wiring + raw-key dispatch, not groupMutations).

`go test -race ./kv/... ./adapter/...` and `golangci-lint` clean.

bootjp:
@claude review

Round 7 — addresses Round 5's malformed-queue-encoding nit:

**Round 5 nit**: `RecognisesPartitionedKey` delegated to `parsePartitionedSQSKey`, which fails when the queue segment is malformed base64 or the partition segment is truncated. For such corrupt-shape keys the predicate returned false → engine fall-through → silent mis-route via routeKey's `!sqs|route|global` collapse to the SQS catalog default group. That's exactly the failure mode the round 5 fail-closed change was introduced to prevent.

Fix: `RecognisesPartitionedKey` now does a prefix-only check (matches `stripPartitionedFamilyPrefix`), aligning with the kv.PartitionResolver contract's documented "purely on prefix / structural inspection" intent. ResolveGroup still returns (0, false) for malformed keys; the router pairs that with `Recognised=true` to fail closed.

New `TestSQSPartitionResolver_RecognisesMalformedPartitionedKey` (3 sub-cases): prefix-only, prefix + invalid base64, prefix + valid queue + truncated partition. All assert `Recognised=true` AND `ResolveGroup` ok=false.

`go test -race ./kv/... ./adapter/... .` and `golangci-lint` clean. All previous-round fixes from rounds 1–6 stand from earlier pushes.

bootjp:
@claude review

Round 8 — addresses Codex P1 missed in earlier rounds (見落としすみません):

**P1 (Codex)**: Round 5 introduced fail-closed semantics in `ShardRouter.ResolveGroup` but the read-key path in `ShardedCoordinator` wasn't audited for the new contract. `engineGroupIDForKey` returns 0 for any failure (recognised-but-unresolved or unrecognised), and `groupReadKeysByShardID` had `if gid == 0 { continue }` which silently dropped unrouted read keys. Result: a partitioned read key whose queue has drifted out of `--sqsFifoPartitionMap` gets dropped from OCC validation → FSM never sees it in ReadKeys → concurrent write to the same key commits without conflict → **SSI broken**.

Fix: `groupReadKeysByShardID` returns `(map, error)`. Any unroutable key surfaces as `ErrInvalidRequest`; `dispatchTxn` propagates and the transaction aborts before prewrite. Calls `c.router.ResolveGroup` directly to preserve the (gid, ok) signal through the boundary.

Tests:
- `TestGroupReadKeysByShardID_FailsClosedOnUnroutable` replaces the old `TestGroupReadKeysByShardID_SkipsUnroutableKeys` (which was pinning the **buggy** skip-silently behaviour).
- `TestShardedCoordinator_TxnFailsClosedForUnresolvedReadKey` (new): coordinator-level regression — a txn with a recognised-but-unresolved partitioned read key MUST abort before any prewrite. Asserts neither group received an RPC.

`go test -race ./kv/... ./adapter/...` and `golangci-lint` clean. All previous-round fixes from rounds 1–7 stand from earlier pushes.

PR #717, opened 2026-04-30
s3keys: export ParseBlobKey for offline blob-key consumers

bootjp:
## Summary

Tiny addition to `internal/s3keys`. Exports `ParseBlobKey` so the offline Phase 0a logical-backup decoder (next stacked PR) can route `!s3|blob|` records to their assembled object without holding a live cluster.

The package already exports parsers for object manifests and upload parts; blob keys are constructable via `BlobKey` / `VersionedBlobKey` but were not parseable from the outside.

## What `ParseBlobKey` does

Returns all components: `bucket`, `generation`, `object`, `uploadID`, `partNo`, `chunkNo`, `partVersion`. Both the 6-component (un-versioned) and 7-component (versioned) shapes are decoded. `partVersion=0` reverts to the un-versioned shape, matching `VersionedBlobKey`'s documented fallback. Truncated keys, malformed segments, and trailers that aren't either zero bytes or exactly one u64 are rejected with `ok=false`.

## Test plan

- [x] `go test -race ./internal/s3keys/...` — pass.
- [x] `golangci-lint run ./internal/s3keys/...` — clean.
- [x] 5 new tests: un-versioned round-trip, versioned round-trip, `partVersion=0` fallback, rejection of bucket-meta / object-manifest / upload-part / junk inputs, rejection of trailing-garbage keys.
- [x] Implementation split into `parseBlobKeyHead` + `parseOptionalPartVersion` so cyclomatic complexity stays under the package cap (no `//nolint`).

## Why this is its own PR

Branch stack discipline. The S3 backup encoder uses this; landing the parser separately keeps the encoder PR focused on the dump-format mapping.

### DIFF
--- a/internal/backup/dynamodb.go
+++ b/internal/backup/dynamodb.go
@@ -1,0 +1,482 @@
+package backup
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	pb "github.com/bootjp/elastickv/proto"
+	"github.com/cockroachdb/errors"
+	gproto "google.golang.org/protobuf/proto"
+)
+
+// Snapshot prefixes the DynamoDB encoder dispatches on. Mirror the live
+// constants in kv/shard_key.go (DynamoTableMetaPrefix etc.) so a renamed
+// prefix is caught by the dispatch tests below.
+const (
+	DDBTableMetaPrefix = "!ddb|meta|table|"
+	DDBTableGenPrefix  = "!ddb|meta|gen|"
+	DDBItemPrefix      = "!ddb|item|"
+	DDBGSIPrefix       = "!ddb|gsi|"
+)
+
+// Stored value magic prefixes. Mirror adapter/dynamodb_storage_codec.go:15-16.
+// Values that lack the right magic are rejected as either future-schema or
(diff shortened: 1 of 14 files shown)


bootjp:
PR #717 itself is `internal/s3keys/keys.go` + its tests (the `ParseBlobKey` addition). The 4 review comments above target code that lives in other stacked PRs:

- **Gemini #1** (`writeFileAtomic`, `openJSONL`, `stripPrefixSegment`, `mustMarshalIndent` should be centralised) — those are in PRs #713 (Redis) and #714 (SQS). The "centralise into `util.go`" suggestion is a structural cleanup that's better as a follow-up PR after the per-adapter stack lands; doing it now creates merge churn across every branch in the stack.
- **Gemini #383** (`IsBlobAtomicWriteRetriable`) — that function lives in PR #713; the dead-code / ENOSPC concern is addressed there in `945d73c9` (added `IsBlobAtomicWriteOutOfSpace`, kept retriable as the explicit short-write path).
- **Codex P1 #227** (DynamoDB generation filter) — addressed in PR #716 in `fdc9b36b`.
- **Codex P2 #285** (SQS side records when zero messages) — addressed in PR #714 in `3455ba7f`.

PR #717's actual code (s3keys.ParseBlobKey + 5 tests) has no outstanding feedback. Will land a follow-up `backup: centralise file/JSONL helpers into util.go` once the stack lands.

@claude review

PR #718, opened 2026-04-30
backup: S3 encoder for buckets, objects, and blob reassembly (Phase 0a)

bootjp:
## Summary

Stacked on top of #716 (DynamoDB) and #717 (s3keys.ParseBlobKey). Adds the S3 encoder for the Phase 0 logical-backup decoder — the most complex per-adapter piece because it must reassemble multipart object bodies from independent blob chunks.

**Snapshot prefixes handled:**
- `!s3|bucket|meta|<bucket>` → `s3/<bucket>/_bucket.json`
- `!s3|obj|head|<bucket><gen><object>` → `s3/<bucket>/<object>.elastickv-meta.json` (sidecar)
- `!s3|blob|<bucket><gen><object><uploadID><partNo><chunkNo>[<partVersion>]` → spilled to scratch on arrival; concatenated into `s3/<bucket>/<object>` at Finalize in `(partNo, partVersion, chunkNo)` order
- `!s3|upload|meta|`/`!s3|upload|part|`: excluded by default; opt in via `--include-incomplete-uploads` to emit under `_incomplete_uploads/records.jsonl`
- `!s3|bucket|gen|`, `!s3|gc|upload|`, `!s3route|`: ignored

## How body assembly works

Each blob chunk is written atomically to a per-(bucket, object) scratch directory as it arrives, registered under the `(uploadID, partNo, chunkNo, partVersion)` routing key. At Finalize, chunks for each object are sorted and concatenated into a single body file via tmp+rename. **Body bytes are never held in memory** — only the scratch path map. The scratch tree is removed on Finalize.

## Reserved-suffix collision handling

A user object key ending in `.elastickv-meta.json` is rejected with `ErrS3MetaSuffixCollision` by default. `WithRenameCollisions(true)` appends `.user-data` to the body file and records the rename in `s3/<bucket>/KEYMAP.jsonl` with `KindMetaCollision` so the dump remains reversible. Per-bucket `KEYMAP.jsonl` is opened lazily and dropped on Finalize if no records were written.

## Test plan

- [x] 11 tests covering bucket-meta round-trip, multipart out-of-order chunk reassembly, orphan-chunks warning, sidecar collision (rejected/renamed), JSON-decode rejection on both paths, ignored-prefix no-op, `--include-incomplete-uploads` round-trip + default skip, versioned blob assembly.

## Stacking

Base: `feat/backup-phase0a-dynamodb` (PR #716). Also depends on `feat/s3keys-parse-blob-key` (PR #717).

### DIFF
--- a/internal/backup/s3.go
+++ b/internal/backup/s3.go
@@ -57,11 +57,17 @@ var (
 )
 
 // verifyChunkCompleteness checks every (partNo, partVersion) entry in
-// declaredParts has the right number of chunkNo values present in
-// chunks. Chunks are expected at chunkNo in [0, chunk_count); a
-// missing index in that range surfaces as
+// declaredParts has exactly the set of chunkNo values {0, 1, …,
+// chunk_count-1} present in chunks. Chunks are expected at chunkNo in
+// [0, chunk_count); a missing index in that range surfaces as
 // ErrS3IncompleteBlobChunks rather than letting the assembler emit a
 // truncated body.
+//
+// We track the actual set of seen chunk indexes (not just count and
+// maxIndex) because count + maxIndex alone admits false positives:
+// for declared chunk_count=3, observed `{0, 2, 2}` produces count=3
+// and maxIndex=2 but is missing chunkNo=1, which would silently
+// assemble a corrupted body. Codex P1 round 12.
 //
 // declaredParts == nil means "no contract to verify" — used by tests
 // that pre-date the manifest-parts feature; production callers
@@ -70,31 +76,30 @@ func verifyChunkCompleteness(chunks []s3ChunkKey, declaredParts map[s3PartKey]s3
 	if declaredParts == nil {
 		return nil
 	}
-	// Count present chunks per (partNo, partVersion).
-	type observed struct {
-		count    uint64
-		maxIndex uint64
-	}
-	got := make(map[s3PartKey]observed, len(declaredParts))
+	got := make(map[s3PartKey]map[uint64]struct{}, len(declaredParts))
 	for _, k := range chunks {
 		pk := s3PartKey{partNo: k.partNo, partVersion: k.partVersion}
-		o := got[pk]
-		o.count++
-		if k.chunkNo > o.maxIndex {
-			o.maxIndex = k.chunkNo
+		if got[pk] == nil {
+			got[pk] = make(map[uint64]struct{})
 		}
-		got[pk] = o
+		got[pk][k.chunkNo] = struct{}{}
 	}
 	for pk, want := range declaredParts {
-		o := got[pk]
-		// We accept o.count == want.chunkCount AND
-		// o.maxIndex == chunkCount-1, because a snapshot with N
-		// duplicates of chunkNo=0 would satisfy the count check
-		// alone. Both the count and the highest index must match.
-		if want.chunkCount > 0 && (o.count != want.chunkCount || o.maxIndex+1 != want.chunkCount) {
+		if want.chunkCount == 0 {
+			continue
+		}
+		seen := got[pk]
+		if uint64(len(seen)) != want.chunkCount { //nolint:gosec // bounded
 			return errors.Wrapf(ErrS3IncompleteBlobChunks,
-				"partNo=%d partVersion=%d declared chunks=%d, observed count=%d maxIndex=%d",
-				pk.partNo, pk.partVersion, want.chunkCount, o.count, o.maxIndex)
+				"partNo=%d partVersion=%d declared chunks=%d, observed unique=%d",
+				pk.partNo, pk.partVersion, want.chunkCount, len(seen))
+		}
+		for i := uint64(0); i < want.chunkCount; i++ {
+			if _, ok := seen[i]; !ok {
+				return errors.Wrapf(ErrS3IncompleteBlobChunks,
+					"partNo=%d partVersion=%d declared chunks=%d, missing chunkNo=%d",
+					pk.partNo, pk.partVersion, want.chunkCount, i)
+			}
 		}
 	}
 	return nil
--- a/internal/backup/s3.go
+++ b/internal/backup/s3.go
@@ -147,8 +147,14 @@ type s3BucketState struct {
 	// every object flushes (the prior orphan-warning path covers it).
 	activeGen uint64
 	objects   map[string]*s3ObjectState // keyed by "object\x00generation"
-	keymap    *KeymapWriter
-	keymapDir string
+	// keymap / keymapFile / keymapDir are lazily set on the first
+	// rename. KeymapWriter.Close only flushes the bufio buffer, so
+	// the *os.File is tracked separately to be closed at finalize —
+	// otherwise a dump that produces keymaps for many buckets
+	// accumulates descriptors until EMFILE (Codex P1 round 9).
+	keymap     *KeymapWriter
+	keymapFile *os.File
+	keymapDir  string
 	// incompleteUploadsJL is opened lazily on the first
 	// !s3|upload|meta or !s3|upload|part record under
 	// --include-incomplete-uploads, then reused for every subsequent
@@ -564,18 +570,27 @@ func (s *S3Encoder) computeDirPrefixes(b *s3BucketState) map[string]bool {
 }
 
 // closeBucketKeymap closes the per-bucket KEYMAP.jsonl writer (if
-// opened) and removes the file when no rename was recorded.
+// opened) and removes the file when no rename was recorded. The
+// *os.File is closed separately because KeymapWriter.Close only
+// flushes its bufio buffer; without explicit fd close, dumps that
+// produce keymaps for many buckets leak descriptors until EMFILE
+// (Codex P1 round 9).
 func closeBucketKeymap(b *s3BucketState) error {
 	if b.keymap == nil {
 		return nil
 	}
-	if err := b.keymap.Close(); err != nil {
-		return err
+	flushErr := b.keymap.Close()
+	var closeErr error
+	if b.keymapFile != nil {
+		closeErr = b.keymapFile.Close()
+	}
+	if flushErr == nil && closeErr != nil {
+		flushErr = errors.WithStack(closeErr)
 	}
 	if b.keymap.Count() == 0 && b.keymapDir != "" {
 		_ = os.Remove(filepath.Join(b.keymapDir, "KEYMAP.jsonl"))
 	}
-	return nil
+	return flushErr
 }
 
 func (s *S3Encoder) flushObjectWithCollision(b *s3BucketState, bucketDir string, obj *s3ObjectState, needsLeafDataRename bool) error {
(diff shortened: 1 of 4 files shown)


bootjp:
3 critical correctness bugs + 3 quality issues addressed in `19ae3289`. All driven by reviewer findings.

**Critical**

**Codex P1 #425 — Path traversal.** A user-supplied key like `../../../etc/passwd` would have escaped the bucket dir and overwritten host files. Added `safeJoinUnderRoot` which `filepath.Clean`s the joined path and asserts it stays rooted under the bucket dir; `..` traversal returns `ErrS3MalformedKey`. Absolute paths (leading `/`) collapse safely under the bucket dir via `filepath.Join`, which I assert with `TestS3_AbsolutePathObjectKeyConfinedUnderBucket`.

**Codex P1 #500 + Gemini HIGH #106/#476/#504 — Stale upload-id chunks.** A snapshot mid-delete-and-recreate or mid-retry can carry blob chunks for multiple upload attempts under the same `(bucket, gen, object)`. The prior `assembleObjectBody` concatenated every chunk regardless of `upload_id` → corrupted bytes. `s3ObjectState.uploadID` is now set by `HandleObjectManifest`; new `filterChunksForManifest` returns only matching chunks. `TestS3_StaleUploadIDChunksFilteredFromAssembledBody`.

**Codex P2 #318 + Gemini HIGH+MEDIUM #318 — Truncated `records.jsonl`.** `HandleIncompleteUpload` reopened the file on every call (`os.Create` → truncate); only the last record survived. Now `s3BucketState.incompleteUploadsJL` is opened lazily once per bucket, cached, closed in `flushBucket` (with error surfaced). `TestS3_IncompleteUploadsAppendsAcrossCalls`.

**Quality**

**Gemini MEDIUM #285 — `MkdirAll` per blob.** `s3ObjectState.scratchDirCreated` makes the call once per object.

**Gemini MEDIUM #318 — closeJSONL error.** `closeBucketKeymap` and the `incompleteUploadsJL` close path now both surface the error.

**Gemini MEDIUM #386 — `includeOrphans` unimplemented.** Orphan chunks (chunks present without matching manifest) now write to `<bucket>/_orphans/<encoded-object>/<chunk>.bin` under `WithIncludeOrphans(true)`. `TestS3_OrphanChunksWrittenWhenIncludeOrphans`.

Removed the now-unused `attachManifestParts` placeholder; `sortChunkKeys`'s logic moved into `filterChunksForManifest`.

@claude review

bootjp:
3 Codex follow-ups addressed in `92ee22ef`:

**P1 #619 — Assemble only manifest-declared parts.** `filterChunksForManifest` previously matched on uploadID only; the manifest declares specific `(partNo, partVersion)` tuples, and S3's overwrite-then-async-cleanup window can leave older `partVersion` chunks present. Mixing them produced corrupted bytes. `s3ObjectState.declaredParts` is now populated from `manifest.parts` and used as a strict allow-list. `TestS3_StalePartVersionExcludedFromAssembledBody`.

**P2 #497 — Preserve dot segments.** `safeJoinUnderRoot` used `filepath.Clean` which collapsed `a/../b` to `b`. S3 treats those bytes literally; two distinct keys would have silently merged. Now rejects any object key whose path segments are `.` or `..` with `ErrS3MalformedKey`. NUL bytes also rejected. `TestS3_DotSegmentObjectKeyRejected`.

**P2 #521 — Cross-generation collision.** `s3BucketState.activeGen` captures the bucket-meta record's generation. Objects from older incarnations of the same bucket name are now suppressed (with `s3_stale_generation_objects` warning) or routed to `_orphans/` under `--include-orphans`. `TestS3_StaleGenerationObjectExcluded`.

`flushBucket` complexity kept under the cap by extracting `flushBucketObjects`.

@claude review

bootjp:
**Codex P1 #615** addressed in `2c44292f`. When a bucket holds both `path/to` and `path/to/sub` the natural-path mapping requires `path/to` to be a regular file AND a directory; POSIX can't, so the prior code failed with EEXIST/ENOTDIR at finalize.

`flushBucketObjects` now pre-computes the set of "directory prefixes" required by the union of active-gen object keys (object `a/b/c` contributes `a` and `a/b`). Each object whose own key is a member of that set is flagged for the documented leaf-data rename: `path/to` → `path/to.elastickv-leaf-data` + KEYMAP.jsonl record with `KindS3LeafData`. The longer key flushes at its natural path unchanged.

`TestS3_FileVsDirectoryKeyCollisionGetsLeafDataRename` covers the round-trip end-to-end (both files exist, KEYMAP record has `KindS3LeafData` and original `path/to`).

@claude review

bootjp:
2 Codex P1 follow-ups addressed in `ba33df8e`:

**#729 — Chunk completeness.** `assembleObjectBody` previously appended whatever chunk files happened to exist; a partial/racy/corrupted snapshot would silently emit a truncated body. New `verifyChunkCompleteness` counts chunks per `(partNo, partVersion)` and asserts both the count equals manifest's `chunk_count` AND `maxChunkNo + 1 == chunk_count`. Mismatch returns `ErrS3IncompleteBlobChunks` before any bytes hit disk. `declaredParts` value type changed from `struct{}` to `s3DeclaredPart{chunkCount}`. Test `TestS3_IncompleteChunksRejected` (manifest declares 3 chunks, snapshot has chunkNo 0 + 2; chunk 1 missing → error).

**#614 — Empty slash segments.** `safeJoinUnderRoot` rejected `.` and `..` but allowed `""` segments ("a//b", "a/", trailing "/"). `filepath.Join` collapses these; distinct S3 keys would silently overwrite each other. Now rejects `""` segments anywhere except the leading position (where the leading "/" → initial empty segment is handled safely by the already-tested absolute-path-confined-under-bucket behaviour). Test `TestS3_EmptySlashSegmentsRejected` covers "a//b", "a/", "/a//", "x/".

@claude review

bootjp:
@claude review

Round 5 fix for Codex P1 on commit 2f87b843:

`safeJoinUnderRoot` previously permitted an empty first segment, so a leading-slash S3 key like `/a` would be accepted and then collapsed by `filepath.Join` onto the same output path as `a`. S3 treats `/a` and `a` as distinct keys (the literal byte `/` is part of the key), so a bucket containing both produced last-flush-wins corruption with no KEYMAP record.

Now empty segments are rejected everywhere — leading, mid-path, and trailing — with `ErrS3MalformedKey`. The previous test `TestS3_AbsolutePathObjectKeyConfinedUnderBucket` (which asserted the wrong-by-design "confine and merge" behaviour) is replaced by `TestS3_LeadingSlashObjectKeyRejected`. Operators with such keys must rename them in S3 first.

`go test -race ./internal/backup/...` and `golangci-lint run` clean.

bootjp:
@claude review

Round 6 fix for Codex P1 on commit 09c2a0e0:

`safeJoinUnderRoot` now refuses object keys containing backslashes. The dot-segment scan splits on `/` only, but `filepath.Join` treats `\` as a separator on Windows; a key like `a\..\b` would bypass the dot-segment check (where `\` is a literal byte) and then collapse to `b` on Windows, silently merging two distinct S3 keys. Dumps must produce identical output regardless of host OS, so backslashes are refused on every platform with `ErrS3MalformedKey`. Operators with such keys must rename them in S3 first.

Test: `TestS3_BackslashObjectKeyRejected` covers the dot-segment escape and leading/trailing/embedded backslashes. `go test -race` and `golangci-lint` clean.

bootjp:
@claude review

Round 7 fixes for Codex round 9 on commit ab38eb0a (commit `46cb56fe`):

1. **P1 — KEYMAP file descriptor leak.** `recordKeymap` stored only the `*KeymapWriter`; `closeBucketKeymap` called `KeymapWriter.Close()` which flushes the bufio buffer but does NOT close the underlying `*os.File`. A dump producing keymaps for many buckets accumulated descriptors until `EMFILE` and subsequent flushes failed. Now `s3BucketState` tracks the `*os.File` and `closeBucketKeymap` closes it explicitly.

2. **P2 — KEYMAP open followed symlinks.** `recordKeymap` was using `os.Create` directly, which follows symlinks and clobbers hard links. The redis encoder already routes through `openSidecarFile` (atomic O_NOFOLLOW + nlink check); s3 now uses the same helper. Test: `TestS3_KeymapRefusesSymlinkAtFinalize` pre-places a symlink at the KEYMAP path and asserts the bait file is untouched.

`go test -race` and `golangci-lint` clean.

bootjp:
@claude review

Round 8 fixes for Codex round 9 on commit 1dc68842 (commit `90d33feb`):

1. **P1 — Leaf-data rename collided with real user keys.** `needsLeafDataRename` rewrote `<object>` to `<object>.elastickv-leaf-data` without checking whether a real key with that suffix already existed in the same bucket. Two distinct objects mapped to one filesystem path → finalize was last-flush-wins. `resolveObjectFilename` now consults a per-bucket active-gen object-key set via `computeActiveGenObjectKeys`; rename targets that collide surface `ErrS3MetaSuffixCollision`. Test: `TestS3_LeafDataRenameRejectsCollidingUserKey`.

2. **P1 — Meta-suffix rename collided too.** Same fix path: rename-collisions mode rewrites `<object>.elastickv-meta.json` to `<object>.elastickv-meta.json.user-data`; the same set lookup now refuses when that suffixed key is itself a real object. Test: `TestS3_MetaSuffixRenameRejectsCollidingUserKey`.

3. **P2 — `last_modified` never populated.** The decoded `s3LiveManifest` carried `last_modified_hlc`, but `s3PublicManifest` silently dropped it. Added `formatHLCAsRFC3339Nano` which extracts the millisecond half (HLC >> 16, see `kv/hlc.go`) and renders RFC3339Nano UTC. HLC zero maps to "" so omitempty drops the field. Test: `TestS3_LastModifiedSidecarPopulated`.

`go test -race` and `golangci-lint` clean.

bootjp:
@claude review

Round 9 fix for Codex P1 round 11 on commit 9a63e326 (commit `d016ea71`):

`HandleBlob` composed scratch paths with `filepath.Join(scratchRoot, EncodeSegment(bucket), EncodeSegment(object))`. EncodeSegment uses the RFC3986 unreserved set (`ALPHA/DIGIT/-/./_`) — `/` is percent-encoded but `.` is preserved, so the literal segment `..` survives unchanged. A crafted `bucket=".."` and/or `object=".."` would resolve to `<scratchRoot>/../...`, letting `writeFileAtomic` land outside the decoder's controlled directory before `safeJoinUnderRoot` ever ran at output time.

Added `scratchDirForBlob` which rejects `.` / `..` / "" bucket and object literals at the encoder boundary so the spill-to-disk step inherits the same containment invariant the final output path enforces. Same guard applied to `flushOrphanObject` (which shares the failure mode under `--include-orphans`). Multi-segment dot keys like `a/../b` continue to be caught at Finalize via `safeJoinUnderRoot` (EncodeSegment keeps the whole key as one filename segment).

Tests: `TestS3_HandleBlobRejectsScratchEscape` (5 sub-cases covering bucket/object/both variants of `.`/`..`); `TestS3_DotSegmentObjectKeyRejected` updated to accept either HandleBlob or Finalize as the rejection point. `go test -race` and `golangci-lint` clean.

bootjp:
@claude review

Round 10 fix for Codex P1 round 12 on commit e91f0866 (commit `4505df35`):

`verifyChunkCompleteness` only checked `count` and `maxIndex` per (partNo, partVersion). For declared `chunk_count=3`, observed chunk numbers `{0, 2, 2}` satisfied both thresholds (`count==3`, `maxIndex+1==3`) while `chunk_no=1` was absent — `assembleObjectBody` would then emit a corrupted body silently.

Replaced the dual-threshold heuristic with a set-membership check: track every observed chunk index per (partNo, partVersion) and verify the set is exactly `{0, 1, …, chunk_count-1}`. Both the unique-count guard and a per-index lookup surface `ErrS3IncompleteBlobChunks` before the assembler runs.

Test: `TestS3_DuplicateChunksWithMissingIndexRejected` drives the `{0, 2, 2}` shape; existing `TestS3_IncompleteChunksRejected` still passes against the new code path. `go test -race` and `golangci-lint` clean.

PR #720, opened 2026-04-30
feat(keyviz): leader-term wire format + main.go ticker (Phase 2-C+ PR-3b)

bootjp:
## Summary

Build the end-to-end pipeline that takes the sampler-side `RaftGroupID + LeaderTerm` fields shipped in PR #709 (Phase 2-C+ PR-3a) and surfaces them on the wire (proto + JSON + gRPC) so the fan-out aggregator (PR-3c, next) has data to dedupe by `(group, term)` instead of the conservative §4.2 max-merge.

## Why

Per `docs/design/2026_04_27_proposed_keyviz_cluster_fanout.md` §4.2, the cluster fan-out today max-merges write samples per row — under stable leadership this is exact, but under a leadership flip it understates the true window total by at most one leader's pre-transfer slice. §9.1's canonical fix is to dedupe by `(bucket_id, raft_group_id, leader_term, column)` and sum across terms; that needs `raft_group_id + leader_term` on every row of the wire format. PR #709 added the fields to `MatrixRow`; this PR plumbs them through.

## Changes

**Wire-format additions (forward-compatible):**

- `proto/admin.proto`: `KeyVizRow.raft_group_id = 13` and `leader_term = 14`. Old SPAs ignore the new fields; old servers don't emit them — the zero value is the documented "term not tracked" fallback that triggers the legacy max-merge.
- `internal/admin/keyviz_handler.go`: JSON `KeyVizRow` gains `RaftGroupID` and `LeaderTerm` (omitempty); `newKeyVizRowFrom` copies `mr.RaftGroupID` / `mr.LeaderTerm`.
- `internal/admin/keyviz_fanout.go`: `mergeRowInto` seeds the destination row with `RaftGroupID + LeaderTerm` on first-seen (PR-3c will switch to a per-cell merge).
- `adapter/admin_grpc.go`: `newKeyVizRowFrom` copies the same two fields onto the proto `KeyVizRow`.

**main.go ticker:**

- `startKeyVizLeaderTermPublisher` launches a goroutine that polls each `runtime.engine.Status().Term` every `Step` (matching the flush cadence so each column observes a fresh term). Skips entirely when the sampler is nil or no runtimes are wired.
- `publishLeaderTerms` / `publishLeaderTermsFromSnapshots` split into two helpers so unit tests can exercise publication without standing up a full `raftengine.Engine` fake.
- The ticker publishes once immediately so the very first flush column carries a non-zero term.

## Scope (PR-3b only)

- Wire format **emission**: ✓ this PR
- Ticker that **publishes** terms via `SetLeaderTerm`: ✓ this PR
- Aggregator **consumption** (per-cell `(group, term)` merge replacing max-merge): **deferred to PR-3c**
- `Conflict` per-cell promotion: **deferred to PR-3c**
- SPA changes: **none required** — the wire-format extension is forward-compatible

With this PR alone, the new fields are visible on every response but the aggregator still uses §4.2 max-merge → no behavior change for SPA consumers.

## Test plan

- [x] `go test -race -count=1 ./keyviz/... ./internal/admin/...` — passes
- [x] New test coverage:
  - `TestKeyVizHandlerStampsRaftIdentity` (JSON path)
  - `TestMergeKeyVizMatricesPreservesRaftIdentity` (fan-out merge)
  - `TestGetKeyVizMatrixStampsRaftIdentity` (gRPC path)
  - `TestPublishLeaderTermsFromSnapshotsStampsRows` (end-to-end through sampler)
  - `TestPublishLeaderTermsFromSnapshotsNilSafe` (nil-receiver contract)
  - `TestStartKeyVizLeaderTermPublisherSkipsWhenSamplerNil / ...WhenNoRuntimes` (errgroup lifecycle)
- [x] `golangci-lint run` — 0 issues
- [ ] Jepsen — N/A (admin-side wire extension, no replication/MVCC/OCC impact)

## Five-lens self-review

1. **Data loss** — wire-format extension only; legacy max-merge fallback preserves today's behavior when `LeaderTerm=0`.
2. **Concurrency** — ticker is a single goroutine; `SetLeaderTerm` takes `groupTermsMu` (Phase 2-C+ PR-3a contract); no new shared state.
3. **Performance** — `Observe` hot path is unchanged; ticker adds `N × Status()` calls per `Step` where `N` is the runtime count (typically ≤ a few groups), and `Status()` is a non-blocking read.
4. **Data consistency** — `RaftGroupID + LeaderTerm` round-trip through every wire path verified by table-driven tests.
5. **Test coverage** — each layer (JSON, fan-out, gRPC, publisher) has its own focused regression guard; end-to-end test pins the sampler→column propagation.

## Summary by CodeRabbit

* **New Features**
  * Keyviz matrix now includes Raft leadership metadata with group ID and leader term fields in responses.
  * Leadership term information is automatically tracked and included in key visualization output.

### DIFF
--- a/adapter/admin_grpc.go
+++ b/adapter/admin_grpc.go
@@ -612,6 +612,12 @@ func matrixToProto(cols []keyviz.MatrixColumn, pick func(keyviz.MatrixRow) uint6
 				order = append(order, mr.RouteID)
 			}
 			pr.Values[j] = pick(mr)
+			// Per-cell Raft identity stamped from this column's
+			// MatrixRow (Gemini HIGH on PR #720 — a row-level
+			// scalar would only capture the first column's
+			// identity and break the per-cell dedupe goal).
+			pr.RaftGroupIds[j] = mr.RaftGroupID
+			pr.LeaderTerms[j] = mr.LeaderTerm
 		}
 	}
 	resp.Rows = make([]*pb.KeyVizRow, len(order))
@@ -675,9 +681,12 @@ func newKeyVizRowFrom(mr keyviz.MatrixRow, numCols int) *pb.KeyVizRow {
 		Aggregate:         mr.Aggregate,
 		RouteCount:        total,
 		RouteIdsTruncated: mr.Aggregate && total > uint64(len(mr.MemberRoutes)),
-		RaftGroupId:       mr.RaftGroupID,
-		LeaderTerm:        mr.LeaderTerm,
 		Values:            make([]uint64, numCols),
+		// Per-cell parallel arrays — populated in the column-loop
+		// above, not here, because the first MatrixRow seen for a
+		// route only carries that one column's identity.
+		RaftGroupIds: make([]uint64, numCols),
+		LeaderTerms:  make([]uint64, numCols),
 	}
 	if mr.Aggregate {
 		row.RouteIds = append([]uint64(nil), mr.MemberRoutes...)
--- a/adapter/admin_grpc.go
+++ b/adapter/admin_grpc.go
@@ -675,6 +675,8 @@ func newKeyVizRowFrom(mr keyviz.MatrixRow, numCols int) *pb.KeyVizRow {
 		Aggregate:         mr.Aggregate,
 		RouteCount:        total,
 		RouteIdsTruncated: mr.Aggregate && total > uint64(len(mr.MemberRoutes)),
+		RaftGroupId:       mr.RaftGroupID,
+		LeaderTerm:        mr.LeaderTerm,
 		Values:            make([]uint64, numCols),
 	}
 	if mr.Aggregate {
--- a/adapter/admin_grpc_keyviz_test.go
+++ b/adapter/admin_grpc_keyviz_test.go
@@ -274,20 +274,31 @@ func TestGetKeyVizMatrixHonorsRowsBudget(t *testing.T) {
 	require.Equal(t, "route:4", resp.Rows[1].BucketId)
 }
 
-// TestGetKeyVizMatrixStampsRaftIdentity pins the Phase 2-C+ wire
-// extension: MatrixRow.RaftGroupID and MatrixRow.LeaderTerm propagate
-// through matrixToProto into the proto KeyVizRow's
-// raft_group_id (field 13) and leader_term (field 14). The fan-out
-// aggregator's per-term dedupe key requires both fields on the wire.
+// TestGetKeyVizMatrixStampsRaftIdentity pins the Phase 2-C+ per-cell
+// wire extension: MatrixRow.RaftGroupID and MatrixRow.LeaderTerm
+// propagate per-column through matrixToProto into the proto
+// KeyVizRow's raft_group_ids (field 13) and leader_terms (field 14)
+// parallel-to-values arrays. The fan-out aggregator's per-cell
+// dedupe key requires per-column identity (Gemini HIGH on PR #720
+// — row-level scalars only captured the first column's identity
+// and broke dedupe under mid-window leader flips).
 func TestGetKeyVizMatrixStampsRaftIdentity(t *testing.T) {
 	t.Parallel()
 	t0 := time.Unix(1_700_000_000, 0)
+	t1 := t0.Add(time.Minute)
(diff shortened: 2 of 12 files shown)


bootjp:
@claude review

Round-1 follow-ups (`d7e5c43a`) addressing 3 review items:

- **Gemini HIGH (proto/admin.proto:116)** + **Gemini MEDIUM (keyviz_fanout.go:529)** — both resolved by switching the wire format from row-level scalars to **per-cell parallel arrays** (`repeated uint64 raft_group_ids = 13; repeated uint64 leader_terms = 14;` parallel to `values[]`). The pivot loops in `adapter/admin_grpc.go` and `internal/admin/keyviz_handler.go` now stamp `row.RaftGroupIDs[j] = mr.RaftGroupID` per column. `mergeRowInto` stamps the destination cell's identity from whichever source contributed the value kept at that cell — for `maxMerge` that's the larger source; for `sumMerge` it's best-effort last-touched (sum has no single owner; steady state has agreeing terms).

- **Codex P2 (main.go:1565)** — `publishLeaderTerms` snapshotted `rt.engine` into a local once before nil-check + `Status()` call, eliminating the race against `rt.Close()` setting `rt.engine = nil` during startup-error teardown.

**Caller audit** (per loop instructions, semantic change requires audit): the only production consumers of `KeyVizRow.RaftGroupID/LeaderTerm` were the pivot/merge paths I'm updating. `grep -rn 'KeyVizRow.*\\.RaftGroupId\\b'` returns nothing outside `proto/admin.pb.go` (generated). No external consumer reads these fields. The PR-3c aggregator that will USE the per-cell dedupe hasn't been written yet, so there are no callers expecting the old scalar shape.

**New test:** `TestMergeKeyVizMatricesPerCellIdentityMatchesValueOwner` drives a leadership flip across two cells with each leader winning one cell, asserts merged identity at each cell matches the value's owner (Gemini MEDIUM regression guard). Existing `TestKeyVizHandlerStampsRaftIdentity` and `TestGetKeyVizMatrixStampsRaftIdentity` extended to two-column scenarios with mid-window flips.

`go test -race -count=1 ./keyviz/... ./internal/admin/...` — pass; `golangci-lint run` (full project) — 0 issues.

bootjp:
@claude review

Round-2 follow-ups (`c3f0ce66`) addressing 2 Codex P2 items on `d7e5c43a`:

- **Codex P2 #1 (keyviz_fanout.go) — fail-closed for legacy peer winners**: previous `j < len(row.RaftGroupIDs)` guard SKIPPED assignment when the incoming had empty arrays, leaving stale identity from a previous higher-versioned source. **Fixed**: explicitly reset to 0 (the documented "term not tracked" sentinel) when the incoming wins but its metadata array is shorter than j.

- **Codex P2 #2 (main.go publishLeaderTerms) — race with rt.Close()**: even after round-1's local snapshot, the initial read of `rt.engine` was unsynchronized with `Close()`'s write. **Fixed structurally** by adding `engineMu sync.RWMutex` to `raftGroupRuntime`, exposing `snapshotEngine()` accessor (read lock) and updating `Close()` to clear the field under the write lock.

**Caller audit** (per loop instruction — both fixes are semantic changes):

- `mergeRowInto` reset semantics: only caller is `mergeKeyVizMatrices` (same file). No external consumer of the stale-identity behavior.
- `raftGroupRuntime.engine` field access: grepped all `rt.engine` / `r.engine` reads:
  - `main.go:858, 1104, 1108, 1115` — startup wiring, single-threaded BEFORE goroutines launch. **Safe to keep direct field reads.**
  - `main_admin.go:828, 831, 837` (`newClusterInfoSource`) — HTTP/gRPC handler closure, runs concurrent with `rt.Close()`. **Same race class as the publisher.** Migrated to `snapshotEngine()` for consistency.
  - `multiraft_runtime.go:Close()` — now takes `engineMu.Lock()` around clearing the field.

**New test**: `TestMergeKeyVizMatricesLegacyPeerWinnerResetsIdentity` drives mixed-version cluster (modern reports 30 with identity, legacy reports 50 with empty arrays). Legacy wins maxMerge → asserts `dst.RaftGroupIDs/LeaderTerms` reset to 0 instead of inheriting modern's identity.

`go test -race -count=1 ./keyviz/... ./internal/admin/...` — pass; `golangci-lint run` (full project) — 0 issues.

PR #721, opened 2026-04-30
feat(sqs): htfifo capability poller (Phase 3.D PR 4-B-3a)

bootjp:
## Summary

Phase 3.D PR 4-B-3a — adds the stateless `htfifo` capability poller that PR 5's CreateQueue gate consumes. Stacks on the now-merged #715 (PR 4-B-2, partition resolver). This PR is purely additive: new helper file, new test file, no existing code touched. Next is PR 4-B-3b (leadership-refusal hook + flag flip).

## What's added

- `adapter/sqs_capability_poller.go`:
  - `HTFIFOCapabilityReport{AllAdvertise, Peers}` — binary go/no-go signal + per-peer detail for operator triage.
  - `HTFIFOCapabilityPeerStatus{Address, HasHTFIFO, Capabilities, Error}` — one peer's polling result.
  - `PollSQSHTFIFOCapability(ctx, client, peers)` — concurrent goroutine-per-peer poll, indexed-channel result aggregation (race-free).
  - Per-peer timeout `defaultSQSCapabilityPollTimeout = 3s` so a single hung peer can't stall the cluster-wide poll.
  - Body capped at 1 KiB via `io.LimitReader` so a misconfigured peer can't drain memory.
  - Bare `host:port` and full `http://…` / `https://…` URLs both accepted.
  - Fail-closed on every failure mode: timeout, transport error, non-200, malformed JSON, missing capability. Empty peer list → vacuously `AllAdvertise=true` (caller validates list completeness).

## What's NOT added (deferred)

- `htfifoCapabilityAdvertised` stays `false`. PR 4-B-3b adds the §8 leadership-refusal hook + per-acquisition observer in `kv/raftengine/etcd` and flips the flag.
- `CreateQueue` does NOT yet call this helper. PR 5 lifts the `PartitionCount > 1` dormancy gate AND wires the capability check in the same commit (per the §11 rollout plan's "gate-and-lift atomically" rule).

## Test plan

9 top-level tests covering the contract surface:
- [x] `TestPollSQSHTFIFOCapability_AllAdvertise` — happy path, multiple peers.
- [x] `TestPollSQSHTFIFOCapability_OneMissingFailsClosed` — old-binary peer with empty capabilities drops `AllAdvertise`.
- [x] `TestPollSQSHTFIFOCapability_HTTPErrorFailsClosed` — HTTP 500, connection refused, malformed JSON all surface as `Error`.
- [x] `TestPollSQSHTFIFOCapability_TimeoutFailsClosed` — hung peer respects per-peer timeout, full poll bounded.
- [x] `TestPollSQSHTFIFOCapability_EmptyPeersIsVacuouslyTrue` — empty peer list contract.
- [x] `TestPollSQSHTFIFOCapability_EmptyPeerAddressFailsClosed` — `""` entry in peers slice surfaces explicit Error.
- [x] `TestPollSQSHTFIFOCapability_FullURLPeer` — `http://` and `https://` URLs accepted alongside bare `host:port`.
- [x] `TestPollSQSHTFIFOCapability_ConcurrentPolling` — 5×200ms peers finish in well under 1s.
- [x] `TestPollSQSHTFIFOCapability_RespectsBodyLimit` — 10 KiB response truncated mid-string surfaces as JSON parse error, not garbage decode.
- [x] `TestBuildSQSHealthURL` — URL construction edge cases.
- [x] `go test -race ./adapter/...` pass.
- [x] `golangci-lint ./adapter/...` clean.

1. **Data loss** — read-only HTTP poll; no FSM/Pebble/retention path. No issue.
2. **Concurrency / distributed failures** — peer polls run in independent goroutines; results land via an indexed channel so slice writes are obviously race-free. Per-peer timeout enforced via `context.WithTimeout` so a slow peer can't stall the rest. Body capped via `io.LimitReader`. No issue.
3. **Performance** — N peers polled concurrently, not serially; the test pins this. Per-peer cost is one HTTP round-trip + a JSON parse of a tiny body. No hot-path impact (CreateQueue is a control-plane operation, not request hot path). No issue.
4. **Data consistency** — fail-closed on every failure mode preserves the §8.5 "any peer that doesn't respond is treated as not-yet-upgraded" rule. The vacuously-true empty-peer-list case is documented and the caller's responsibility. No issue.
5. **Test coverage** — every documented failure path (HTTP error, transport error, JSON parse, timeout, missing capability, empty peer, body-size cap) is pinned. Concurrent polling is pinned (would have caught a regression to serial). URL construction edges pinned.

### DIFF
--- a/adapter/sqs_capability_poller.go
+++ b/adapter/sqs_capability_poller.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/http"
 	"strings"
 	"sync"
@@ -63,12 +64,34 @@ type HTFIFOCapabilityPeerStatus struct {
 }
 
 // defaultSQSCapabilityPollTimeout caps how long the poller waits on
-// any single peer. The §8.5 design's "fail-closed default for
-// nodes that don't respond within a short timeout" turns into a
-// concrete bound here. Operators wanting a longer wait can pass
-// their own context with a deadline; the per-peer cap is enforced
-// in addition so a single slow peer cannot stall the whole poll.
+// any single peer when PollerConfig.PerPeerTimeout is zero. The
+// §8.5 design's "fail-closed default for nodes that don't respond
+// within a short timeout" turns into a concrete bound here.
+// Operators wanting a longer or shorter wait can override via
+// PollerConfig; the cap is enforced in addition to any
+// caller-supplied context deadline so a single slow peer cannot
+// stall the whole poll.
 const defaultSQSCapabilityPollTimeout = 3 * time.Second
+
+// PollerConfig tunes PollSQSHTFIFOCapability for a specific call
+// site. All fields are optional — the zero value picks safe
+// defaults. Tests use the explicit PerPeerTimeout to exercise the
+// per-peer cap independently of any caller-supplied context
+// deadline.
+type PollerConfig struct {
+	// HTTPClient is the client used for /sqs_health GETs. Nil
+	// falls back to http.DefaultClient. Callers wanting connection
+	// pooling, custom Transport, or shorter Client.Timeout pass
+	// their own.
+	HTTPClient *http.Client
+
+	// PerPeerTimeout caps how long any single peer's poll runs
+	// before being abandoned. Zero defaults to
+	// defaultSQSCapabilityPollTimeout (3s). Tests pass a small
+	// value (e.g. 100ms) so the per-peer cap path can be
+	// exercised quickly without a parent context deadline.
+	PerPeerTimeout time.Duration
+}
 
 // PollSQSHTFIFOCapability polls each peer's /sqs_health endpoint
 // concurrently and reports whether all advertise htfifo. The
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

Phase 3.D PR 4-B-3a — htfifo capability poller. Pure-additive helper that PR 5's CreateQueue gate will consume; `htfifoCapabilityAdvertised` stays false (PR 4-B-3b will flip it after the leadership-refusal hook lands). PR description has the full self-review and test plan.

bootjp:
@claude review

Round 2 — addresses Round 1 review:

**1. Gemini medium — Body.Close error silently ignored**: now logs via `slog.Warn` with peer address. A failed Close means the http.Transport will tear down rather than reuse the connection — silently dropping it could mask leaking connections under load.

**2. Claude minor — non-200 body not drained**: `io.Copy(io.Discard, ...)` before the early return so the transport can reuse the connection. Drain is wrapped in the same `io.LimitReader` cap so a misbehaving peer can't drain memory through the discard path.

**3. Claude nit — per-peer cap not exercised independently**: refactored `PollSQSHTFIFOCapability` signature to `(ctx, peers, cfg PollerConfig)` where `PollerConfig.PerPeerTimeout` is a tunable. Tests now have:
- `TestPollSQSHTFIFOCapability_ParentContextDeadlineFailsClosed` (renamed from `TimeoutFailsClosed`) — pins parent-ctx deadline path
- `TestPollSQSHTFIFOCapability_PerPeerTimeoutFailsClosed` (new) — uses `context.Background()` + `PerPeerTimeout=100ms` to exercise the cap **independently** of any parent deadline

Bonus: `PerPeerTimeout` is also a useful operator knob — different cluster latencies want different bounds. **Audit per lessons-learned**: `grep -rn PollSQSHTFIFOCapability` confirmed only the test file references the function — API change is safe.

**4. Claude minor — `buildSQSHealthURL` double-path edge case**: added an explicit test case documenting the behavior (caller must not pass a URL that already includes `/sqs_health`).

`go test -race ./adapter/...` and `golangci-lint` clean.

PR #723, opened 2026-04-30
feat(sqs): leadership-refusal hook + flag flip (Phase 3.D PR 4-B-3b)

bootjp:
## Summary

Phase 3.D PR 4-B-3b — closes the routing+leadership-refusal pair the §11 PR 4 contract requires before a binary is "marked htfifo-eligible". With this PR landed, every node's `/sqs_health` advertises `htfifo` and the §8 downgrade-protection safeguard is in place. PR 5 (next) lifts the `PartitionCount > 1` dormancy gate and wires `PollSQSHTFIFOCapability` (#721) into the CreateQueue gate in the same commit.

### raftengine: leader-acquired observer (mirror of leader-loss)

- `raftengine.Admin` gains `RegisterLeaderAcquiredCallback`. Same contract as `RegisterLeaderLossCallback` (non-blocking, panic-contained, sentinel-pointer deregister) but fires on the `previous!=Leader → status==Leader` edge in `refreshStatus`.
- The etcd backend's slot-management is now shared between leader-loss and leader-acquired via `registerLeaderCallback` + `gatherLeaderCallbacks` helpers — this satisfies the `dupl` lint check and keeps both paths consistent in one place.
- 7 observer tests mirror the existing leader-loss tests: panic-containment, empty-list safety, deregister removal, idempotence, nil-fn safe, nil-receiver safe, identical-fn disambiguation.

### main: SQS leadership-refusal hook

- `main_sqs_leadership_refusal.go`:
  - `installSQSLeadershipRefusal(ctx, admin, gid, partitionedGroups, advertisesHTFIFO, logger) func()` — startup check + per-acquisition observer.
  - `installSQSLeadershipRefusalAcrossGroups(...)` — composite installer iterating every shard runtime.
  - `partitionedGroupSet(partitionMap, logger)` — flattens `--sqsFifoPartitionMap` into the `{gid → bool}` set the hook consumes.
  - `sqsAdvertisesHTFIFO()` — wraps `adapter.AdvertisesHTFIFO()`.
- `run()` installs the composite refusal across runtimes after the coordinator is built; deregister flows through `cleanup`.
- `TransferLeadership` runs in a goroutine because the leader-acquired callback contract is non-blocking — a synchronous admin RPC inside the callback would stall `refreshStatus`.
- 11 helper tests cover the contract: htfifo no-op, no-partitioned-queue no-op, startup-already-leader refuses, startup-follower waits, per-acquisition fires, deregister propagates, transfer-error logged, nil-admin safe, `partitionedGroupSet` flatten / empty-input / malformed-ref-skip.

### adapter: `AdvertisesHTFIFO` + flag flip

- `adapter.AdvertisesHTFIFO()` exposes the package-private flag.
- `htfifoCapabilityAdvertised = false → true`. Both the routing wiring (PR 4-B-2 #715) and the leadership-refusal hook (this PR) are now in the binary, so the design's "marked htfifo-eligible" bar is met.

## What's still gated

PR 5 lifts the `PartitionCount > 1` dormancy gate AND wires `PollSQSHTFIFOCapability` (#721) into the CreateQueue gate in the same commit. Until PR 5 lands, no partitioned queue can land in production — the leadership-refusal hook is dormant in the happy-path runtime (every binary past this PR advertises `htfifo`, and the per-group early return keeps the hook out of the hot path).

## Test plan

- [x] `go test -race ./internal/raftengine/etcd/` — 7 new + existing tests pass.
- [x] `go test -race ./kv/` — existing 30+ tests pass (verified the tests that previously failed due to the misplaced interface method now pass).
- [x] `go test -race ./adapter/` + `go test -race .` — all pass.
- [x] `golangci-lint ./kv/... ./adapter/... ./internal/raftengine/... .` — clean.

1. **Data loss** — control-plane only; no FSM/Pebble/retention path. The hook calls TransferLeadership which is itself an admin action with the same data-loss profile as a graceful manual transfer. No issue.
2. **Concurrency / distributed failures** — leader-acquired callback contract mirrors leader-loss (non-blocking, panic-contained, sentinel-pointer deregister). `refuse()` offloads `TransferLeadership` to a goroutine so `refreshStatus` stays non-blocking. Multiple goroutines firing for the same group serialize on raft's admin channel; worst case is one redundant transfer attempt, which is idempotent on the raft side. No issue.
3. **Performance** — leader-acquired callbacks fire only on the transition edge (rare event); no per-request hot path cost. The early return on `advertisesHTFIFO=true` means production-binary hosts pay zero overhead. No issue.
4. **Data consistency** — the hook protects against the §8 downgrade scenario: a node rolled back to a pre-htfifo binary that still gets elected leader of a partitioned-queue shard would otherwise read/write under the legacy keyspace and silently corrupt the queue. The hook steps it down via TransferLeadership before any client request lands. No issue.
5. **Test coverage** — 7 raftengine observer tests + 11 main-side helper tests (18 new). Existing kv lease tests confirm the misplaced-interface bug from first draft was caught and fixed.

## Summary by CodeRabbit

* **New Features**
  * SQS health endpoint now advertises HT-FIFO capability status for client detection.
  * Raft administration adds leader-acquired callbacks so systems can react immediately when a node becomes leader.
  * SQS FIFO queue groups can automatically refuse and transfer leadership when HT-FIFO capability is not available.

* **Tests**
  * Added comprehensive tests for leader-acquired callbacks, deregistration, panic isolation, and SQS leadership-refusal behavior.

### DIFF
--- a/adapter/sqs.go
+++ b/adapter/sqs.go
@@ -60,14 +60,40 @@ const (
 const sqsCapabilityHTFIFO = "htfifo"
 
 // htfifoCapabilityAdvertised gates whether this binary lists
-// "htfifo" on /sqs_health. The flag is set to true only when the
-// binary contains BOTH the routing-layer wiring AND the
-// leadership-refusal safeguard from §8 — the design's "marked
-// htfifo-eligible" bar (§11 PR 4). Lower-numbered PRs in the rollout
-// keep this false so a partial deploy never advertises a capability
-// it cannot safely back up. Phase 3.D PR 4-B flips this to true in
-// the same commit that wires routing + leadership-refusal together.
-const htfifoCapabilityAdvertised = false
+// "htfifo" on /sqs_health. The §11 PR 4 contract requires BOTH
+// the routing-layer wiring AND the leadership-refusal safeguard
+// from §8 to be in place before this flag is true:
+//
+//   - Routing wiring: kv.PartitionResolver +
+//     adapter.SQSPartitionResolver, merged via #715 (Phase 3.D
+//     PR 4-B-2). Partition-resolver-first dispatch in ShardRouter
+//     routes (queue, partition) keys to the operator-chosen Raft
+//     group; coordinator helpers (groupForKey,
+//     routeAndGroupForKey, groupMutations) consult the resolver
+//     before falling through to the byte-range engine; OCC read
+//     keys fail closed for recognised-but-unresolved partitioned
+//     keys.
+//   - Capability poller: PollSQSHTFIFOCapability, merged via
+//     #721 (Phase 3.D PR 4-B-3a). PR 5 will use this for the
+//     CreateQueue capability gate.
+//   - Leadership-refusal hook:
+//     raftengine.RegisterLeaderAcquiredCallback +
+//     main_sqs_leadership_refusal.go (Phase 3.D PR 4-B-3b, this
+//     PR). On startup AND on every leader-acquired transition,
+//     the hook refuses leadership of any Raft group hosting a
+//     partitioned queue when the binary lacks htfifo.
+//
+// Both pieces are now in the binary, so the flag flips to true.
+// PR 5 lifts the PartitionCount > 1 dormancy gate AND wires the
+// CreateQueue capability poll in the same commit, at which point
+// a partitioned queue can land in production and every node in
+// the cluster must report htfifo for the gate to allow it.
+//
+// Stays a const (not a var) because the flag is build-time. A
+// future runtime override (env var, --no-htfifo flag for
+// graceful degradation) would reroute through
+// adapter.AdvertisesHTFIFO() without changing the call sites.
+const htfifoCapabilityAdvertised = true
 
 // sqsAdvertisedCapabilities returns the capability list emitted on
 // /sqs_health (JSON mode). Stable iteration order is significant —
@@ -81,6 +107,20 @@ func sqsAdvertisedCapabilities() []string {
 		caps = append(caps, sqsCapabilityHTFIFO)
 	}
 	return caps
+}
+
+// AdvertisesHTFIFO reports whether this binary's /sqs_health
+// endpoint lists the htfifo capability. Mirror of the package-
+// internal htfifoCapabilityAdvertised constant, exposed for the
+// SQS leadership-refusal hook in main.go that uses this signal
+// to decide whether to refuse leadership of any Raft group hosting
+// a partitioned FIFO queue.
+//
+// Stays a function (not an exported constant) so a future runtime
+// override (env var, --no-htfifo flag for graceful degradation)
+// can be threaded through here without changing the call site.
+func AdvertisesHTFIFO() bool {
+	return htfifoCapabilityAdvertised
 }
 
 const (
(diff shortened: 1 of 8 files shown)


bootjp:
@claude review

Phase 3.D PR 4-B-3b — leadership-refusal hook (raftengine leader-acquired observer + main-side helper) and `htfifoCapabilityAdvertised` flag flip. Closes the §11 PR 4 routing+leadership-refusal pair so the binary is now "marked htfifo-eligible". PR 5 lifts the dormancy gate next.

bootjp:
@claude review

Round 2 — all four round 1 findings addressed:

**P1 (Codex) / HIGH (Gemini)** — real data race in `gatherLeaderCallbacks`. My round 1 refactor passed the slice by value (`cbs []leaderCallbackSlot`), so the slice header was dereferenced at the call site BEFORE `mu.Lock()` ran inside the helper. Concurrent `registerLeaderCallback` could mutate the header during that window. Fix: signature changed to `*[]leaderCallbackSlot`; both call sites updated. The dereference now happens inside the locked section.

**MEDIUM (Gemini) / Claude finding 1** — `invokeLeaderAcquiredCallback` panic log now matches `invokeLeaderLossCallback` (node_id / raft_node_id / panic / debug.Stack).

**Claude finding 2 (TOCTOU)** — `installSQSLeadershipRefusal` now re-checks `State()` after `RegisterLeaderAcquiredCallback` returns. If the engine wins an election in the registration window, `refuse()` fires for that in-flight acquisition. `refuse()` is idempotent so a double-fire across the boundary is safe.

**Claude finding 3 (branch coverage)** — 4 new `installSQSLeadershipRefusalAcrossGroups` tests cover composite deregister fan-out, empty partition map, nil-runtime / nil-engine guard, and engine-doesn't-implement-observer. Plus `TestInstallSQSLeadershipRefusal_TOCTOUWindowCovered` directly exercises fix #2.

**Lessons-learned audit**: semantic change is `gatherLeaderCallbacks` signature. `grep -rn` confirmed only the two `fire*Callbacks` methods call it; both updated together. No external callers.

`go test -race ./kv/... ./internal/raftengine/etcd/ .` and `golangci-lint` clean.

bootjp:
@claude review

CI failure fix (round 3):

**Failure**: `test (ubuntu-latest)` → `TestServerMapsEngineAdminMethods` returned "add voter is not supported by this raft engine". Root cause: the type-assertion at `internal/raftadmin/server.go:21` (`any(engine).(raftengine.Admin)`) silently fell back to `admin=nil` after this PR extended the Admin interface with `RegisterLeaderAcquiredCallback`. The `fakeEngine` stub never gained the method, so it stopped satisfying the interface — and every admin RPC mapped through NewServer returned Unimplemented.

**Lessons-learned miss**: my round 2 audit was `grep -rn gatherLeaderCallbacks`, scoped to the function I was renaming, not to the Admin interface itself. Should have also done `grep -rn 'raftengine.Admin\\b'` to catch all type-assertion sites — `internal/raftadmin/server.go:21` and `adapter/test_util.go:424` were the two hits. Production sites (etcd Engine + adapter test_util's real raftengine) satisfy the new interface; only the raftadmin stub didn't.

**Fix**: `fakeEngine.RegisterLeaderAcquiredCallback` returns a no-op deregister. raftadmin doesn't exercise the leader-acquired observer; the stub's only obligation is to satisfy the type assertion.

`go test -race ./internal/raftadmin/ ./internal/raftengine/etcd/ ./kv/ .` and `golangci-lint` clean.

PR #731, opened 2026-05-02
feat(sqs): per-key dispatch helpers for partitioned-FIFO routing (Phase 3.D PR 5b-1)

bootjp:
## Summary

Foundation for PR 5b's Send/Receive partition fanout. Adds `sqsMsg*KeyDispatch` wrappers that route to the legacy or partitioned constructor based on `meta.PartitionCount`. **Pure scaffold**: helpers are added but no production call site uses them yet — those switch in PR 5b-2.

## Why a separate scaffold PR

PR 5b's full scope (Send/Receive fanout + receipt-handle v2 wiring + dormancy gate-lift + CreateQueue capability check) is ~1000 LOC across multiple files. The lessons-learned discipline from recent rounds (PR #723 missed the raftadmin stub because the audit was function-scoped, not interface-scoped) favours smaller diffs.

Splitting PR 5b into three:
- **PR 5b-1** (this PR): dispatch helpers + tests, zero behaviour change.
- **PR 5b-2**: SendMessage / ReceiveMessage / DeleteMessage / ChangeMessageVisibility wire through the helpers. Receipt-handle v2 dispatch wired. Still gated by §11 PR 2 dormancy.
- **PR 5b-3** (atomic): gate-lift + CreateQueue capability check via `PollSQSHTFIFOCapability`. Small-but-critical.

## What's added

`adapter/sqs_keys_dispatch.go` (new file):
- `sqsMsgDataKeyDispatch`, `sqsMsgVisKeyDispatch`, `sqsMsgDedupKeyDispatch`, `sqsMsgGroupKeyDispatch`, `sqsMsgByAgeKeyDispatch`, `sqsMsgVisPrefixForQueueDispatch`
- `effectivePartitionCount` helper for fanout iteration

Each helper takes `(meta, queueName, partition, gen, …)` and routes to the legacy single-partition constructor when `meta.PartitionCount <= 1` (existing behaviour, byte-identical output) or to the partitioned constructor (#703's keyspace) when `> 1`.

## What does NOT change yet

- 14 call sites in `adapter/sqs_messages.go` still call legacy `sqsMsg*Key` constructors directly. PR 5b-2 swaps them.
- `encodeReceiptHandleV2` stays dormant.
- `decodeClientReceiptHandle` still rejects v2 unconditionally.
- CreateQueue still rejects `PartitionCount > 1` with `InvalidAttributeValue`.

## Test plan

5 top-level tests (19 sub-cases):
[...]

PR #732, opened 2026-05-02
feat(sqs): wire partitioned-FIFO data plane through dispatch helpers (Phase 3.D PR 5b-2)

bootjp:
## Summary

Stage 2 of Phase 3.D PR 5b: wire `SendMessage` / `ReceiveMessage` / `DeleteMessage` / `ChangeMessageVisibility` through the per-key dispatch helpers landed in PR 5b-1 (#731). The §11 PR 2 dormancy gate still rejects `PartitionCount > 1` at `CreateQueue`, so production queues stay on the legacy keyspace and observable behaviour is byte-identical. **PR 5b-3 lifts the gate atomically with the capability check.**

### Send paths
- `sendMessage`, `sendBatchStandardOnce`: partition is a compile-time `0` (Standard queues reject `PartitionCount > 1` via the cross-attribute validator). Dispatch helpers route to the legacy keyspace on the steady-state path.
- `sendFifoMessage`: hashes `MessageGroupId` once via `partitionFor` and threads the result through every key construction (data, vis, byage, dedup, group-lock).

### Receive fanout
- `scanAndDeliverOnce` wraps the scan loop with a per-partition fanout over `effectivePartitionCount(meta)` iterations. The shared wall-clock + per-call `Max` budget caps apply across the whole call, not per-partition.
- `sqsMsgCandidate` carries a new `partition` field stamped at scan time, so `loadCandidateRecord` / `expireMessage` / `commitReceiveRotation` / `classifyFifoGroupLock` all route to the partition the message was originally stored under.

### Delete + ChangeMessageVisibility
- `handle.Partition` from the v2 receipt handle drives every key construction.
- `loadMessageForDelete` / `loadAndVerifyMessage` invoke the new `validateReceiptHandleVersion` once meta is in scope.

### Receipt-handle dispatch (the contract shift)
- `encodeReceiptHandleDispatch(meta, partition, ...)` is the single point that picks v1 vs v2.
- `decodeClientReceiptHandle` is now version-agnostic; the queue-aware version check moved to `validateReceiptHandleVersion`. v1-on-partitioned and v2-on-non-partitioned both surface as `ReceiptHandleIsInvalid`, **preserving the dormancy promise** (no v2 wire-format probability from the public API) under the new contract. Since the §11 PR 2 dormancy gate keeps every production queue non-partitioned in this PR, every v2 handle still surfaces as `ReceiptHandleIsInvalid` downstream — exactly the PR 5a observable behaviour.

### Reaper
The reaper iterates legacy byAge keys only — partition-iterating enumeration ships in the later partition-reaper PR. `buildReapOps` / `reapOneRecord` pass `nil` meta + `partition 0` through dispatch helpers so the keys are byte-identical to the pre-PR-5b layout.

## What does NOT change yet
- `CreateQueue` still rejects `PartitionCount > 1` with `InvalidAttributeValue` (the §11 PR 2 dormancy gate, lifted atomically in PR 5b-3).
- Production queues never enter the partitioned dispatch branches in this PR — observable behaviour is byte-identical.
- The reaper does not yet enumerate partitioned byAge keys (a later PR).

## Test plan

10 new tests across the contract surface.

**Unit tests** (`adapter/sqs_keys_dispatch_v2_test.go`):
- [x] `TestEncodeReceiptHandleDispatch_PicksVersionByPartitionCount` — pins the version dispatch decision across `nil` / `0` / `1` / `4` metas.
- [x] `TestEncodeReceiptHandleDispatch_LegacyByteIdenticalToV1` — protects the byte-identical guarantee on legacy queues.
- [x] `TestEncodeReceiptHandleDispatch_PerQueueUsesV2` — `perQueue + PartitionCount=4` still produces v2 handles (keyspace is partitioned even when `partitionFor` collapses every group to partition 0).
- [x] `TestValidateReceiptHandleVersion_QueueAwareRules` — 10 sub-cases covering the (`meta.PartitionCount` × `handle.Version`) matrix, including `perQueue + PartitionCount=4`.
- [x] `TestValidateReceiptHandleVersion_NilHandle` — defensive nil branch.
- [x] `TestValidateReceiptHandleVersion_RejectsV2OnNonPartitioned` — named regression for the dormancy guarantee under the new contract.
- [x] `TestSQSMsgVisScanBoundsDispatch_LegacyMatchesLegacy` — byte-identical to legacy `sqsMsgVisScanBounds` on legacy meta.
- [x] `TestSQSMsgVisScanBoundsDispatch_PartitionedUsesPartitionedPrefix` — different partitions yield disjoint scan ranges.
- [x] `TestSQSMsgVisScanBoundsDispatch_PerQueueOnPartitionedKeyspace` — pins the **PR #731 round 2 forward-note invariant**: `perQueue + PartitionCount=4` keeps the partitioned vis prefix at partition 0; collapsing to legacy would silently strand send writes.

**Integration tests** (`adapter/sqs_partitioned_dispatch_test.go`) install a partitioned meta directly on a queue created via the public API, short-circuiting the dormancy gate without disabling it for production `CreateQueue`:
- [x] `TestSQSServer_PartitionedFIFO_SendReceiveDeleteRoundTrip` — end-to-end smoke test: send 6 groups, receive surfaces all via fanout, every handle is v2, delete via v2 handle, queue is empty afterwards, legacy keyspace stays empty.
- [x] `TestSQSServer_PartitionedFIFO_RejectsV1Handle` — forged v1 handle on a partitioned queue surfaces as `ReceiptHandleIsInvalid` via `DeleteMessage` and `ChangeMessageVisibility`.
- [x] `TestSQSServer_PartitionedFIFO_PerQueueCollapsesToPartitionZero` — `perQueue + PartitionCount=4` receive surfaces every message in one fanout pass; every v2 handle records `Partition=0`.

Updated `TestDecodeClientReceiptHandle_RejectsV2` → `TestDecodeClientReceiptHandle_AcceptsV2` to reflect the contract shift (rejection moved from API boundary to meta-aware `validateReceiptHandleVersion`).

- [x] `go test -race ./adapter/...` (targeted SQS scope) clean.
- [x] `golangci-lint run ./adapter/...` clean.

1. **Data loss** — Dispatch helpers byte-identical on legacy queues (`PartitionCount<=1` routes to legacy constructors). DLQ FIFO computes `dlqPartition` via `partitionFor(dlqMeta, srcRec.MessageGroupId)` so cross-queue redrive lands in the right partition. Receive fanout scans every partition. No issue.
2. **Concurrency** — Receive fanout iterates partitions sequentially under one shared wall-clock + max budget, so the original per-call SLA is preserved. `meta` is loaded once and passed by pointer through helpers; no concurrent meta refresh in a single call. `validateReceiptHandleVersion` runs after the gen check so a `SetQueueAttributes` race cannot flip the answer (`PartitionCount` is immutable). `go test -race` clean.
3. **Performance** — Hot path adds 1 nil-check + 1 `PartitionCount` compare per dispatch. Receive fanout iterates `effectivePartitionCount(meta)` times: 1 on legacy / `perQueue`, N on `perMessageGroupId` with N partitions splitting roughly 1/N of the messages each.
4. **Data consistency** — Receipt handle's recorded partition matches the storage partition (`commitReceiveRotation` uses `cand.partition` for both `newVisKey` and the handle). Cross-version handles rejected as `ReceiptHandleIsInvalid` — no leak between keyspaces. `perQueue + PartitionCount=4` invariant pinned by both unit and integration tests.
5. **Test coverage** — 10 new tests across the contract surface. Existing legacy-path SQS tests pass unchanged.

## Caller audit (semantic-change discipline)

`decodeClientReceiptHandle` semantics changed from "reject all v2" to "decode any version, defer version validation to meta-aware caller". All 3 production callers audited via grep:

- `parseQueueAndReceipt` (`sqs_messages.go:1576`) → `deleteMessageWithRetry` → `loadMessageForDelete` → `validateReceiptHandleVersion` after meta load. ✅
- `DeleteMessageBatch` entry (`sqs_messages_batch.go:475`) → `deleteMessageWithRetry` → same as above. ✅
- `ChangeMessageVisibilityBatch` entry (`sqs_messages_batch.go:575`) → `changeVisibilityWithRetry` → `loadAndVerifyMessage` → `validateReceiptHandleVersion`. ✅

The dormancy guarantee (v2 → `ReceiptHandleIsInvalid` on non-partitioned) is preserved across every caller path.

## Follow-ups

- PR 5b-3 (next): atomic gate-lift + `CreateQueue` capability check via `PollSQSHTFIFOCapability` (#721). Removes the `validatePartitionDormancyGate` call and verifies cluster-wide HT-FIFO support before accepting `PartitionCount > 1`.
- PR 6: partition-iterating reaper (drops the "legacy byAge only" caveat in `buildReapOps`).

## Summary by CodeRabbit

* **New Features**
  * Partition-aware message routing for FIFO and standard queues.
  * Receipt-handle v2 introduced for partitioned queues; v1 retained for non-partitioned queues.
  * Per-queue receive fanout rotation to round-robin partitions.

* **Bug Fixes**
  * Stronger queue-aware receipt-handle validation and bounds checking.
  * Partition-aware fixes across send/receive/delete/change-visibility/redrive/reaper flows.
  * Per-queue fanout state reset on queue create/delete.

* **Tests**
  * Extensive partitioned dispatch, receipt-handle, visibility-scan, dedup, rotation, and end-to-end tests.

### DIFF
--- a/adapter/sqs.go
+++ b/adapter/sqs.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/bootjp/elastickv/kv"
@@ -167,6 +168,15 @@ type SQSServer struct {
 	// throttle config so unconfigured queues pay one nil-check per
 	// request and nothing else (see sqs_throttle.go).
 	throttle *bucketStore
+	// receiveFanoutCounter advances once per partitioned receive call
+	// to pick the fanout starting partition. Using a per-server
+	// atomic counter (instead of the readTS-bit fold the round-1 fix
+	// shipped) guarantees true round-robin coverage across partitions
+	// even when readTS advances by a structured stride — the case
+	// Codex flagged as P1 round 2 (HLC's 16-bit logical counter +
+	// fixed per-receive transaction count can produce stride patterns
+	// that masked-AND alias to a subset of partitions).
+	receiveFanoutCounter atomic.Uint32
 }
 
 // WithSQSLeaderMap configures the Raft-address-to-SQS-address mapping used to
--- a/adapter/sqs.go
+++ b/adapter/sqs.go
@@ -7,7 +7,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
-	"sync/atomic"
+	"sync"
 	"time"
 
 	"github.com/bootjp/elastickv/kv"
@@ -168,15 +168,22 @@ type SQSServer struct {
 	// throttle config so unconfigured queues pay one nil-check per
 	// request and nothing else (see sqs_throttle.go).
 	throttle *bucketStore
-	// receiveFanoutCounter advances once per partitioned receive call
-	// to pick the fanout starting partition. Using a per-server
-	// atomic counter (instead of the readTS-bit fold the round-1 fix
-	// shipped) guarantees true round-robin coverage across partitions
-	// even when readTS advances by a structured stride — the case
-	// Codex flagged as P1 round 2 (HLC's 16-bit logical counter +
-	// fixed per-receive transaction count can produce stride patterns
-	// that masked-AND alias to a subset of partitions).
-	receiveFanoutCounter atomic.Uint32
+	// receiveFanoutCounters maps queueName → *atomic.Uint32 so each
+	// partitioned queue gets its own round-robin starting partition.
+	// Codex P1 round 4 flagged that a server-wide counter aliases
+	// across queues: when other queues' receives interleave with a
+	// stride that shares a factor with PartitionCount, the queue's
+	// observed counter subsequence cycles through only a subset of
+	// partitions, which can starve the rest under MaxNumberOfMessages
+	// pressure on the early-scanned ones. Per-queue isolation makes
+	// each queue's rotation depend solely on its own receive cadence.
+	//
+	// sync.Map is the right shape: lookups are read-mostly (the same
+	// queue keeps getting the same counter), and the keyset grows
+	// only with the number of distinct queues this server has handled
+	// receives for in-process — bounded by the operator-controlled
+	// CreateQueue rate.
+	receiveFanoutCounters sync.Map
 }
 
 // WithSQSLeaderMap configures the Raft-address-to-SQS-address mapping used to
--- a/adapter/sqs_catalog.go
+++ b/adapter/sqs_catalog.go
@@ -1012,6 +1012,12 @@ func (s *SQSServer) tryCreateQueueOnce(ctx context.Context, requested *sqsQueueM
 	// the new queue starts with a fresh full-capacity bucket
 	// regardless of in-flight traffic to the prior incarnation.
 	s.throttle.invalidateQueue(requested.Name)
+	// Mirror the throttle invalidate for the per-queue fanout-rotation
+	// counter. A delete-then-create race could otherwise leave the
+	// new queue starting partitioned receives at the previous
+	// incarnation's counter offset (harmless for correctness, but
+	// ties new-queue routing to old-queue history).
+	s.dropReceiveFanoutCounter(requested.Name)
 	return true, nil
 }
 
@@ -1038,6 +1044,12 @@ func (s *SQSServer) deleteQueue(w http.ResponseWriter, r *http.Request) {
 	// surprising operators who use DeleteQueue+CreateQueue to reset
 	// queue state.
 	s.throttle.invalidateQueue(name)
+	// Drop the per-queue fanout-rotation counter as well. Without
+	// this, repeated DeleteQueue of unique queue names retains one
+	// receiveFanoutCounters entry per name for the process lifetime
+	// — a leak in multi-tenant / high-churn deployments. Codex P2,
+	// PR #732 round 5.
+	s.dropReceiveFanoutCounter(name)
 	// SQS DeleteQueue returns 200 with an empty body.
 	writeSQSJSON(w, map[string]any{})
 }
--- a/adapter/sqs_fifo.go
+++ b/adapter/sqs_fifo.go
@@ -91,8 +91,8 @@ func resolveFifoDedupID(meta *sqsQueueMeta, in sqsSendMessageInput) string {
 // or (nil, nil) when there is no live record for this dedup-id.
 // Expired records are surfaced as nil so a stale entry does not block
 // a fresh send within the same FIFO queue.
-func (s *SQSServer) loadFifoDedupRecord(ctx context.Context, queueName string, gen uint64, dedupID string, readTS uint64) (*sqsFifoDedupRecord, []byte, error) {
-	key := sqsMsgDedupKey(queueName, gen, dedupID)
+func (s *SQSServer) loadFifoDedupRecord(ctx context.Context, queueName string, meta *sqsQueueMeta, partition uint32, gen uint64, dedupID string, readTS uint64) (*sqsFifoDedupRecord, []byte, error) {
+	key := sqsMsgDedupKeyDispatch(meta, queueName, partition, gen, dedupID)
 	raw, err := s.store.GetAt(ctx, key, readTS)
 	if err != nil {
 		if errors.Is(err, store.ErrKeyNotFound) {
@@ -130,11 +130,11 @@ func (s *SQSServer) loadFifoSequence(ctx context.Context, queueName string, read
 
 // loadFifoGroupLock fetches the in-flight lock for a group, if any.
 // Returns nil when no lock is held. Callers that also need the key
-// can recompute it via sqsMsgGroupKey — the helper used to return it
-// alongside the lock, but every caller already had the key in scope
-// from a different code path.
-func (s *SQSServer) loadFifoGroupLock(ctx context.Context, queueName string, gen uint64, groupID string, readTS uint64) (*sqsFifoGroupLock, error) {
-	key := sqsMsgGroupKey(queueName, gen, groupID)
+// can recompute it via sqsMsgGroupKeyDispatch — the helper used to
+// return it alongside the lock, but every caller already had the
+// key in scope from a different code path.
+func (s *SQSServer) loadFifoGroupLock(ctx context.Context, queueName string, meta *sqsQueueMeta, partition uint32, gen uint64, groupID string, readTS uint64) (*sqsFifoGroupLock, error) {
+	key := sqsMsgGroupKeyDispatch(meta, queueName, partition, gen, groupID)
 	raw, err := s.store.GetAt(ctx, key, readTS)
 	if err != nil {
 		if errors.Is(err, store.ErrKeyNotFound) {
@@ -172,7 +172,14 @@ func (s *SQSServer) sendFifoMessage(
 	delay int64,
 	readTS uint64,
 ) (map[string]string, bool, error) {
-	dedup, dedupKey, err := s.loadFifoDedupRecord(ctx, queueName, meta.Generation, dedupID, readTS)
+	// HT-FIFO: hash the MessageGroupId once at the entry point so
+	// every key built in this transaction (data, vis, byage, dedup,
+	// group-lock, sequence) lands in the same partition. partitionFor
+	// returns 0 on legacy / non-partitioned queues and on the
+	// perQueue throughput short-circuit, so the dispatch helpers
+	// round-trip to legacy output for those cases.
+	partition := partitionFor(meta, in.MessageGroupId)
+	dedup, dedupKey, err := s.loadFifoDedupRecord(ctx, queueName, meta, partition, meta.Generation, dedupID, readTS)
 	if err != nil {
 		return nil, false, err
 	}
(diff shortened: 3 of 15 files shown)


bootjp:
@claude review

bootjp:
@claude review

Round 1 addressing Codex P1: rotate the receive-fanout starting partition by readTS so a sustained backlog on a single partition cannot starve higher-index partitions.

Changes:
- `adapter/sqs_messages.go` — new `startPartitionOffset(partitions, readTS)` helper. `scanAndDeliverOnce` now iterates `(startOffset + i) % partitions`. Implementation folds the upper and lower 32 bits of the HLC-derived readTS and mask-ANDs by `partitions - 1` (power-of-two by validator invariant), keeping the computation in uint32 space (gosec G115 clean, no //nolint).
- `adapter/sqs_partitioned_dispatch_test.go` — `TestStartPartitionOffset_RotatesByReadTS` pins: legacy/perQueue → 0, every partition appears as a starting point across consecutive readTS, and the output is always in [0, partitions).

FIFO ordering: unaffected. `partitionFor` is deterministic, so a MessageGroupId always lands in one partition; cross-partition iteration order never reorders messages within any group.

Caller audit: `scanAndDeliverOnce` and `scanAndDeliverPartition` signatures and error semantics are unchanged — only the internal iteration order rotates. The new helper has one in-package caller.

P2/medium findings (out-of-range v2 partition validation, prefix-scan helper in test) are intentionally deferred — below the loop's P0/P1/high/critical/major threshold.

bootjp:
@claude review

Round 2 — addressed three CodeRabbit 🟠 Major findings:

1. `validateReceiptHandleVersion` now bounds-checks `handle.Partition < meta.PartitionCount` (the v2 partition is client-controlled once decoding accepts v2). Audit: both production callers (`loadMessageForDelete`, `loadAndVerifyMessage`) already surface any non-nil return as `ReceiptHandleIsInvalid` HTTP 400, so the new failure mode joins the existing public-error bucket — no caller-side semantic change.
2. Round-trip test now asserts `parsed.Partition == partitionFor(meta, group)` per message (was: only `< 4`), so a dispatch regression that collapsed every group to partition 0 would actually fail.
3. Post-delete emptiness check now probes the partitioned data keyspace directly per partition, and the public-API receive uses a 1s visibility timeout after a 1.1s sleep — a regression that left messages invisible but not deleted can no longer pass under the active 60s window.

Plus a new regression `TestValidateReceiptHandleVersion_RejectsOutOfRangePartition` covering several out-of-range partitions and the count-1 boundary.

bootjp:
@claude review

Round 3 — addressed two new Codex 🟠 P1 findings on the round-2 commit:

**1. Dedup scoping (`sqs_fifo.go:95`):** `sqsPartitionedMsgDedupKey` was `(queue, partition, dedupID)` — but `DeduplicationScope=messageGroup` (the only scope reachable on partitioned queues) requires the dedup window to be per (queue, group, dedupID). Two distinct MessageGroupIds that FNV-collide onto the same partition would share a dedup namespace, so a fresh send in group B reusing group A's dedup-id would be silently acked with group A's MessageId — data loss. Threaded `groupID` through `sqsPartitionedMsgDedupKey`, `sqsMsgDedupKeyDispatch`, and `loadFifoDedupRecord`.

**Caller audit:** `loadFifoDedupRecord` has exactly one production caller (`sendFifoMessage` at `sqs_fifo.go:186`); the same dedupKey returned from `loadFifoDedupRecord` populates both the OCC `ReadKeys` (line 238) and the `Put` write (line 243), so read-guard and write share one key. Legacy (non-partitioned) shape is unchanged on disk — the legacy branch ignores groupID, preserving compatibility for queues created before partitioning landed.

**2. Fanout aliasing (`sqs_messages.go:933`):** Round-1's XOR-fold of `readTS` bits aliases under structured strides — HLC packs a 16-bit logical counter in low bits and ReceiveMessage commits a fixed number of per-message transactions per call, so consecutive readTS deltas have a fixed stride. With partitions=4 and stride=10 the masked fold alternates 0,2,0,2,… and partitions 1,3 never appear as the start. Replaced with a per-server `atomic.Uint32` counter — strict round-robin regardless of HLC behaviour. Counter is per-server (not Raft-replicated) because fanout distribution is a local-fairness property.

Two new regression tests:
- `TestSQSDedupKeyDispatch_PartitionedScopesByMessageGroupId` — distinct MessageGroupIds on the same (queue, partition, dedupID) must produce distinct dedup keys; legacy keyspace unchanged.
- `TestNextReceiveFanoutStart_RoundRobin` — 16 calls over 4 partitions produce exactly 4 hits per partition (strict round-robin); concurrent receivers tested under `-race`.

`go build`, `go vet`, `golangci-lint`, and the targeted SQS test suite all clean.

bootjp:
@claude review

Round 4 — addressed Codex 🟠 P1 round 4: per-queue fanout counter.

The round-3 server-wide counter aliased across queues — when queues interleave with strides that share a factor with `PartitionCount`, a queue cycles through only a subset of partitions. Concrete: 4-partition queue receiving every other tick observes 1,3,5,7,… mod 4 = 1,3,1,3 → partitions 0 and 2 never appear.

Replaced `atomic.Uint32` field with `sync.Map` keyed by queue name; each queue gets its own `*atomic.Uint32`. Each queue's rotation now depends only on its own receive cadence — strict round-robin regardless of cross-queue interleaving.

**Caller audit:** `nextReceiveFanoutStart` has one production caller (`scanAndDeliverOnce` at `sqs_messages.go:885`), which now passes `queueName` from existing scope. Return type and range unchanged ([0, partitions)).

**Test:** `TestNextReceiveFanoutStart_PerQueueRoundRobin` (renamed) now asserts queueB and queueC interleaved each see strict round-robin — the round-3 server-wide counter would fail this. Concurrent receivers across many queue names exercise `LoadOrStore` under `-race`.

`go build`, `go vet`, `golangci-lint`, and the broader SQS test suite all clean (10.5s).

bootjp:
@claude review

Round 5 — addressed Codex P2 (user-elevated): per-queue fanout counter eviction.

The round-4 per-queue `receiveFanoutCounters` sync.Map populates one entry per queue name on the first partitioned receive but never removes it. Repeated DeleteQueue+CreateQueue of unique names then leaks one entry per name for the process lifetime — a real concern in multi-tenant / high-churn deployments.

**Fix:** new `dropReceiveFanoutCounter(queueName)` helper, mirrored to the same two call sites as the existing `throttle.invalidateQueue`:
- `createQueue` genuine-create branch — covers a delete-then-create race where a partitioned receive holding pre-delete meta could repopulate the entry between the delete-side cleanup and this create commit.
- `deleteQueue` HTTP handler — drops the counter alongside the throttle bucket so DeleteQueue+CreateQueue resets every piece of in-memory queue state.

`setQueueAttributes` is intentionally NOT mirrored — it only mutates throttle config, never PartitionCount, so the counter stays valid across attribute changes (same conditional shape the throttle invalidate already uses there).

**Caller audit:** new helper has exactly two production callers, both adjacent to existing `throttle.invalidateQueue` calls. `AdminDeleteQueue` does not call `throttle.invalidateQueue` today either — staying consistent (not fixing that pre-existing inconsistency in this round).

**Test:** `TestDropReceiveFanoutCounter_ClearsEntry` pins
- no-op on never-touched queues (no allocation, no panic)
- removal after a populated counter
- per-queue isolation (dropping queueB leaves queueC intact)
- fresh allocation after drop (first offset == 1 on 4-partition queue, proving the old pointer was not recycled)

Build, `golangci-lint`, and the targeted SQS test sweep (`TestDropReceiveFanoutCounter | TestNextReceiveFanoutStart | TestSQSServer_PartitionedFIFO | TestValidateReceiptHandleVersion`, plus a wider `CreateQueue|DeleteQueue|PurgeQueue|SetQueueAttributes|SendMessage|ReceiveMessage|DeleteMessage` regression sweep) all clean under `-race`.

bootjp:
@claude review

Round 6 — addressed CodeRabbit 🟠 Major: dedup key segment-boundary ambiguity.

`sqsPartitionedMsgDedupKey` was appending `encodeSQSSegment(groupID)` and `encodeSQSSegment(dedupID)` back-to-back. `encodeSQSSegment` uses `base64.RawURLEncoding` (no padding, alphabet `[A-Za-z0-9_-]`), so the boundary is variable-length and ambiguous: distinct `(group, dedup)` pairs can collapse onto the same byte sequence — most blatantly when one is empty (`("", "abcd")` and `("abcd", "")` both produce `"...QUJDRA"`). This is the on-disk format PR 5b-3 will rely on, so it had to be fixed before that PR.

**Fix:** insert `sqsPartitionedQueueTerminator` `|` between the two segments. `|` is outside RawURLEncoding's alphabet so neither segment can contribute one of its own — boundary unambiguous regardless of input length. Mirrors how `sqsPartitionedMsgDataKey`, `MsgVisKey`, `MsgGroupKey`, and `MsgByAgeKey` already use the same terminator at the queue/partition boundary.

**Caller audit:** `sqsPartitionedMsgDedupKey` is reached only through `sqsMsgDedupKeyDispatch`, which has exactly two production callers — `loadFifoDedupRecord` (read-side) and `sendFifoMessage` (write-side, using the key returned from the same `loadFifoDedupRecord` call). Read and write share one helper invocation, so the new format takes effect symmetrically — no read/write skew window. `parsePartitionedSQSKey` only inspects the key through the partition field and stops; the new terminator after groupID does not affect parsing.

**Dormancy:** partitioned queues are still gated at CreateQueue (§11 PR 2, lifted atomically with the capability check in PR 5b-3), so no production data carries the old format. Pure forward-format correction.

**Test (`TestSqsPartitionedMsgDedupKey_GroupDedupSeparator`):**
- Empty-segment collision class (the most blatant ambiguity)
- New terminator exact placement (key ends with `b64(group)|b64(dedup)`)
- Non-empty pair non-collision (`("ab", "cd")` vs `("abcd", "")`)
- Read-write symmetry through the dispatch helper

Build, `golangci-lint`, and the targeted dedup/dispatch/partitioned-FIFO test sweep all clean under `-race`. CodeRabbit's other finding on the test (🟡 Minor on the post-delete check comment) is below the loop threshold and not addressed in this round.

PR #736, opened 2026-05-04
feat(sqs): live-queue reaper enumerates partitioned keyspace (Phase 3.D PR 6b)

bootjp:
## Summary

Phase 3.D PR 6b: live-queue reaper enumerates partitioned keyspace. Follow-up to PR #735 (6a) — the tombstone-driven sweep already walks partitioned data on DeleteQueue / PurgeQueue, but the live-queue retention reaper still only saw the legacy keyspace, so:

- retention-expired messages on partitioned queues leaked their data / vis / byage / group rows forever (`reapQueue` walked `sqsMsgByAgePrefixAllGenerations` only),
- expired dedup records on partitioned FIFO queues leaked forever (`reapExpiredDedup` scanned `SqsMsgDedupPrefix` only — empty for partitioned queues since `sqsMsgDedupKeyDispatch` routes their writes under `SqsPartitionedMsgDedupPrefix`).

Closes the live-queue half of the Codex P2 from PR #732 round 0; PR 6a covered the tombstoned-cohort half.

## What changes

- **`reapQueue`**: legacy byage walk extracted as `reapQueueLegacy` (byte-identical to pre-PR-6b for non-partitioned queues). Adds `reapQueuePartition` step that runs once per partition for `PartitionCount > 1` queues. Per-partition budget per the §6 design ("partitions × budget per cycle"); 30s tick interval comfortably absorbs.

- **`reapPartitionedPage`**: partitioned twin of `reapPage`. Same live-vs-orphan classification, but parses each entry with `parseSqsPartitionedMsgByAgeKey` and routes the dispatch through `reapOneRecordPartitioned`.

- **`classifyPartitionedByAgeEntry`**: helper extracted from `reapPartitionedPage` so the loop body stays under the cyclop ceiling. Returns `(parsedKey, reapable bool)`.

- **`reapExpiredDedup`** (signature changed): now takes `*sqsQueueMeta` and routes by `PartitionCount`. Legacy meta → `reapExpiredDedupLegacy` (byte-identical). Partitioned meta → `reapExpiredDedupPartitioned` (NEW), iterates each partition's dedup prefix under its own per-partition budget.

## Caller audit

- `reapQueue` — one production caller (`reapAllQueues`); signature unchanged. Non-partitioned queues byte-identical; partitioned get the extra per-partition pass.
- `reapExpiredDedup` — signature changed to take `*sqsQueueMeta`; one production caller (`reapAllQueues`), updated. No tests called it directly.
[...]

PR #735, opened 2026-05-04
feat(sqs): tombstone reaper enumerates partitioned keyspace (Phase 3.D PR 6a)

bootjp:
## Summary

Phase 3.D PR 6a: tombstone reaper enumerates partitioned keyspace. Closes the Codex P2 from PR #732 round 0 ("Reap partitioned dedup records to prevent key growth"), deferred at the time as a "PR 6 concern". With PR #734 (5b-3) lifting the dormancy gate, partitioned FIFO queues can now land in production — but until this PR the tombstone reaper only swept the legacy keyspace, so partitioned data / vis / byage / dedup / group rows leaked permanently after DeleteQueue / PurgeQueue.

## What changes

- **Tombstone schema**: tombstone value now encodes `PartitionCount` so the reaper can drive partition iteration off the tombstone alone (the meta record is gone by the time the reaper observes the tombstone). 8 bytes big-endian uint64. Backward compat: `PartitionCount<=1` keeps the byte-identical legacy `[]byte{1}` sentinel; non-canonical values (empty, 1-byte, 4-byte, out-of-range 8-byte) decode to `PartitionCount=1` so a partial rollback or future encoding revision degrades safely to the legacy single-partition reap.

- **DeleteQueue + PurgeQueue** write the encoded value, reading `PartitionCount` from the pre-delete / pre-purge meta (immutable across SetQueueAttributes / PurgeQueue per §3.2).

- **Tombstone reaper**: `reapTombstonedGeneration` takes the decoded count, sweeps the legacy keyspace (always — covers pre-HT-FIFO queues plus a defensive pass) and on `PartitionCount > 1` ALSO sweeps every partition's byage / dedup / group prefix family via the new `reapPartitionedGeneration` helper. `allDone` gating delays tombstone deletion until every partition is fully drained.

- **Per-record dispatch**: `reapOneRecord` and `buildReapOps` now take `*sqsQueueMeta` + `partition uint32` instead of hardcoding `nil + 0`. Legacy callers pass `nil + 0` (byte-identical to pre-PR-5b layout); the partitioned twin `reapOneRecordPartitioned` synthesises a meta carrying `PartitionCount > 1` so the dispatch helpers route to `sqsPartitionedMsg{Data,Vis,Group}Key`.

- **New parsed-key helpers** in `sqs_keys.go`:
  - `sqsPartitionedMsgByAgePrefixForPartition(queue, partition, gen)`
  - `sqsPartitionedMsgDedupKeyPrefix(queue, partition, gen)`
[...]

PR #737, opened 2026-05-04
feat(sqs): HT-FIFO partition metrics counter (Phase 3.D PR 7a)

bootjp:
## Summary

Phase 3.D PR 7a — adds the `elastickv_sqs_partition_messages_total{queue, partition, action}` Prometheus counter so dashboards and alerts can spot uneven `MessageGroupId` distributions across partitioned-FIFO queues. Metrics-only: the Jepsen HT-FIFO workload (PR 7b) ships separately so the two review loops do not cross.

- `monitoring/sqs.go` (new): `SQSMetrics` with cardinality cap (`sqsMaxTrackedQueues = 512`, overflow collapses to `_other`) mirroring `DynamoDBMetrics`. Drops empty queue names and unknown action labels so a future call-site bug cannot pollute the series space dashboards have to learn about. Nil-receiver-safe so adapter call sites do not need to nil-guard.
- `adapter/sqs.go`: `SQSPartitionObserver` interface + `WithSQSPartitionObserver` option. Re-declared in `adapter` so it doesn't import `monitoring` at the package boundary (matches the DynamoDB/Redis observer pattern). Action constants (`send`/`receive`/`delete`) re-declared on the adapter side and validated at runtime by the monitoring side — drift between the two surfaces as a dropped observation, not a wedge.
- `adapter/sqs_fifo.go`, `adapter/sqs_messages.go`: emit the counter on the **partitioned** commit branch only (`PartitionCount > 1`) for send / receive / delete. Legacy single-partition queues stay off the metric since partition is always 0 and the cardinality cost would buy nothing.
- `monitoring/registry.go`, `main_sqs.go`, `main.go`: wire the registry's `SQSPartitionObserver()` into `startSQSServer` so the SQS server picks up the production observer on cluster boot. Test fixtures and CLI tools that build `SQSServer` without a registry pass `nil` and the metric stays at zero.

## Tests

`monitoring/sqs_test.go` (new, 6 cases):

- `TestSQSMetrics_ObservePartitionMessage_IncrementsByLabelTriple` — pin the `(queue, partition, action)` counter contract.
- `TestSQSMetrics_ObservePartitionMessage_DropsInvalidAction` — pin the typo guard against future drift between adapter and monitoring constants.
- `TestSQSMetrics_ObservePartitionMessage_DropsEmptyQueue` — pin that an empty queue name does not collapse with valid observations onto a shared series.
- `TestSQSMetrics_NilReceiverIsSafe` — pin the nil-receiver short-circuit the adapter relies on.
- `TestSQSMetrics_QueueLabelOverflow` — pin the cap-and-collapse so a misbehaving caller cannot exhaust the Prometheus series budget.
- `TestSQSMetrics_RegistryWiring` — pin that the public `Registry` exposes the metric under the documented name.

## Self-review (5 lenses)

1. **Data loss** — N/A; metrics-only, no storage / Raft / FSM touch.
2. **Concurrency** — counter increments are atomic via Prometheus; the `trackedQueues` map is only consulted from the dispatch-success path under the SQS server's existing concurrency model. No new locks.
3. **Performance** — one map lookup + one `CounterVec` lookup per partitioned send/receive/delete on the success branch. Legacy queues skip the call entirely. Cardinality bounded at 512 queue × 32 partition (`htfifoMaxPartitions`) × 3 action ≈ 49k series worst case; in practice a 32-partition queue yields 96 series, so the budget is plenty for the SLO panels.
4. **Data consistency** — the metric is observed AFTER OCC dispatch succeeds, so the counter reflects committed state. Receive/delete branches that return on retryable errors deliberately do not increment (the retry path will observe on the eventual success).
5. **Test coverage** — 6 unit tests in `monitoring/`, plus the adapter-side nil-observer path is exercised by all existing partitioned-FIFO tests in `adapter/sqs_partitioned_dispatch_test.go` (they pass `nil` observer through the test fixture).

## Test plan

- [x] `go test -race -count=1 ./monitoring/...`
- [x] `go test -race -count=1 -run 'TestSQS' ./adapter/...`
- [x] `go test -race -count=1 ./...` (full suite)
- [x] `golangci-lint --config=.golangci.yaml run ./...` (full repo)
- [ ] Jepsen HT-FIFO workload — deferred to PR 7b

## Refs

- `docs/design/2026_05_01_partial_split_queue_fifo.md` §11 PR 7
- Builds on PR 5b-3 (#734) capability gate, PR 6a (#735) tombstone reaper, PR 6b (#736 in flight) live-queue reaper.

### DIFF
--- a/adapter/sqs.go
+++ b/adapter/sqs.go
@@ -200,7 +200,35 @@ type SQSServer struct {
 	// check so partitioned queues can land on a single-shard
 	// cluster and route through the engine's default group.
 	partitionResolver *SQSPartitionResolver
+	// partitionObserver records the
+	// elastickv_sqs_partition_messages_total{queue, partition,
+	// action} counter for HT-FIFO operations (PR 7a). nil on
+	// non-monitored test fixtures and on single-binary CLI
+	// tools that build SQSServer without a monitoring registry.
+	// Increment call sites use a nil-receiver-safe call so the
+	// metrics path costs nothing when unwired.
+	partitionObserver SQSPartitionObserver
 }
+
+// SQSPartitionObserver is the metrics-package interface
+// (monitoring.SQSPartitionObserver) re-declared here so the
+// adapter does not import monitoring at the package boundary —
+// matches the existing observer pattern for DynamoDB / Redis.
+type SQSPartitionObserver interface {
+	ObservePartitionMessage(queue string, partition uint32, action string)
+}
+
+// SQSPartitionAction* mirror the action label values from
+// monitoring.SQSPartitionAction*. Re-declared so adapter call
+// sites do not need a monitoring import; the observer interface
+// validates the value at runtime so a drift between these
+// constants and the monitoring side surfaces as a dropped
+// observation rather than a wedge.
+const (
+	SQSPartitionActionSend    = "send"
+	SQSPartitionActionReceive = "receive"
+	SQSPartitionActionDelete  = "delete"
+)
 
 // WithSQSLeaderMap configures the Raft-address-to-SQS-address mapping used to
 // forward requests from followers to the current leader. Format mirrors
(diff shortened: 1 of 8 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

PR #740, opened 2026-05-05
feat(sqs): listenless SQSServer for admin-only deployments

bootjp:
## Summary

Make `/admin/api/v1/sqs/*` reachable on builds that start without `--sqsAddress`. The admin handlers (list, describe, delete queues) only need the coordinator and store — not the public SigV4 listener — so this PR teaches `startSQSServer` to construct a listenless `*adapter.SQSServer` instead of returning `(nil, nil)` and dropping the admin bridge.

Concretely:

- `startSQSServer` no longer short-circuits when `sqsAddr` is empty. It still skips `lc.Listen` (no port is opened), but it constructs `NewSQSServer(nil, …)` and threads the result back through the runtime.
- `(*SQSServer).Run()` detects the nil listener and blocks on `reaperCtx.Done()` instead of calling `httpServer.Serve`. The reaper and throttle-sweep goroutines still run so retention math behind the admin counters stays correct.
- `main_admin.go`'s `newSqsQueuesSource` now sees a non-nil `*adapter.SQSServer` and wires `/admin/api/v1/sqs/*`; the `if sqsServer == nil { return nil }` guard stays as a defensive check for tests that still pass nil.
- The runtime-struct comment for `r.sqsServer` updated to reflect that the field is always non-nil after startup.

## Why this is "option A" of the prior discussion

We discussed three options earlier:
- **A** — listenless construction across all three adapters (SQS / S3 / Dynamo).
- **B** — refactor admin to use the catalog/coordinator directly, no adapter dependency. Crosses adapter and admin package boundaries; would warrant a `*_proposed_*.md` design doc per CLAUDE.md.
- **C** — keep current behaviour, document the gating.

You picked **A**. This PR scopes to **SQS only** since that was the immediate ask. Symmetric Dynamo/S3 changes can land as a parallel PR if you want; both follow the same structural pattern (drop early-return in `start*Server`, teach `(*XxxServer).Run()` to handle nil listener) but each adapter's `Run()` body has different background work to keep alive (S3 manifest cleanup workers, etc.) so a separate review pass per adapter is the cleaner cut.

## Caller / semantic audit

Per the standing rule on semantic-changing edits:

- `adapter/sqs_catalog.go::queueURL` already nil-guards `s.listen`; no other reads of `s.listen` exist in the adapter package (verified via grep).
- `main.go` runtime field: only comment updated. The Go type is unchanged; the only change is that the zero-value pathway is no longer reachable.
- `newSqsQueuesSource` still nil-guards its argument; no caller in `main.go` passes nil any more, but the guard remains for tests that construct admin servers without an SQS adapter.
- `(*SQSServer).Run()` now has two terminating paths (reaper-ctx-block and `httpServer.Serve` return). `Stop()` cancels the reaper ctx, which both paths observe — same shutdown semantics on either branch.

## Tests

- **`TestSQSServer_ListenlessRunStopRoundTrip`** (new) — pins the nil-listener Run/Stop round trip. Constructs `NewSQSServer(nil, nil, nil)`, launches `Run()` in a goroutine, calls `Stop()`, and asserts `Run()` returns nil within 2s. Without the new `s.listen == nil` branch in `Run()`, `httpServer.Serve(nil)` would panic and this test would fail.
- **`TestSQSServer_StopShutsDown`** (existing) — unchanged. Anchors the listening-path round trip so a regression in either branch does not silently swap them.

## Self-review (5 lenses)

1. **Data loss** — N/A; no storage / Raft / FSM touch.
2. **Concurrency** — `Run()` adds a second terminating path; both paths observe the same `reaperCtx` cancellation that `Stop()` triggers.
3. **Performance** — listenless mode adds zero work to the listening path. Listenless construction adds no Raft round-trips on startup.
4. **Data consistency** — admin handlers take a fresh `nextTxnReadTS` per call; the listener's presence never affected snapshot semantics.
5. **Test coverage** — `TestSQSServer_ListenlessRunStopRoundTrip` pins the new branch; the existing `TestSQSServer_StopShutsDown` anchors the listening branch.

## Test plan

- [x] `go test -race -count=1 -run 'TestSQS' ./adapter/...` (all SQS tests pass)
- [x] `golangci-lint --config=.golangci.yaml run ./...` clean
- [x] `go build ./...` clean
- [ ] Live manual smoke: start a cluster with `--sqsAddress=""`, confirm `/admin/api/v1/sqs/queues` returns the catalog list (operator-driven, out of scope for the merge gate)

## Refs

- `adapter/sqs_admin.go` — admin entrypoints
- `main_admin.go` — admin bridge

## Summary by CodeRabbit

* **New Features**
  * SQS server supports a listenless mode so admin operations run without a network listener; admin endpoints remain available.

* **Tests**
  * Added regression test to validate listenless startup and graceful shutdown behavior.

* **Chores**
  * Improved SQS initialization/error handling and added structured logging during listener and credential setup.

* **Documentation**
  * Clarified runtime behavior and when a listenless server is constructed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/740)

### DIFF
--- a/adapter/sqs.go
+++ b/adapter/sqs.go
@@ -326,6 +326,18 @@ func (s *SQSServer) Run() error {
 	// request hot path never pays the O(N) sweep cost. Cleaned up by
 	// the same reaperCtx cancellation that stops the message reaper.
 	go s.throttle.runSweepLoop(s.reaperCtx)
+	if s.listen == nil {
+		// Listenless mode: the SQS adapter was constructed without a
+		// public HTTP listener (--sqsAddress empty). Admin endpoints
+		// in adapter/sqs_admin.go still work because they go through
+		// the coordinator/store — only the SigV4 wire surface is
+		// suppressed. Block until Stop() cancels reaperCtx so the
+		// errgroup task lifetime matches the listening branch
+		// (callers wait on Run() returning to know it's safe to tear
+		// down the underlying coordinator/store).
+		<-s.reaperCtx.Done()
+		return nil
+	}
 	if err := s.httpServer.Serve(s.listen); err != nil && !errors.Is(err, http.ErrServerClosed) {
 		return errors.WithStack(err)
 	}
--- a/adapter/sqs.go
+++ b/adapter/sqs.go
@@ -349,6 +349,11 @@ func (s *SQSServer) Stop() {
 		s.reaperCancel()
 	}
 	if s.httpServer != nil {
+		// http.Server.Shutdown returns immediately when Serve was
+		// never called (listenless mode — no public listener was
+		// constructed), so this is a no-op for admin-only
+		// deployments. The branch is still gated on httpServer
+		// being non-nil because Shutdown panics on a nil receiver.
 		_ = s.httpServer.Shutdown(context.Background())
 	}
 }
--- a/adapter/sqs_test.go
+++ b/adapter/sqs_test.go
@@ -205,3 +205,33 @@ func TestSQSServer_StopShutsDown(t *testing.T) {
 		t.Fatal("Run did not return within timeout after Stop")
 	}
 }
+
+// TestSQSServer_ListenlessRunStopRoundTrip pins the listenless
+// construction path used by startSQSServer when --sqsAddress is
+// empty: NewSQSServer must accept a nil net.Listener, Run() must
+// return cleanly when Stop() cancels the reaper context, and the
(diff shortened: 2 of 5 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

PR #747, opened 2026-05-08
fix(raft): stream receive snapshot to disk + emit EKVT token (memory safety)

bootjp:
## Summary

PR #746 raised the receive-side spool cap to 16 GiB so a 1.35 GiB FSM transfer no longer fails mid-stream. But apply was still going through `buildSnapshotMessage` → `spool.Bytes()` → `io.ReadAll`, materializing the entire payload as `[]byte` for `raftpb.Snapshot.Data`. On a 2.5 GiB-container production node a 1.35 GiB allocation in addition to existing process state (Pebble cache, raft state, runtime) inflates heap past the OOM kill line before `RawNode.Step` ever runs `Restore`.

This PR routes receive into the existing token-format streaming path so heap usage during receive becomes O(chunk size), not O(FSM size).

> **Stacked on #746.** Target base: `fix/snapshot-spool-size-cap`. When #746 merges, this rebases against main automatically.

## What was already in place

The codebase already had all the apply-side machinery for streaming snapshot restore:

- `isSnapshotToken` / `decodeSnapshotToken` / `encodeSnapshotToken` (`fsm_snapshot_file.go`) — the 17-byte EKVT token format
- `openAndRestoreFSMSnapshot` — streams a `.fsm` file's payload into `StateMachine.Restore(io.Reader)` without materialization
- `engine.go::applySnapshot` already branches on `isSnapshotToken` and calls `openAndRestoreFSMSnapshot` when it sees a token
- `restoreSnapshotState` (`wal_store.go`) does the same on the startup-restore path
- `e.transport.SetFSMSnapDir(e.fsmSnapDir)` is wired at engine `Open`

What was missing was the **receive-side producing tokens**.

## What this PR adds

1. **`snapshotSpool.FinalizeAsFSMFile(fsmSnapDir, index, crc32c)`** — appends the 4-byte CRC32C footer, fsyncs, closes, atomically renames the spool file into `fsmSnapPath(fsmSnapDir, index)`. On success, ownership transfers to `fsmSnapDir` and `Close()` becomes a no-op so the renamed file is NOT removed.

2. **`receiveSnapshotStream`** wraps the spool with `crc32CWriter` so CRC accumulates as bytes hit disk (no extra read pass). After the final chunk, `finalizeReceivedSnapshot`:
   - **Streaming token path** (when `fsmSnapDir` is wired AND `metadata.Snapshot.Metadata.Index > 0`): calls `FinalizeAsFSMFile`, then sets `Snapshot.Data = encodeSnapshotToken(index, crc)` — 17 bytes.
   - **Legacy fallback** (otherwise): falls through to `buildSnapshotMessage` so tests without an `fsmSnapDir` keep behaving exactly as before this PR.

3. Receive log line gains a `format` field (`token` or `inline`) so an operator can distinguish the streaming path from legacy at a glance.

## Memory profile change

| | before | after |
|---|---|---|
| Disk write during receive | FSM size | FSM size |
| Heap during receive | chunk buffers (~MiB) | chunk buffers (~MiB) |
| Heap when `Snapshot.Data` is set | **FSM size** ← OOM hazard | **17 bytes** |
| `RawNode.Step` peak heap | FSM size | 17 bytes |
| `fsm.Restore` peak heap | FSM size (materialized) | chunk-bounded (`io.Reader`) |

A 16 GiB FSM would write 16 GiB to disk and allocate a 17-byte token plus per-chunk buffers. Container memory limits no longer constrain receive size.

## Self-review (5 lenses)

1. **Data loss** — none. The on-disk format produced by `FinalizeAsFSMFile` is byte-identical to `writeFSMSnapshotFile` (payload + 4-byte big-endian CRC32C footer). `openAndRestoreFSMSnapshot` reads either source uniformly. Tests prove a sender-FSM round-trips through streaming receive to a fresh receiver-FSM with identical `Applied()`.

2. **Concurrency** — no new locks. The spool's `path` and `file` fields are written from a single goroutine (the receive loop). `FinalizeAsFSMFile` nulls them on success so deferred `Close()` becomes a no-op without a separate `keep` flag. `crc32CWriter` wraps the spool linearly; never shared across goroutines.

3. **Performance** — one CRC32C pass over receive bytes. CRC32C on modern x86 (PCLMULQDQ) runs multi-GiB/s, well above gRPC stream throughput; invisible in profiles. Heap allocation for `Snapshot.Data` drops from O(FSM) to constant 17 bytes. Disk pattern unchanged: sequential append to spool, single rename (metadata-only), single fsync.

4. **Data consistency** — the EKVT token CRC is computed by the same `crc32CWriter` wrapper used by `writeFSMSnapshotPayload` (sender side), so the footer the receiver writes matches what `openAndRestoreFSMSnapshot`'s pre-Restore footer check expects. `ApplySnapshot` ordering is unchanged; this PR only swaps how `Snapshot.Data` is shaped before `RawNode.Step` sees it, and the apply branch on `isSnapshotToken` was already in place.

5. **Test coverage**:
   - `TestReceiveSnapshotStream_StreamingTokenWhenFSMSnapDirSet` pins:
     - `Snapshot.Data` is the 17-byte token, not the payload
     - spool dir is empty post-receive (file was renamed, not deleted, not orphaned)
     - `.fsm` file exists at `fsmSnapPath` with payload + 4-byte CRC footer
     - end-to-end round trip: `sender.Applied() == receiver.Applied()` after `restoreSnapshotState` reads the token, opens the `.fsm` file, streams it into `FSM.Restore`.
   - `TestReceiveSnapshotStream_LegacyFallbackWhenNoFSMSnapDir` pins the unchanged legacy path: when `fsmSnapDir` is empty, payload remains inline so existing tests keep passing.

## Test plan

- [x] `go test -race -count=1 -short ./internal/raftengine/etcd` — 11.5s, all green
- [x] New token-path test exercises the full sender-FSM → receive → token → `restoreSnapshotState` → receiver-FSM round-trip
- [ ] After merge: deploy to 192.168.0.x cluster, verify 213 receives a fresh snapshot, `applied_index` advances, container memory peak during receive stays well under 2.5 GiB limit (the original OOM hazard)

## Production incident context — 2026-05-08

Two followers (211, 213) fell behind the leader's log during an earlier OOM cascade, leader truncated past their match indices, snapshot transfer became required. With the pre-PR-746 1 GiB receive cap they failed mid-stream → leader retry loop → 100 MB/s outbound for hours.

PR #746 fixed the on-disk cap. This PR closes the remaining memory hazard so the same followers don't OOM during apply once the receive completes — the cluster genuinely catches up rather than trading one loop for another.

## Follow-up (separate PRs)

- Extend the streaming-token path to the leader-side `applyBridgeMode` legacy path so an upgrade in progress (some peers on the pre-token binary) doesn't fall back to materialization on the leader.
- Add a Prometheus counter for token vs inline receive so an operator can detect a regression where a peer hits the legacy fallback in production.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed potential snapshot file leaks when snapshot operations fail and retry
  * Improved startup cleanup to remove stale snapshot files

* **New Features**
  * Enhanced snapshot streaming with optimized disk-based transfer handling
  * Added support for token-based snapshot format for improved storage efficiency

### DIFF
--- a/internal/raftengine/etcd/grpc_transport.go
+++ b/internal/raftengine/etcd/grpc_transport.go
@@ -683,28 +683,25 @@ func (t *GRPCTransport) handle(ctx context.Context, msg raftpb.Message) error {
 	return errors.WithStack(handler(ctx, msg))
 }
 
-func (t *GRPCTransport) receiveSnapshotStream(stream pb.EtcdRaft_SendSnapshotServer) (raftpb.Message, error) {
-	var metadata raftpb.Message
-	seenMetadata := false
+// snapshotSpoolPlacement returns (spoolDir, fsmSnapDir) under the transport
+// lock. When fsmSnapDir is wired, the spool itself is placed inside it so
+// FinalizeAsFSMFile's rename stays intra-filesystem and cannot fail with
+// EXDEV. Standard engine wiring puts both under cfg.DataDir, but the
+// receive code should not assume that. The legacy fallback path
+// (fsmSnapDir == "") keeps the spool in spoolDir because it never renames
+// — Bytes() materializes the payload in place.
+func (t *GRPCTransport) snapshotSpoolPlacement() (placement, fsmSnapDir string) {
 	t.mu.RLock()
-	spoolDir := t.spoolDir
-	fsmSnapDir := t.fsmSnapDir
-	t.mu.RUnlock()
-
-	// Place the spool file inside fsmSnapDir (not spoolDir) when the
-	// streaming-token path is wired, so FinalizeAsFSMFile's os.Rename
-	// stays within a single filesystem and cannot fail with EXDEV. An
-	// operator who mounts spoolDir and fsmSnapDir on different volumes
-	// would otherwise hit a hard receive failure (the leader retries
-	// indefinitely with no chance of success). Standard engine wiring
-	// already puts both under cfg.DataDir, but the receive code should
-	// not assume that. The legacy fallback path (fsmSnapDir empty)
-	// keeps the spool in spoolDir because it never renames — Bytes()
-	// materializes the payload in place.
-	spoolPlacement := spoolDir
+	defer t.mu.RUnlock()
+	fsmSnapDir = t.fsmSnapDir
 	if fsmSnapDir != "" {
-		spoolPlacement = fsmSnapDir
+		return fsmSnapDir, fsmSnapDir
 	}
+	return t.spoolDir, ""
+}
+
+func (t *GRPCTransport) receiveSnapshotStream(stream pb.EtcdRaft_SendSnapshotServer) (raftpb.Message, error) {
+	spoolPlacement, fsmSnapDir := t.snapshotSpoolPlacement()
 	spool, err := newSnapshotSpool(spoolPlacement)
 	if err != nil {
 		return raftpb.Message{}, err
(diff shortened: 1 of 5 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

PR #749, opened 2026-05-09
refactor(kv): plumb caller context through write + verify-leader paths

bootjp:
## Summary

Follow-up to #745. PR #745 capped `verifyLeaderEngine`'s no-context call site at 5s as an incident hotfix; this PR plumbs the caller's `context.Context` end-to-end through the kv write path and the verify-leader path so callers that hold a request deadline see THEIR deadline honoured, not the 5s safety bound.

The 5s bound stays in place as defense-in-depth for callers without an upstream context (LockResolver, HLC lease tick).

## What changed

**Interface signatures (kv-internal):**

```go
type Transactional interface {
    Commit(ctx context.Context, reqs []*pb.Request) (*TransactionResponse, error)
    Abort(ctx context.Context, reqs []*pb.Request) (*TransactionResponse, error)
}

type Coordinator interface {
    // …
    VerifyLeader(ctx context.Context) error
    VerifyLeaderForKey(ctx context.Context, key []byte) error
}

type LeaderProbe interface {
    IsVerifiedLeader(ctx context.Context) bool
}
```

**Plumbing chains** (every step now carries ctx):

```
Dispatch path:
  ShardedCoordinator.Dispatch(ctx)
    → dispatchTxn(ctx) / dispatchSingleShardTxn(ctx) /
      commitPrimaryTxn(ctx) / commitSecondaryTxns(ctx) /
      abortPreparedTxn(ctx) / broadcastToAllGroups(ctx) /
      dispatchDelPrefixBroadcast(ctx)
    → Transactional.Commit/Abort(ctx)
    → leaseRefreshingTxn / LeaderProxy / TransactionManager / ShardRouter
    → applyRequests(ctx) → proposer.Propose(ctx)
    → verifyLeaderEngineCtx(ctx)

Lock-resolver / shard-store path:
  LockResolver.resolveExpiredLock(ctx)
    → applyTxnResolution(ctx) / applyScanLockResolutions(ctx) /
      tryAbortExpiredPrimary(ctx)
    → g.Txn.Commit(ctx)

Healthz:
  S3 / SQS / DynamoDB /healthz/leader handlers
    → isVerifiedSQSLeader(ctx, …) / isVerifiedDynamoLeader(ctx, …) /
      S3Server.isVerifiedS3Leader(ctx) /
      DistributionServer.verifyCatalogLeader(ctx)
    → coordinator.VerifyLeader(ctx) / VerifyLeaderForKey(ctx, key)

Admin LeaderProbe:
[...]

PR #752, opened 2026-05-10
refactor(redis): bound rangeList / readValueAt VerifyLeaderForKey to per-call ctx

bootjp:
## Summary
- Redis adapter follow-up to PR #749. `rangeList` and `readValueAt` now accept a `ctx context.Context` parameter and forward it to `VerifyLeaderForKey` instead of using the long-lived `r.handlerContext()`.
- `lrange` wraps `r.handlerContext()` with `redisDispatchTimeout` and passes it down. Inside MULTI/EXEC, `txnContext` carries the EXEC-scoped dispatch ctx (set at construction in `runTransaction`) so `load() → readValueAt()` respects the caller's deadline.
- Not a correctness fix: the `verifyLeaderEngineCtx` default-deadline guard introduced in PR #749 r1 already bounded these paths at 5 s. This brings them under the same per-command dispatch budget as the rest of the adapter so a slow leader probe can't outlive the user-visible command timeout via the 5 s fallback.

## Background
PR #749 review (gemini/Codex) flagged that `keys()` and `FLUSHDB` had escaped the dispatch deadline by feeding `r.handlerContext()` (server baseCtx, no deadline) into `VerifyLeader*`. Those were fixed inline. Two remaining call sites — `rangeList` (LRANGE leader path) and `readValueAt` (single-key reads from EXEC bodies and string GETs) — were called out as a non-blocking follow-up because the new `verifyLeaderEngineCtx` 5 s default already protected against the unbounded-wait regression. This PR tightens them to match.

## Why `txnContext.ctx`
`txnContext` is the value type that wraps one in-flight EXEC body. `load()` is called repeatedly during EXEC and may transitively call `readValueAt()`. Plumbing ctx through each method would mean threading it through every `working[…]` access — the struct is exactly the right scope to carry the EXEC dispatch ctx, so it gets a `ctx` field with a `//nolint:containedctx` annotation explaining the rationale.

## Test plan
- [x] `go test -race -count=1 -short ./adapter` — 531s, all green
- [x] `go vet ./...` — clean
- [ ] CI green

## Summary by CodeRabbit

* **Bug Fixes**
  * Transactions now honor per-call timeouts during execution, preventing unexpected hangs.
  * List/lookup commands use deadline-aware leader checks for more consistent timeout behavior.

* **Tests**
  * Test updated to explicitly handle value reads with a bound context to ensure reliable timeout semantics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/752)

### DIFF
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -1876,7 +1876,16 @@ func (t *txnContext) load(key []byte) (*txnValue, error) {
 		tv.ttl = ttl
 	} else {
 		var err error
-		val, err = t.server.readValueAt(t.ctx, storageKey, t.startTS)
+		// Some redis_txn_test.go fixtures build a minimal txnContext
+		// literal without setting ctx; fall back to Background so
+		// readValueAt's coordinator.VerifyLeaderForKey does not panic
+		// when wrapped via context.WithTimeout(nil, …). Same defensive
+		// pattern as streamDeletions / loadListState.
+		ctx := t.ctx
+		if ctx == nil {
+			ctx = context.Background()
+		}
+		val, err = t.server.readValueAt(ctx, storageKey, t.startTS)
 		if err != nil && !errors.Is(err, store.ErrKeyNotFound) {
 			return nil, errors.WithStack(err)
 		}
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -1761,7 +1761,13 @@ type txnValue struct {
 }
 
 type txnContext struct {
-	server     *RedisServer
+	server *RedisServer
+	// ctx is the per-EXEC dispatch context (redisDispatchTimeout-bounded
+	// at the call site in runTransaction). Plumbed through so reads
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

Round-1: addressed CodeRabbit minor at adapter/redis.go:1770 — nil-safe `t.ctx` fallback in `txnContext.load` (mirrors the streamDeletions defensive pattern for test fixtures that build a minimal txnContext literal).

Gemini's two medium comments suggest threading ctx through `keyTypeAt` / `resolveListMeta` / `fetchListRange` / `store.GetAt` / `hasExpired`. That's a wider scope than this PR's tightening of leader-verify ctx (the original goal), and the existing `verifyLeaderEngineCtx` default-deadline guard from PR #749 r1 still bounds those paths at 5 s. Leaving full-path ctx plumbing as a follow-up audit item per the original PR description.

PR #753, opened 2026-05-11
fix(redis): prefix transient-leader error replies with NOTLEADER

bootjp:
## Summary
- Centralize a small rewrite at `redisMetricsConn.WriteError` that prefixes transient-leader error replies (e.g. `leader not found`, `not leader`, `leadership lost`, `leadership transfer in progress`) with `NOTLEADER ` on the Redis wire.
- Replies already prefixed with a semantic Redis error code (`ERR`, `WRONGTYPE`, `NOTLEADER`, `MOVED`, `ASK`, `READONLY`, ...) are passed through unchanged.
- New unit test `TestNormalizeRedisErrorReply` covers the prefix/no-op/empty cases.

## Why
Carmine (and through it `jepsen-io/redis`) classifies Redis error replies by their first whitespace-delimited token, converted to a keyword. The upstream `with-exceptions` macro catches `:err`, `:moved`, `:nocluster`, `:clusterdown`, **`:notleader`**, `:timeout` and maps them to clean `:fail` / `:info` ops.

Our adapter was emitting bare `\"leader not found\"` / `\"not leader\"` strings. Carmine parsed those as `:prefix :leader`, which no upstream catch handles, so the Jepsen Redis worker crashed with `IllegalArgumentException: Don't know how to create ISeq from: clojure.lang.ExceptionInfo` instead of recording a retryable failure. Visible on the scheduled stress run https://github.com/bootjp/elastickv/actions/runs/25644042060 where multiple workers crashed in the post-cluster-launch window before leader election settled, taking the whole job down.

Suffix matching mirrors `kv.hasTransientLeaderPhrase`, so wrapping at higher layers (cockroachdb/errors `%w`, gRPC status, etc.) does not defeat the rewrite, and adding a new transient-leader sentinel only requires updating one place.

## Self-review
1. **Data loss** — no change to durability paths; only wire-format of error replies.
2. **Concurrency / distributed failures** — the wrapper runs synchronously per WriteError on the dispatched command; no shared state, no new lock.
3. **Performance** — one `IndexByte` + at most one `ToLower` + four `HasSuffix` calls per error reply (error path only).
4. **Data consistency** — wire format only; sentinels and error chains in Go are unchanged. `hasTransientLeaderPhrase` still classifies `\"NOTLEADER leader not found\"` correctly via `HasSuffix(\"leader not found\")`.
5. **Test coverage** — table-driven `TestNormalizeRedisErrorReply` covers all branches and the no-op cases.

## Test plan
- [x] `go test -race -count=1 -short ./adapter` — 569s, all green
- [x] `go vet ./adapter/` — clean
- [ ] CI green
- [ ] Next scheduled Jepsen run progresses past the leader-election bootstrap window without crashing workers

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized Redis error responses so transient leader/unavailable conditions are consistently surfaced with a NOTLEADER prefix; normalized some “leader redis address unknown” errors to include the ERR prefix.

* **Tests**
  * Added tests verifying detection of leader-related errors (including wrapped forms) and correct NOTLEADER/ERR response behavior.

* **Chores**
  * CI: Cache Maven/Leiningen artifacts and add a warmed dependency step with retries to stabilize scheduled Jepsen runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/753)

### DIFF
--- a/.github/workflows/jepsen-test-scheduled.yml
+++ b/.github/workflows/jepsen-test-scheduled.yml
@@ -57,12 +57,48 @@ jobs:
           curl -L https://raw.githubusercontent.com/technomancy/leiningen/stable/bin/lein > ~/lein
           chmod +x ~/lein
           ~/lein version
+      # Cache Maven artifacts across runs. Maven Central (repo1.maven.org)
+      # serves intermittent 429s during dependency download, which
+      # uncached scheduled runs trip frequently — the lein project pulls
+      # ~100 jars at first launch. Keying on project.clj invalidates the
+      # cache only when dependencies actually change.
+      - name: Cache Maven and Leiningen artifacts
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.m2/repository
+            ~/.lein
+          key: ${{ runner.os }}-maven-${{ hashFiles('jepsen/project.clj') }}
+          restore-keys: |
+            ${{ runner.os }}-maven-
       - name: Pre-fetch Go modules
         run: |
           mkdir -p "$GOCACHE" /tmp/go-tmp
           GOPATH=$(go env GOPATH)
           export GOCACHE GOTMPDIR=/tmp/go-tmp
           go mod download
+      # Warm the Maven cache up front so unit tests / workloads start
+      # against a populated ~/.m2. `lein deps` is wrapped with a retry
+      # loop because Maven Central transiently returns 429 during peak
+      # hours, and the default lein behaviour is to fail the build on
+      # the first checksum miss. Three attempts with linear backoff is
+      # enough to ride out a typical rate-limit window.
+      - name: Warm Leiningen Maven cache
+        working-directory: jepsen
+        run: |
+          set -uo pipefail
+          n=0
+          max=3
+          until ~/lein deps; do
+            n=$((n + 1))
+            if [ "$n" -ge "$max" ]; then
+              echo "lein deps failed after $n attempts" >&2
+              exit 1
+            fi
+            sleep_secs=$((n * 30))
+            echo "lein deps failed (attempt $n/$max), retrying in ${sleep_secs}s..." >&2
+            sleep "$sleep_secs"
+          done
       - name: Run Jepsen unit tests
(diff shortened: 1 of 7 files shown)


bootjp:
@claude review

Round-2: caller audit caught additional sites in `adapter/redis_compat_commands.go` (30) and `adapter/redis_lua.go` (6) that bypassed `writeRedisError`. Now routed through the helper. All adapter race-short tests green (564s), vet clean.

bootjp:
@claude review

Round-3: addressed your finding on `redis.go:1737`. Grep audit found 5 more sites that emit the same "leader redis address unknown for %s" wire string via `errors.Newf` (proxyKeys / proxyLRange / proxyRPush / proxyLPush x2 / leaderClientForKey). All six now prefix with `ERR ` so Carmine maps to `:prefix :err` (caught by upstream with-exceptions). NOTLEADER would be wrong here — these are config-gap errors with the raft leader known but the raft→redis address mapping missing.

Caller audit: no errors.Is consumer reads the message; no test / classifier / business-logic branch grep'd "leader redis address unknown" — wire-format-only change.

go vet clean, targeted tests green. Full `./adapter` race-short re-run pending.

bootjp:
@claude review

Round-4: thanks for catching the redis_proxy.go gap I missed in r3's grep audit. Applied `ERR ` prefix to all 4 sites (`proxyDBSize`/`proxyDel`/`proxyFlushDatabase`/`proxyFlushLegacy`) at the `errors.Newf` source. Added regression test `leader-address-unknown config-gap is already ERR-prefixed` per CLAUDE.md's test-with-fix convention.

Caller audit: 4 call sites (DBSIZE/DEL/FLUSHDB/FLUSHLEGACY) all flow through writeRedisError; no consumer pattern-matches err.Error(); wire-format-only change. Confirmed via grep across adapter/, kv/, store/, internal/.

Branch was rebased onto main (PR #752 merge + others). go vet clean, targeted tests green.

PR #758, opened 2026-05-15
backup: Redis set encoder (Phase 0a)

bootjp:
## Summary
- Adds the `!st|meta|` + `!st|mem|` → `sets/<key>.json` encoder per the Phase 0 design doc (`docs/design/2026_04_29_proposed_snapshot_logical_decoder.md`).
- Members are emitted as a sorted byte-order array (not a JSON object) so binary-safe member bytes round-trip via the typed `{"base64":"..."}` envelope without colliding under percent-encoded JSON object keying — same shape as the hash encoder's `fields` array.
- Duplicate `HandleSetMember` calls collapse via a `map[string]struct{}` buffer; Redis sets are mathematical sets, so a snapshot iterator that re-emits an `!st|mem|` key is harmless.
- TTL records on `!redis|ttl|` for a set user key fold into the set JSON's `expire_at_ms` field via the `HandleTTL` switch — same policy as hash + list encoders, no separate sidecar.
- The `uint64 → int64` overflow guard applied symmetrically to the hash + list encoders (see PR #755 round 2) is also enforced here on the set's declared-length field.
- `!st|meta|d|` (LenDelta) records are silently skipped: `!st|mem|` keys are the source of truth at backup time.

## Test plan
- [x] `go test -race ./internal/backup/...` — all 11 new set tests + existing hash/list/string suites pass (1.4s)
- [x] `golangci-lint run ./internal/backup/...` — 0 issues
- [x] Coverage: sorted byte-order round-trip; empty set still emits a file (SCARD==0 observable); TTL inlined from scan index; length-mismatch warning shape; binary member base64 envelope; meta-delta key silently skipped; malformed meta length rejected; overflow guard rejected; members-without-meta path; duplicate members collapse; parser-level delta-key rejection; `math.MaxInt64` boundary accepted.

## Self-review (5 lenses)
1. **Data loss**: encoder writes via `writeFileAtomic` (tmp+rename). Only meta-deltas are skipped, which is consistent with the hash/list policy and the live read path's source-of-truth invariant.
2. **Concurrency / distributed**: `RedisDB.Handle*` methods documented not goroutine-safe; decoder pipeline is sequential per scope. No shared mutable state.
3. **Performance**: members buffered in `map[string]struct{}`; single allocation per member. Sort at flush is O(n log n); set member count bounded by `maxWideColumnItems` on the live side.
4. **Data consistency**: `redis_set_length_mismatch` warning fires on declared-vs-observed drift; sorted byte order is deterministic across runs.
5. **Test coverage**: 11 named tests cover every public handler plus the parse-error paths, the new overflow guard, and the duplicate-members idempotency contract.

## Phase 0a remaining after this PR
- `redis_zset.go` (`!zs|...`) encoder — sorted-set with `(member, score)` records, JSON shape `{"format_version": 1, "members": [{"member":..., "score":...}], "expire_at_ms": null}`
- `redis_stream.go` (`!stream|...`) encoder — JSONL output (`streams/<key>.jsonl` with `_meta` trailer)
- `cmd/elastickv-snapshot-decode/` CLI binary
- `cmd/elastickv-snap-token` helper
- `docs/operations/snapshot_restore.md` runbook

### DIFF
--- a/internal/backup/redis_hash.go
+++ b/internal/backup/redis_hash.go
@@ -20,11 +20,6 @@ const (
 	RedisHashMetaPrefix      = "!hs|meta|"
 	RedisHashFieldPrefix     = "!hs|fld|"
 	RedisHashMetaDeltaPrefix = "!hs|meta|d|"
-
-	// hashUserKeyLenSize is the fixed BE-uint32 width of the
-	// per-key length prefix used by every wide-column key shape.
-	// Mirrors store/wideColKeyLenSize.
-	hashUserKeyLenSize = 4
 )
 
 // ErrRedisInvalidHashMeta is returned when the !hs|meta| value is not
@@ -165,7 +160,7 @@ func parseHashFieldKey(key []byte) ([]byte, []byte, bool) {
 	if !ok {
 		return nil, nil, false
 	}
-	fieldName := rest[hashUserKeyLenSize+len(userKey):]
+	fieldName := rest[wideColumnUserKeyLenSize+len(userKey):]
 	return userKey, fieldName, true
 }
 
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

Round 1 — gemini medium (review 4301424038): hash-specific `hashUserKeyLenSize` constant was leaking into the set encoder via the shared `parseUserKeyLenPrefix` helper. Cross-encoder naming gets more misleading with each new wide-column type.

Renamed to `wideColumnUserKeyLenSize` and moved to `redis_string.go` (foundation file). All 7 references in `redis_hash.go` (5) and `redis_set.go` (1) migrated. Docstring now describes the cross-encoder reuse explicitly.

No functional change — same value (4 bytes), identifier-only refactor. `go test -race ./internal/backup`, `go vet`, `golangci-lint` all clean.

Gemini's second medium ("ensure HandleTTL / Finalize handle all wide-column types correctly") was already addressed in the original PR — `redisKindSet` is routed in the `HandleTTL` switch and `flushSets` is wired into `Finalize`. No additional change needed.

PR #778, opened 2026-05-17
feat(encryption): Stage 6C-1 — §9.1 startup-refusal guards (flag + sidecar state)

bootjp:
## Summary

Stage 6C-1 per the [PR #762 plan](https://github.com/bootjp/elastickv/pull/762). Decomposes the original monolithic Stage 6C into four sub-milestones (6C-1 → 6C-4) per dependency footprint, then ships 6C-1: the flag + sidecar-state §9.1 startup-refusal guards.

The Stage 6B-2 [`encryptionMutatorsEnabled`](https://github.com/bootjp/elastickv/blob/main/main_encryption_admin.go) triple gate is the RPC-boundary safety net; this PR adds the process-boundary safety net on top.

## Stage 6C sub-decomposition (design doc update)

The original Stage 6C bundled guards with three different dependency footprints. Shipping the load-bearing downgrade-prevention guard (`ErrSidecarPresentWithoutFlag`) alone would have been blocked on Stage 7's writer-registry work (the `local_epoch` rollback peer). The sub-decomposition lets 6C-1 ship now and unblocks Stage 6D once 6C-1 + 6C-2 land:

| Sub-milestone | Scope | Status |
|---|---|---|
| **6C-1 (this PR)** | Flag + sidecar-state guards | this PR |
| 6C-2 | Raftengine integration (sidecar/raft index gap, fsync probe, `local_epoch` exhaustion) | open |
| 6C-3 | Membership-view guards (`node_id` collision, `local_epoch` rollback) — bundles with Stage 6D | open |
| 6C-4 | Phase-2-specific guards — bundles with Stage 6E | open |

### Three new §9.1 startup-refusal sentinels

| Sentinel | Catches |
|---|---|
| `ErrSidecarPresentWithoutFlag` | Sidecar on disk but `--encryption-enabled` off — would silently downgrade new writes to cleartext |
| `ErrKEKRequiredWithFlag` | `--encryption-enabled` set but no `--kekFile` — flag-on / KEK-off is misconfigured at the operator-intent level |
| `ErrKEKMismatch` | KEK loaded but at least one wrapped DEK fails to unwrap — classic "wrong --kekFile from a different cluster" error |

`ErrKEKMismatch` annotates the offending `key_id` and `purpose` so runbooks can grep `purpose="raft"` / `purpose="storage"` to identify the failing DEK.

### `internal/encryption/startup.go` (new)

```go
type StartupConfig struct {
    EncryptionEnabled bool
    KEKConfigured     bool
[...]

PR #762, opened 2026-05-16
docs(encryption): Stage 6 plan — decompose into 6A-6F + record 5D shipped

bootjp:
## Summary

- Record Stage 5D (PR #760, merged 2026-05-16T16:35:03Z) as shipped in the milestone table.
- Mark Stage 5E (§5.6 step 1a capability fan-out helper) as **deferred** — its scope is narrowly the auto-batch CLI convenience and does not gate any Stage 6 work.
- Decompose the original Stage 6 row (`3-phase rollout flags + applier wiring`) into six sub-milestones **6A–6F**, each independently reviewable, leaving the cluster in a production-safe state, unblocking the next.
- Add a Stage 6 sub-milestone rationale section explaining why the decomposition follows this shape — why 6A is the load-bearing safety boundary (it lands the §6.3 applier whose absence was PR #760 round-3 codex P1), why 6D/6E must be separate cutovers (intermediate Phase-1 storage-only state is a valid operator dwell), and why 6F is the last ergonomics PR.

## Stage 6 breakdown

| Sub-milestone | Scope |
|---|---|
| **6A** | §6.3 `EncryptionApplier` concrete implementation (writer-registry insert + sidecar mutate + keystore update) + `kv.NewKvFSMWithHLC(... WithEncryption(applier))` wiring + re-enable mutating-RPC wiring in `registerEncryptionAdminServer` (gated on `--encryption-enabled`) |
| **6B** | §6.5 KEK plumbing — `--kekFile` / `--kekUri` flags + `internal/encryption/kek` selection at startup + applier KEK-unwrap path |
| **6C** | §6.5 `--encryption-enabled` + §9.1 startup refusal guards (sidecar present without flag, KEK mismatch, sidecar/raft index gap, fsync support, `local_epoch` rollback / exhaustion, `node_id` collision) |
| **6D** | §6.6 `enable-storage-envelope` admin RPC + §7.1 Phase-1 storage cutover (§6.2 toggle ON) + Voters ∪ Learners capability gate |
| **6E** | §6.6 `enable-raft-envelope` admin RPC + §7.1 Phase-2 raft cutover + `raft_envelope_cutover_index` sidecar record + `internal/raftengine/etcd/engine.go` `applyNormalEntry` unwrap hook + `ErrRaftUnwrapFailed` HaltApply path |
| **6F** | §6.5 `--encryption-rotate-on-startup` request flag + leader-elected rotation proposal |

## Why doc-first

[...]

PR #781, opened 2026-05-17
feat(encryption): Stage 6C-2 — local_epoch exhaustion guard + filesystem startup probe

bootjp:
## Summary

Stage 6C-2 per the [PR #762 plan](https://github.com/bootjp/elastickv/pull/762) and the 6C-1 → 6C-4 sub-decomposition landed in [PR #778](https://github.com/bootjp/elastickv/pull/778).

Ships the **process-local subset** of the §9.1 startup-refusal guards — those that read only the encryption package's own state and don't need raftengine integration. The remaining `ErrSidecarBehindRaftLog` guard (which needs the raftengine's persisted applied-index) is split into a new **6C-2b** sub-milestone for a dedicated follow-up PR.

Also bundles the gemini-r3 medium parseErr-annotation fix deferred from PR #778.

### Two new §9.1 startup-refusal guards

| Sentinel | Catches |
|---|---|
| `ErrLocalEpochExhausted` | Active DEK has `local_epoch == 0xFFFF` — one bump from GCM nonce reuse. Refuses to start until rotated. |
| `ErrUnsupportedFilesystem` (via new `ProbeSidecarFilesystem`) | Filesystem doesn't actually honor `dir.Sync` (NFS, some FUSE). Surfaces BEFORE the first encryption-relevant Raft entry, not on first `WriteSidecar` call hours into operation. |

### Gemini r3 medium fix (deferred from PR #778)

`ErrKEKMismatch`'s defensive non-decimal-`key_id` branch now includes `parseErr` alongside the unwrap error in the annotation. An operator triaging a manually-edited sidecar can see BOTH the unwrap failure AND the malformed `key_id` reason in a single log line.

### New exported helper

```go
func ProbeSidecarFilesystem(sidecarPath string) error
```

Exercises the §5.1 `write + fsync + rename + dir.Sync` sequence on a sentinel file in the sidecar's parent directory, then removes it. Side-effect-free on success. Returns `ErrUnsupportedFilesystem`-wrapped on any fsync/dir.Sync failure (matching the existing `WriteSidecar` path so `errors.Is` works identically across both).

### Guard ordering in `CheckStartupGuards`

```
1. guardSidecarWithoutFlag       (6C-1, downgrade prevention)
2. guardKEKRequired              (6C-1, fail-fast)
3. guardKEKMatchesSidecar        (6C-1, wrong-KEK catch)
4. guardLocalEpochExhausted      (6C-2, GCM nonce-reuse prevention) ←NEW
5. guardFilesystemDurability     (6C-2, NFS/FUSE catch)             ←NEW
```

Each step refuses on the most specific failure first, so the operator's log line points at the actual cause rather than a downstream symptom.

## Design doc updates

- **6C-1** row marked shipped (PR #778)
- **6C-2** row rewritten to reflect the process-local-only scope of this PR
- **6C-2b** row added — `ErrSidecarBehindRaftLog` is split out because it needs raftengine integration (outside the encryption package boundary)
- Rationale section updated with the 6C-2 / 6C-2b split + shipping order

## Why split `ErrSidecarBehindRaftLog` into 6C-2b

It needs the raftengine's persisted applied-index, which requires plumbing the engine into the startup-guard path. That's load-bearing cross-package work distinct from the encryption-package-only guards in this PR. Keeping them in separate PRs lets reviewers focus on one concern at a time and gets the simpler guards shipped sooner.

## Caller audit

- No public function signatures changed.
- `CheckStartupGuards` gains two new internal guard calls; the `StartupConfig` input shape is unchanged.
- `ProbeSidecarFilesystem` is new and has no existing callers.
- `ErrKEKMismatch` format-string gains a `(parse error: %v)` segment in the defensive branch only — annotation-only change, callers that only inspect `errors.Is(err, ErrKEKMismatch)` see no behavior change.

## Five-lens self-review

1. **Data loss** — net-positive. `ErrLocalEpochExhausted` prevents GCM nonce reuse (reveals plaintext XOR via the keystream). The filesystem probe catches a misconfiguration that would silently degrade the §5.1 crash-durability guarantee. Both run before any encryption-relevant Raft entry can commit.
2. **Concurrency / distributed failures** — both guards run at process startup before any goroutine fan-out. `ProbeSidecarFilesystem` uses `os.CreateTemp + os.Rename + dir.Sync`, all process-local. No shared state.
3. **Performance** — single sidecar read + single tmpfile sync/rename/sync per process start. Zero hot-path impact. Probe is gated on `--encryption-enabled`.
4. **Data consistency** — `ErrLocalEpochExhausted` catches the most severe nonce-reuse failure mode in the §4.1 set. The filesystem probe ensures that when `WriteSidecar` IS called later, `dir.Sync`'s durability guarantee is real.
5. **Test coverage** — 7 new tests cover the boundary (`0xFFFE` pass / `0xFFFF` fire), retired-key skip, probe-OK, probe-missing-dir, gate-off skip, and the deferred parseErr fix. All run with `-race`.

## Test plan

- [x] `go test -race -timeout=120s ./internal/encryption/... .` — PASS
- [x] `go build ./...` — PASS
- [x] `golangci-lint run ./internal/encryption/...` — 0 issues
- [ ] Full Jepsen suite — not run; no envelope cutover happens until Stage 6D and this PR only adds startup refusal logic that runs before any encryption-relevant Raft entry

## Plan

- **6C-2b** — `ErrSidecarBehindRaftLog` (needs raftengine integration)
- **6C-3** — `ErrNodeIDCollision` + `ErrLocalEpochRollback` (bundles with Stage 6D capability fan-out)
- **6D** — `enable-storage-envelope` admin RPC + §7.1 Phase-1 cutover (unblocked once 6C-2b lands)

### DIFF
--- a/docs/design/2026_04_29_partial_data_at_rest_encryption.md
+++ b/docs/design/2026_04_29_partial_data_at_rest_encryption.md
@@ -20,8 +20,9 @@ Date: 2026-04-29
 | 5E | §5.6 step 1a capability fan-out helper + `encryption bootstrap --discover-from=...` auto-batch mode | deferred | — |
 | 6A | §6.3 `EncryptionApplier` concrete implementation (writer-registry insert + sidecar mutate + keystore update; nil-KEK code paths in `ApplyBootstrap` / `ApplyRotation` return a typed `ErrKEKNotConfigured` defense-in-depth marker until 6B fills it) + `kv.NewKvFSMWithHLC(... WithEncryption(applier))` wiring. **Mutator wiring in `registerEncryptionAdminServer` stays OFF, identical to Stage 5D** — see Stage 6 rationale §6A note. No new flag is introduced in 6A; the applier is operator-inert in a same-version 6A cluster (exercised only via direct `FSM.Apply` unit tests). Rolling 6A→6B upgrade has a halt-on-foreign-entry caveat — see rationale | open | — |
 | 6B | §6.5 KEK plumbing — `--kekFile` / `--kekUri` flags + `internal/encryption/kek` selection at startup + thread `KEKUnwrapper` into the applier so `ApplyBootstrap` / `ApplyRotation` actually KEK-unwrap (replaces the 6A `ErrKEKNotConfigured` stubs) + introduce `--encryption-enabled` flag (default off) + re-enable mutating-RPC wiring in `registerEncryptionAdminServer` gated on (`--encryption-enabled` AND `KEKConfigured()`) | open | — |
-| 6C-1 | Subset of §9.1 startup refusal guards that depend only on flags + sidecar state: sidecar present without `--encryption-enabled` (`ErrSidecarPresentWithoutFlag`, downgrade prevention), `--encryption-enabled` without `--kekFile` (`ErrKEKRequiredWithFlag`, fail-fast), KEK loaded but sidecar wrapped DEKs do not unwrap (`ErrKEKMismatch`, operator-error catch). `internal/encryption/startup.go`'s `CheckStartupGuards` runs once in `main.go run()` before `buildShardGroups`. | open | — |
-| 6C-2 | Remaining §9.1 guards that depend on raftengine integration or shared state: `ErrSidecarBehindRaftLog` (needs the persisted applied-index from raftengine), `ErrUnsupportedFilesystem` startup probe (the current code surfaces it only on first write — a dedicated `tmpfile + dir.Sync` probe at startup would surface it before any encryption-relevant write), `ErrLocalEpochExhausted` (active DEK `local_epoch == 0xFFFF` — coupled with the writer-registry record so shipping it without the rollback peer creates an asymmetric posture future readers must reason around). | open | — |
+| 6C-1 | Subset of §9.1 startup refusal guards that depend only on flags + sidecar state: sidecar present without `--encryption-enabled` (`ErrSidecarPresentWithoutFlag`, downgrade prevention), `--encryption-enabled` without `--kekFile` (`ErrKEKRequiredWithFlag`, fail-fast), KEK loaded but sidecar wrapped DEKs do not unwrap (`ErrKEKMismatch`, operator-error catch). `internal/encryption/startup.go`'s `CheckStartupGuards` runs once in `main.go run()` before `buildShardGroups`. | shipped | PR #778 |
+| 6C-2 | Process-local §9.1 guards that read only the encryption package's own state: `ErrLocalEpochExhausted` (any active DEK with `local_epoch == 0xFFFF` — refuses to start until rotated, preventing GCM nonce reuse on the next write), `ErrUnsupportedFilesystem` startup probe (`ProbeSidecarFilesystem` exercises the §5.1 write+rename+dir.Sync sequence on a sentinel file in the sidecar directory at startup so filesystem-capability failures surface BEFORE the first encryption-relevant Raft entry, not on the first WriteSidecar call hours into operation). Also includes the gemini-r3 medium parseErr-annotation fix deferred from PR #778 (`ErrKEKMismatch` now wraps both `parseErr` and the unwrap error when the defensive non-decimal-key_id branch fires). The asymmetric `ErrLocalEpochRollback` peer is split into 6C-3 because it requires the writer-registry record (Stage 7). The `ErrSidecarBehindRaftLog` guard moves to 6C-2b because it requires raftengine integration — outside the encryption package boundary. | open | — |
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

Stage 6C-2 opening review. Scope: 2 new §9.1 startup-refusal guards (`ErrLocalEpochExhausted`, `ErrUnsupportedFilesystem` via new `ProbeSidecarFilesystem` helper) + the gemini-r3 parseErr-annotation fix deferred from PR #778. Built on top of just-merged PR #778 (Stage 6C-1).

The original monolithic Stage 6C-2 was further split: `ErrSidecarBehindRaftLog` moved to a new 6C-2b because it needs raftengine integration that's outside the encryption package boundary. Design doc updated to track the split.

- `ErrLocalEpochExhausted` catches the GCM nonce-reuse footgun: an active DEK at local_epoch=0xFFFF is one bump from rolling back to 0 and re-issuing a nonce already in use
- Boundary discipline: 0xFFFE must pass (one nonce still available), 0xFFFF must fire
- Retired DEK with exhausted local_epoch must be silently ignored (no new write allocates a nonce under it)
- `ProbeSidecarFilesystem` is side-effect-free on success (no sentinel file remains)
- Missing parent dir surfaces as a config error, NOT wrapped with ErrUnsupportedFilesystem (the two failure classes are distinct)
- Probe is gated on --encryption-enabled (pre-rollout cluster on NFS is no worse off than today)
- ErrKEKMismatch parseErr-included branch maintains the fail-closed semantics, just adds the parseErr to the annotation

bootjp:
@claude review

Round-1 (commit `716fd881`): both r1 findings addressed.

## claude r1 [Medium] addressed

Refactored `CheckStartupGuards` to cache the parsed `*Sidecar` once via the new `loadSidecarForGuards` helper, then pass it into `guardKEKMatchesSidecar` and `guardLocalEpochExhausted` by pointer. The two consuming guards' signatures changed from `(cfg, sidecarPresent bool)` to `(cfg, sc *Sidecar)`; `sc == nil` maps to the same "no sidecar to consult" path that `!sidecarPresent` did.

Updated the godoc to explicitly call out the single-read invariant and why it matters for 6C-2b / 6C-3 / 6C-4: shipping each new guard with its own `ReadSidecar` call would silently grow startup IO from O(1) to O(N guards), and a partial-write race between the guards' reads is harder to reason about than a single snapshot shared across all of them.

## claude r1 [Nit] addressed

Replaced the misleading defer comment in `ProbeSidecarFilesystem` with an accurate note: the defer removes whichever file `tmpPath` currently names (either CreateTemp output or the renamed `.final`), and stale `.final` files left by a crashed prior probe are NOT cleaned here — they're operator-removable by the `.encryption-probe-` prefix.

## Caller audit

`guardKEKMatchesSidecar` / `guardLocalEpochExhausted` are package-internal — only `CheckStartupGuards` calls them. The single call site is updated in lockstep. `StartupConfig` and `CheckStartupGuards` external surface unchanged. Fail-closed semantics preserved on both guards.

bootjp:
@claude review

Round-2 (commit `76eea5b7`): both r2 findings addressed.

## claude r2 [Nit] + codex P2 [Skip sidecar read when KEK is not configured] addressed

Added `!cfg.KEKConfigured` short-circuit to `loadSidecarForGuards` (the docstring claimed this skip but the code didn't implement it). The load-bearing concern was **error precedence**: with the r1 refactor moving the sidecar read up into `loadSidecarForGuards` (which runs BEFORE `guardKEKRequired`), a corrupt-sidecar + missing-KEK combination silently started returning a sidecar-parse error instead of the operator's primary misconfiguration (`ErrKEKRequiredWithFlag`). The 3-line short-circuit restores the pre-r1 precedence.

`TestCheckStartupGuards_CorruptSidecarMissingKEK_RefusesWithKEKError` pins the guarantee: malformed sidecar (version=0) with KEKConfigured=false MUST return `ErrKEKRequiredWithFlag`, not the sidecar-parse error.

## claude r2 [Cosmetic nit] addressed

`ProbeSidecarFilesystem` named return `retErr` removed (unused — the defer only calls `os.Remove(tmpPath)`, never inspects/assigns the error). Plain `error` return is clearer about the absence of deferred error mutation.

## Stale gemini findings

The two gemini medium findings posted on this PR were on the pre-r1 commit `1a3ced4d`. Both are already addressed by my r1 fix (sidecar caching eliminated the independent ReadSidecar; defer comment was rewritten). Will resolve on next gemini pass.

## Caller audit

`loadSidecarForGuards` is package-internal — signature unchanged, only body. `ProbeSidecarFilesystem` retErr removal is source-compatible. External callers that errors.Is(err, ErrKEKRequiredWithFlag) now see the restored prior behavior on the corrupt-sidecar + missing-KEK path.

PR #782, opened 2026-05-17
feat(encryption): Stage 6C-2b — ErrSidecarBehindRaftLog guard primitive (encryption-side)

bootjp:
## Summary

Stage 6C-2b per the [PR #762 plan](https://github.com/bootjp/elastickv/pull/762) and the 6C sub-decomposition landed in [PR #781](https://github.com/bootjp/elastickv/pull/781).

Ships the **encryption-side primitive** for the `ErrSidecarBehindRaftLog` §9.1 guard so Stage 6D-and-later code paths (e.g., §7.1 capability fan-out's gap-coverage check) can depend on the predicate without waiting for the raftengine-side scanner helper. The actual raftengine WAL-scan implementation + `main.go` startup-guard phase wiring lands in a follow-up **6C-2c** PR.

### One new sentinel

| Sentinel | Catches |
|---|---|
| `ErrSidecarBehindRaftLog` | Partial-write crash: encryption-relevant Raft entry applied to engine but §5.1 sidecar write didn't complete. Recovery: `encryption resync-sidecar`. |

Without this guard the encryption package would silently rebuild keystore state from an outdated wrapped-DEK snapshot, and a fail-closed read would fire on the next post-cutover entry with `unknown_key_id` (halting apply).

### Three new exported primitives in `internal/encryption/audit.go`

```go
// §5.5 predicate
func IsEncryptionRelevantOpcode(opcode byte) bool

// Cross-package contract (raftengine implements in 6C-2c)
type EncryptionRelevantScanner interface {
    HasEncryptionRelevantEntryInRange(startExclusive, endInclusive uint64) (bool, error)
}

// Pure-function guard
func GuardSidecarBehindRaftLog(sidecarAppliedIdx, engineAppliedIdx uint64,
                               scanner EncryptionRelevantScanner) error
```

The encryption package owns the **semantic-level** predicate; `fsmwire` owns the **wire-level** constants (`OpEncryptionMin` / `OpEncryptionMax`). Decoupling means a future widening of the encryption opcode range (Stage 6E reserves 0x06 / 0x07) is a single-line change in `fsmwire`, not a cross-package edit.

### Guard contract (5 branches, all tested)

| Branch | Behavior |
|---|---|
| `sidecarAppliedIdx >= engineAppliedIdx` | Caught up. Return nil. **Scanner NEVER called.** |
| Non-empty gap, scanner says no hit | Return nil. |
| Non-empty gap, scanner says hit | Fire `ErrSidecarBehindRaftLog` with both indices in annotation. |
| Scanner error | Propagate wrapped, **NOT** marked with `ErrSidecarBehindRaftLog` (scanner failure ≠ gap-coverage refusal). |
| `nil` scanner with non-empty gap | **Fail closed** with `ErrSidecarBehindRaftLog` (cannot prove safety). |
| `nil` scanner on caught-up path | Return nil (scanner is irrelevant when there's no gap). |

## Design doc updates

- **6C-2b** row rewritten to reflect the primitive-only scope of this PR (encryption package only, no `main.go` wiring).
- **6C-2c** row added — raftengine implementation of the scanner + `main.go` startup-guard phase wiring.
- Shipping order updated: `6C-1 ✅ → 6C-2 ✅ → 6C-2b (this PR) → 6C-2c → 6D (bundles 6C-3) → 6E (bundles 6C-4)`.
- Rationale section updated with the 6C-2b / 6C-2c split: the primitive lands in 6C-2b so Stage 6D-and-later RPC paths that need the same predicate (capability fan-out gap check) can depend on the encryption package without waiting for 6C-2c.

## Why split 6C-2b into "primitive" vs "integration"

The §5.5 predicate (`IsEncryptionRelevantOpcode`) is needed by more than just the startup guard. Stage 6E's capability fan-out checks the same predicate when refusing a cutover RPC if any peer's sidecar is behind. Shipping the predicate + sentinel + interface without the raftengine WAL-scan implementation lets 6D / 6E proceed against the encryption package boundary without waiting on the cross-package plumbing.

## Caller audit

No public function signatures changed. All four new exports (`IsEncryptionRelevantOpcode`, `EncryptionRelevantScanner`, `GuardSidecarBehindRaftLog`, `ErrSidecarBehindRaftLog`) are **net-additive**. No existing callers to audit.

## Five-lens self-review

1. **Data loss** — net-positive. The primitive enables the §9.1 refusal that catches partial-write crashes between an encryption-relevant Raft commit and the sidecar update. Until 6C-2c wires it into `main.go` the refusal is operator-inert, but the primitive's failure modes (nil scanner → fail closed, scanner error → propagate distinct) are correct.
2. **Concurrency / distributed failures** — pure function, no shared state. Scanner interface is single-call per guard invocation; implementations are free to lock internally.
3. **Performance** — zero hot-path impact. The guard is called once per shard per process start (in 6C-2c); within the guard it's one integer comparison + at most one scanner call.
4. **Data consistency** — `ErrSidecarBehindRaftLog` is the only §9.1 guard that catches the "Raft committed but sidecar missed it" partial-write window. Without it the next post-cutover read would HaltApply with `unknown_key_id`; with it the operator gets a single typed startup refusal pointing at the right recovery runbook.
5. **Test coverage** — 7 new tests cover the predicate (range-member + known-opcode + below/above boundary) AND every branch of the guard (caught-up, gap-not-covered, gap-covered, scanner-error, nil-scanner-fails-closed, nil-scanner-caught-up). The `fakeScanner` exercises the interface contract in isolation from any real raftengine.

## Test plan

- [x] `go test -race -timeout=60s ./internal/encryption/...` — PASS
- [x] `golangci-lint run ./internal/encryption/...` — 0 issues
- [ ] Full Jepsen suite — not run; the primitive has no caller in this PR (operator-inert until 6C-2c)

## Plan

- **6C-2c** — raftengine implementation of `EncryptionRelevantScanner` (WAL-based) + `main.go` startup-guard phase wiring
- **6C-3** — `ErrNodeIDCollision` + `ErrLocalEpochRollback` (bundles with Stage 6D)
- **6D** — `enable-storage-envelope` admin RPC + §7.1 Phase-1 cutover (unblocked once 6C-2c lands)

## Summary by CodeRabbit

* **New Features**
  * Added a safety guard to prevent sidecar lag from silently bypassing encryption-relevant log entries.
  * Added error handling with operator guidance for sidecar synchronization situations.

* **Tests**
  * Added comprehensive test coverage for encryption relevance detection and sidecar lag validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/782?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/design/2026_04_29_partial_data_at_rest_encryption.md
+++ b/docs/design/2026_04_29_partial_data_at_rest_encryption.md
@@ -22,7 +22,8 @@ Date: 2026-04-29
 | 6B | §6.5 KEK plumbing — `--kekFile` / `--kekUri` flags + `internal/encryption/kek` selection at startup + thread `KEKUnwrapper` into the applier so `ApplyBootstrap` / `ApplyRotation` actually KEK-unwrap (replaces the 6A `ErrKEKNotConfigured` stubs) + introduce `--encryption-enabled` flag (default off) + re-enable mutating-RPC wiring in `registerEncryptionAdminServer` gated on (`--encryption-enabled` AND `KEKConfigured()`) | open | — |
 | 6C-1 | Subset of §9.1 startup refusal guards that depend only on flags + sidecar state: sidecar present without `--encryption-enabled` (`ErrSidecarPresentWithoutFlag`, downgrade prevention), `--encryption-enabled` without `--kekFile` (`ErrKEKRequiredWithFlag`, fail-fast), KEK loaded but sidecar wrapped DEKs do not unwrap (`ErrKEKMismatch`, operator-error catch). `internal/encryption/startup.go`'s `CheckStartupGuards` runs once in `main.go run()` before `buildShardGroups`. | shipped | PR #778 |
 | 6C-2 | Process-local §9.1 guards that read only the encryption package's own state: `ErrLocalEpochExhausted` (any active DEK with `local_epoch == 0xFFFF` — refuses to start until rotated, preventing GCM nonce reuse on the next write), `ErrUnsupportedFilesystem` startup probe (`ProbeSidecarFilesystem` exercises the §5.1 write+rename+dir.Sync sequence on a sentinel file in the sidecar directory at startup so filesystem-capability failures surface BEFORE the first encryption-relevant Raft entry, not on the first WriteSidecar call hours into operation). Also includes the gemini-r3 medium parseErr-annotation fix deferred from PR #778 (`ErrKEKMismatch` now wraps both `parseErr` and the unwrap error when the defensive non-decimal-key_id branch fires). The asymmetric `ErrLocalEpochRollback` peer is split into 6C-3 because it requires the writer-registry record (Stage 7). The `ErrSidecarBehindRaftLog` guard moves to 6C-2b because it requires raftengine integration — outside the encryption package boundary. | open | — |
-| 6C-2b | `ErrSidecarBehindRaftLog` startup guard — reads the persisted applied-index from raftengine and refuses if the sidecar's `raft_applied_index` is behind by an interval covering any encryption-relevant Raft entry (`isEncryptionRelevant` predicate per §5.5). Plumbing raftengine into the startup-guard path is the load-bearing scope here; the guard itself is a one-screen function. | open | — |
+| 6C-2b | `ErrSidecarBehindRaftLog` guard PRIMITIVE — encryption-side sentinel, `IsEncryptionRelevantOpcode` predicate (§5.5), `EncryptionRelevantScanner` cross-package interface, and `GuardSidecarBehindRaftLog` pure-function guard that takes both indices + a scanner callback. Tests use a fake scanner to exercise every branch (caught-up / gap-not-covered / gap-covered / scanner-error / nil-scanner-fails-closed). Wires nothing into main.go; the raftengine-side implementation lands in 6C-2c. | shipped-in-this-PR | this PR |
+| 6C-2c | Raftengine implementation of `EncryptionRelevantScanner` (WAL-based scan of Raft entries in the supplied index range against the §5.5 predicate) + `main.go` startup-guard phase that runs after `buildShardGroups` for each shard with encryption enabled, before any gRPC server starts serving. Inherits the encryption-side primitive from 6C-2b. | open | — |
 | 6C-3 | Cluster-wide §9.1 guards that require the Voters ∪ Learners membership view: `ErrNodeIDCollision` (two members hash to the same 16-bit `node_id`), `ErrLocalEpochRollback` (sidecar `local_epoch` ≤ writer-registry record). Naturally bundles with the Stage 6D capability fan-out which already needs that membership view. | open | — |
 | 6C-4 | Phase-2-specific §9.1 guards: `ErrEnvelopeCutoverDivergence` (sidecar `raft_envelope_cutover_index` disagrees with snapshot header), `ErrEncryptionNotBootstrapped` (`enable-raft-envelope` before bootstrap), `ErrLocalEpochOutOfRange` on the wire side of `GetCapability` / `GetSidecarState`. Bundles with Stage 6E (`enable-raft-envelope` admin RPC + Phase-2 cutover). | open | — |
 | 6D | §6.6 `enable-storage-envelope` admin RPC + §7.1 Phase-1 storage cutover (§6.2 toggle ON) + Voters ∪ Learners capability gate (depends on 6B for mutator wiring AND 6C-1+6C-2 for §9.1 startup-refusal guards; bundles 6C-3 for the membership-view-dependent collision check — see rationale) | open | — |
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

Stage 6C-2b opening review. Scope: encryption-side primitive for the `ErrSidecarBehindRaftLog` §9.1 guard. The raftengine integration + main.go wiring is split into a follow-up 6C-2c PR (per design doc updates in this PR).

What this ships:
- `ErrSidecarBehindRaftLog` sentinel — catches partial-write crashes between encryption-relevant Raft commit and the §5.1 sidecar update
- `IsEncryptionRelevantOpcode` — §5.5 predicate (opcodes in [OpEncryptionMin, OpEncryptionMax])
- `EncryptionRelevantScanner` interface — cross-package contract that the raftengine implements in 6C-2c
- `GuardSidecarBehindRaftLog` — pure-function guard with 5 testable branches

- nil scanner with non-empty gap MUST fail closed with ErrSidecarBehindRaftLog (cannot prove safety)
- nil scanner on caught-up path MUST return nil (scanner irrelevant when no gap)
- Scanner errors MUST propagate distinctly from ErrSidecarBehindRaftLog (different operator triage)
- Scanner is NEVER called when sidecar is caught up (no unnecessary I/O)
- Predicate range matches §5.5 exactly (OpEncryptionMin through OpEncryptionMax)
- Both indices appear in the error annotation when the guard fires (runbook breadcrumb)

bootjp:
@claude review

Round-1 (commit `7e2133f5`): all r1 findings addressed.

## codex P2 [Correct opcode-byte contract in predicate docs] addressed

The previous `IsEncryptionRelevantOpcode` docstring was inverted: it said the opcode byte is "the first non-version byte of the FSM payload (after the 0x01 wire-version prefix)". The actual wire format is the OPPOSITE — opcode is `data[0]` (the leading tag), version 0x01 lives INSIDE the opcode-specific payload that follows.

Confirmed against production dispatcher: `kv/fsm_encryption.go:68` switches on `wireBytes[0]`, and the `EncodeBootstrap` / `EncodeRotation` / `EncodeRegistration` helpers produce `[opcode, version=0x01, ...payload]`. The predicate code was always correct; only the docstring was wrong. A 6C-2c scanner that followed the wrong docstring would have inspected payload bytes (e.g., a rotation sub-tag) and returned false negatives, silently letting `GuardSidecarBehindRaftLog` miss encryption-relevant gaps.

Updated the docstring with an explicit IMPORTANT note pointing at the production dispatcher and the codex P2 reference.

## coderabbit minor + gemini medium [byte-loop overflow] addressed

`TestIsEncryptionRelevantOpcode_AllRangeMembers` iterated with byte arithmetic which would infinite-loop if `OpEncryptionMax` ever reaches 0xFF. Changed to int-typed loop with byte cast inside. The design reserves 0x06/0x07 and Stage 6E will extend the range, so the latent bug needed fixing now.

## claude r1 observation 2 [pin reserved 0x06/0x07 explicitly] addressed

Added `{"Reserved_0x06", 0x06}` and `{"Reserved_0x07", 0x07}` to `TestIsEncryptionRelevantOpcode_KnownRanges`. The reserved slots are now self-documenting in the test table.

## claude r1 observation 3 [replace containsAll with strings.Contains] addressed

Removed the custom `containsAll` helper and replaced the multi-substring assertion with an inline loop that uses `strings.Contains`.

## Caller audit

No public API changes. The docstring correction reflects the ACTUAL wire format (which the predicate code was always reading correctly). All three test-side changes are net-additive or mechanical (loop semantics + helper removal).

bootjp:
@claude review

Round-2 (commit `8945b776`): claude r2 nit addressed.

PR-number attribution suffixes (`(codex P2 on PR #782)`, `(claude r1 observation 2)`, `coderabbit minor + gemini medium on PR #782 caught this`) removed from three call sites. CLAUDE.md convention preserved — the load-bearing why (wire-format contract / latent overflow / reserved-not-just-3-opcodes) stays in the comments; the review-round attribution is dropped to avoid rot.

Also saved this pattern to ~/.claude/.../memory/ so future PRs in this repo don't reintroduce the anti-pattern.

bootjp:
@claude review

Round-3 (commit `50dbeeb0`): codex r3 P2 [Exclude registration opcode from sidecar gap predicate] addressed.

## Semantic fix — IsEncryptionRelevantOpcode narrowed

The previous predicate range `[OpEncryptionMin, OpEncryptionMax]` incorrectly included OpRegistration (0x03). But `ApplyRegistration` in applier.go only mutates writer-registry rows via `SetRegistryRow`; it NEVER calls `WriteSidecar`. So a gap containing OpRegistration entries doesn't actually mean the sidecar is stale — `sidecar.RaftAppliedIndex` is not advanced by registration applies (only WriteSidecar moves it), so every restart after the first RegisterEncryptionWriter would have fired `ErrSidecarBehindRaftLog` and refused to boot. Cluster boot loop bug.

Design §5.5 list confirms: `rotate-dek`, `rewrap-deks`, `bootstrap-encryption`, `enable-storage-envelope`, `enable-raft-envelope`, `retire-dek` — notably NOT register-writer.

Fix: narrow range to `[OpBootstrap, OpEncryptionMax]` = [0x04, 0x07]. OpRegistration (0x03) is explicitly excluded with an IMPORTANT block in the godoc pointing at ApplyRegistration as the authoritative reference.

## Semantic-change caller audit (per cron directive)

`IsEncryptionRelevantOpcode`'s return value changes for opcode=0x03 (true → false). Audited all callers via grep:

- `audit.go` (definition, updated)
- `audit_test.go` (tests updated to assert the new behavior — new `OpRegistration_0x03_NOT_relevant` sub-test pins the exclusion explicitly)
- `errors.go` (`ErrSidecarBehindRaftLog` godoc updated to reflect the narrowed range and the OpRegistration exclusion reason)

No production caller exists yet (the predicate is consumed by `GuardSidecarBehindRaftLog` whose only caller is 6C-2c, not yet landed). The 6C-2c scanner author sees the corrected predicate from day one.

## Test additions

- `TestIsEncryptionRelevantOpcode_AllRangeMembers` asserts OpRegistration NOT relevant before iterating the sidecar-mutating range
- `TestIsEncryptionRelevantOpcode_KnownRanges` adds `OpRegistration_0x03_NOT_relevant` sub-test for grep-able regression anchoring

PR #783, opened 2026-05-17
feat(raftengine): Stage 6C-2c — encryptionScanner against etcdraft.MemoryStorage

bootjp:
## Summary

Stage 6C-2c per the [PR #762 plan](https://github.com/bootjp/elastickv/pull/762) and the 6C sub-decomposition landed in [PR #781](https://github.com/bootjp/elastickv/pull/781) / [PR #782](https://github.com/bootjp/elastickv/pull/782).

Ships the **raftengine-side scanner** that implements the `encryption.EncryptionRelevantScanner` interface defined in [PR #782 (Stage 6C-2b)](https://github.com/bootjp/elastickv/pull/782). The `main.go` startup-guard wiring that ties this scanner to the encryption-side guard is split into a new **6C-2d** sub-milestone per the design doc update in this PR.

### `encryptionScanner` against `etcdraft.MemoryStorage`

```go
func (e *Engine) EncryptionScanner() encryption.EncryptionRelevantScanner
```

Walks `raftpb.Entry` rows in `(startExclusive, endInclusive]` via `storage.Entries(lo, hi, scanMaxBytes)`, filters out:
- `EntryConfChange` / `EntryConfChangeV2` — config changes never carry FSM payloads; their `Data[0]` could coincidentally fall in the encryption opcode range without this gate
- Zero-length normal entries — etcd-raft leadership bumps after election; would `panic: index out of range` on a naive `Data[0]` read

Returns `IsEncryptionRelevantOpcode(ent.Data[0])` over surviving entries — i.e., asks whether any entry's opcode is in `[OpBootstrap, OpEncryptionMax]` = `[0x04, 0x07]`. Note: **OpRegistration (0x03) is correctly excluded** per the predicate's narrowed range from PR #782.

### Error semantics

| Path | Behavior |
|---|---|
| Compacted range (sidecar below snapshot horizon) | Wrapped `ErrCompacted` propagated; caller routes as scanner-error refusal **distinct** from `ErrSidecarBehindRaftLog` |
| `nil` storage (engine not opened) | Returns error rather than nil-deref |
| Empty range | `(false, nil)` fast path; storage NOT consulted |

### Tests (8 cases covering every branch)

| Test | Pins |
|---|---|
| `EmptyRange` (3 sub-cases) | equal / start-ahead / both-zero — storage never consulted |
| `NoRelevantEntries` | OpRegistration (0x03) explicitly NOT hit — pins the PR #782 predicate narrowing end-to-end |
| `BootstrapHit` | OpBootstrap in range → hit |
| `RotationHit` | OpRotation in range → hit |
| `NonNormalEntriesSkipped` | ConfChange / ConfChangeV2 with `Data[0]` in encryption range → NOT hit |
| `EmptyDataSkipped` | nil / empty `Data` → NOT hit AND no panic |
| `CompactedRange` | Snapshot at index 10, scan `[0, 5)` → error, hit=false |
| `NilStorageError` | Zero-value scanner returns an error |

## Design doc updates

- **6C-2c** row rewritten to reflect the scanner-only scope of this PR (engine implementation + `Engine.EncryptionScanner()` accessor, **no** `main.go` wiring).
- **6C-2d** row added — `main.go` startup-guard phase that ties 6C-2b's pure-function guard to 6C-2c's engine scanner. Will ship as its own PR.
- Shipping order updated: `6C-1 ✅ → 6C-2 ✅ → 6C-2b ✅ → 6C-2c (this PR) → 6C-2d → 6D (bundles 6C-3) → 6E (bundles 6C-4)`.
- Rationale section updated with the 6C-2c / 6C-2d split.

## Why split 6C-2c into "scanner" vs "wiring"

Keeping the engine scanner separate from the `main.go` wiring lets:
- Reviewers focus on one concern per PR.
- Stage 6D-and-later code paths that need the scanner (capability fan-out gap-coverage check) depend on the engine package without waiting for 6C-2d.
- 6C-2d's main.go integration touches the multi-shard lifecycle, which is a different concern from the engine-internal scanner implementation here.

## Caller audit

No public API changes outside `Engine.EncryptionScanner()`, which is net-additive. `encryptionScanner` and `isEncryptionRelevantEntry` are package-internal. No existing caller references the new accessor.

The implementation depends only on:
- `encryption.EncryptionRelevantScanner` (interface from PR #782)
- `encryption.IsEncryptionRelevantOpcode` (predicate from PR #782, OpRegistration-excluded)
- `etcdraft.MemoryStorage.Entries` (existing engine internal)
- `raftpb.Entry` / `raftpb.EntryConfChange` / V2 (existing types)

No changes to the engine's apply path or any other lifecycle.

## Five-lens self-review

1. **Data loss** — net-positive. The scanner enables the §9.1 refusal that catches partial-write crashes between an encryption-relevant Raft commit and the sidecar update. Until 6C-2d wires the scanner into main.go the scanner is operator-inert, but its failure modes (compacted-range error, conf-change skip, empty-data skip, nil-storage error) are all correct.
2. **Concurrency / distributed failures** — scanner reads from `MemoryStorage` (thread-safe per etcd-raft contract). Scanner itself is stateless.
3. **Performance** — zero hot-path impact. Called once per shard per process start in 6C-2d, with at most one `Entries()` call in the typical case (gap size << 64 MiB).
4. **Data consistency** — the scanner respects MemoryStorage's compaction boundary: never silently returns "no hit" for a range below the snapshot horizon; propagates the error so the guard routes it as a distinct scanner-error refusal.
5. **Test coverage** — 8 tests cover all 6 predicate paths, all 2 filter gates (type, length), AND the 2 error paths (compacted, nil-storage). The OpRegistration exclusion is pinned end-to-end with a dedicated test case.

## Test plan

- [x] `go test -race -timeout=120s ./internal/raftengine/etcd/... ./internal/encryption/...` — PASS
- [x] `golangci-lint run ./internal/raftengine/etcd/...` — 0 issues
- [ ] Full Jepsen suite — not run; scanner has no caller in this PR (operator-inert until 6C-2d)

## Plan

- **6C-2d** — `main.go` startup-guard phase that calls `encryption.GuardSidecarBehindRaftLog` with the scanner from this PR
- **6C-3** — `ErrNodeIDCollision` + `ErrLocalEpochRollback` (bundles with Stage 6D)
- **6D** — `enable-storage-envelope` admin RPC + §7.1 Phase-1 cutover (unblocked once 6C-2d lands)

## Summary by CodeRabbit

* **New Features**
  * Advances encryption-at-rest implementation with internal validation mechanisms to ensure system consistency during startup.

* **Documentation**
  * Updated design documentation clarifying the encryption-at-rest implementation phases and dependency chain.

* **Tests**
  * Added comprehensive test coverage for encryption validation logic, including edge cases and error scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/783?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/design/2026_04_29_partial_data_at_rest_encryption.md
+++ b/docs/design/2026_04_29_partial_data_at_rest_encryption.md
@@ -23,7 +23,8 @@ Date: 2026-04-29
 | 6C-1 | Subset of §9.1 startup refusal guards that depend only on flags + sidecar state: sidecar present without `--encryption-enabled` (`ErrSidecarPresentWithoutFlag`, downgrade prevention), `--encryption-enabled` without `--kekFile` (`ErrKEKRequiredWithFlag`, fail-fast), KEK loaded but sidecar wrapped DEKs do not unwrap (`ErrKEKMismatch`, operator-error catch). `internal/encryption/startup.go`'s `CheckStartupGuards` runs once in `main.go run()` before `buildShardGroups`. | shipped | PR #778 |
 | 6C-2 | Process-local §9.1 guards that read only the encryption package's own state: `ErrLocalEpochExhausted` (any active DEK with `local_epoch == 0xFFFF` — refuses to start until rotated, preventing GCM nonce reuse on the next write), `ErrUnsupportedFilesystem` startup probe (`ProbeSidecarFilesystem` exercises the §5.1 write+rename+dir.Sync sequence on a sentinel file in the sidecar directory at startup so filesystem-capability failures surface BEFORE the first encryption-relevant Raft entry, not on the first WriteSidecar call hours into operation). Also includes the gemini-r3 medium parseErr-annotation fix deferred from PR #778 (`ErrKEKMismatch` now wraps both `parseErr` and the unwrap error when the defensive non-decimal-key_id branch fires). The asymmetric `ErrLocalEpochRollback` peer is split into 6C-3 because it requires the writer-registry record (Stage 7). The `ErrSidecarBehindRaftLog` guard moves to 6C-2b because it requires raftengine integration — outside the encryption package boundary. | open | — |
 | 6C-2b | `ErrSidecarBehindRaftLog` guard PRIMITIVE — encryption-side sentinel, `IsEncryptionRelevantOpcode` predicate (§5.5), `EncryptionRelevantScanner` cross-package interface, and `GuardSidecarBehindRaftLog` pure-function guard that takes both indices + a scanner callback. Tests use a fake scanner to exercise every branch (caught-up / gap-not-covered / gap-covered / scanner-error / nil-scanner-fails-closed). Wires nothing into main.go; the raftengine-side implementation lands in 6C-2c. | shipped-in-this-PR | this PR |
-| 6C-2c | Raftengine implementation of `EncryptionRelevantScanner` (WAL-based scan of Raft entries in the supplied index range against the §5.5 predicate) + `main.go` startup-guard phase that runs after `buildShardGroups` for each shard with encryption enabled, before any gRPC server starts serving. Inherits the encryption-side primitive from 6C-2b. | open | — |
+| 6C-2c | Raftengine implementation of `EncryptionRelevantScanner` against `etcdraft.MemoryStorage.Entries`. `Engine.EncryptionScanner()` exposes the scanner as the encryption-side interface so callers (Stage 6C-2d, 6D, 6E) inherit it without coupling to the engine internals. Tests pin every branch: empty range, no-relevant entries (with OpRegistration explicitly), bootstrap hit, rotation hit, conf-change skipped, empty-data skipped, compacted-range error propagation, nil-storage error. | shipped-in-this-PR | this PR |
+| 6C-2d | `main.go` startup-guard phase that runs after `buildShardGroups` for each shard with encryption enabled, before any gRPC server starts serving. Calls `encryption.GuardSidecarBehindRaftLog` with the shard's engine `AppliedIndex()` + sidecar `RaftAppliedIndex` + the engine's `EncryptionScanner()` from 6C-2c. The guard's existing branches (caught-up / gap-not-covered / gap-covered / scanner-error / nil-scanner-fails-closed) determine refusal behavior. | open | — |
 | 6C-3 | Cluster-wide §9.1 guards that require the Voters ∪ Learners membership view: `ErrNodeIDCollision` (two members hash to the same 16-bit `node_id`), `ErrLocalEpochRollback` (sidecar `local_epoch` ≤ writer-registry record). Naturally bundles with the Stage 6D capability fan-out which already needs that membership view. | open | — |
 | 6C-4 | Phase-2-specific §9.1 guards: `ErrEnvelopeCutoverDivergence` (sidecar `raft_envelope_cutover_index` disagrees with snapshot header), `ErrEncryptionNotBootstrapped` (`enable-raft-envelope` before bootstrap), `ErrLocalEpochOutOfRange` on the wire side of `GetCapability` / `GetSidecarState`. Bundles with Stage 6E (`enable-raft-envelope` admin RPC + Phase-2 cutover). | open | — |
 | 6D | §6.6 `enable-storage-envelope` admin RPC + §7.1 Phase-1 storage cutover (§6.2 toggle ON) + Voters ∪ Learners capability gate (depends on 6B for mutator wiring AND 6C-1+6C-2 for §9.1 startup-refusal guards; bundles 6C-3 for the membership-view-dependent collision check — see rationale) | open | — |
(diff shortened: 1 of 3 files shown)


bootjp:
@claude review

Stage 6C-2c opening review. Scope: raftengine-side implementation of `encryption.EncryptionRelevantScanner` against `etcdraft.MemoryStorage` + new `Engine.EncryptionScanner()` accessor. Inherits the encryption-side primitive from PR #782. The main.go startup-guard wiring is split into a new 6C-2d sub-milestone per the design doc update in this PR.

- OpRegistration (0x03) entries in the gap MUST NOT trigger the predicate (PR #782 narrowed range — pinned end-to-end here)
- ConfChange / ConfChangeV2 with Data[0] in the encryption range MUST NOT trigger (type filter)
- Zero-length normal entries MUST NOT panic AND MUST NOT trigger (length gate)
- Compacted-range error MUST propagate (not silently "no hit")
- Hit on OpBootstrap / OpRotation
- Empty-range fast path: storage never consulted
- nil-storage zero-value scanner fails closed

bootjp:
@claude review

Round-1 (commit `cee20b3d`): all 5 r1 findings addressed.

## gemini CRITICAL — envelope-strip bug

The scanner was reading `ent.Data[0]` directly, but the engine wraps every EntryNormal in a 9-byte proposal envelope (`[version=0x01][8-byte proposal id][payload]`). In production every Raft entry would have been classified as non-encryption-relevant, **silently defeating the §9.1 ErrSidecarBehindRaftLog guard the scanner exists to enforce**.

Fix: `isEncryptionRelevantEntry` now decodes via `decodeProposalEnvelope` before applying the predicate to `payload[0]`. Malformed entries (envelope version != 0x01 or length < 9) fail the ok-check and short-circuit as not-relevant.

Why the bug slipped past tests: `newEntry` constructed bare-byte `Data: []byte{opcode}` (no envelope), so test `Data[0]` happened to match the opcode by coincidence. The fix updates `newEntry` to wrap via `encodeProposalEnvelope` so tests exercise the production-shape format. New `TestEncryptionScanner_EnvelopeStripping_RawBytesIgnored` explicitly verifies malformed entries are NOT classified as relevant.

## gemini medium [newEntry doesn't match production format] addressed (same fix as CRITICAL above)

## codex P2 [Nil Engine receiver would panic] addressed

`Engine.EncryptionScanner()` is now nil-receiver-safe; returns a zero-value scanner whose `HasEncryptionRelevantEntryInRange` surfaces the nil-storage error. Matches package convention (Status() / AppliedIndex() on nil receivers). New `TestEngineEncryptionScanner_NilReceiver` pins this.

## coderabbit major [Empty batch silently classified as no hit] addressed

The empty-batch + no-error path now returns a wrapped error rather than silently breaking out of the loop. Caller routes scanner errors distinctly from `ErrSidecarBehindRaftLog`.

## coderabbit major [Compacted-range test was permissive] addressed

Changed `t.Logf` → `t.Fatalf` on the `errors.Is(err, etcdraft.ErrCompacted)` assertion. The sentinel is now strictly enforced.

## coderabbit minor [uint64 wraparound] addressed

Explicit guard at the top of `HasEncryptionRelevantEntryInRange` refuses if `endInclusive == math.MaxUint64` (otherwise `endInclusive+1` wraps to 0 and the loop never runs). New `TestEncryptionScanner_EndInclusiveMaxUint64Refused` pins this.

## Semantic-change caller audit (per cron directive)

- `isEncryptionRelevantEntry`: behavior changes from "always false" (bug) → "correct per §5.5". No production caller yet (Stage 6C-2d wiring is the follow-up).
- `HasEncryptionRelevantEntryInRange`: empty-batch and MaxUint64 paths now return errors instead of silent false negatives — fail-closed; consistent with the existing nil-storage error path.
- `Engine.EncryptionScanner()`: nil-receiver branch is new; consistent with package convention.

Three new tests pin all three new fail-closed branches end-to-end.

bootjp:
@claude review

Round-2 (commit `644e428d`): both r2 nits addressed.

## Issue A [stale struct comment] addressed

The `encryptionScanner` struct comment claimed "constructed once per Engine and embedded into the engine struct" but `EncryptionScanner()` allocates a fresh value per call (intentionally — the scanner is stateless). Updated to "wraps the engine's \*MemoryStorage; a fresh value is constructed per Engine.EncryptionScanner() call (cheap — no internal state)" to match implementation.

## Issue B [awkward errors.Wrapf(errors.New(...))] addressed

Replaced `errors.Wrapf(errors.New("no progress"), ...)` with `errors.WithStack(errors.Newf(...))` which is both cleaner reading and matches the existing pattern at `internal/admin/config.go:109` for newly-constructed errors that satisfy wrapcheck.

bootjp:
@claude review

Round-3 (commit `894d2f9f`): claude r3 "check ordering" finding addressed.

## Guard reorder

The `EncryptionRelevantScanner` interface (`audit.go`) requires empty-range to return `(false, nil)` UNCONDITIONALLY. The previous guard order in `HasEncryptionRelevantEntryInRange` checked `s.storage == nil` first, so a nil-storage scanner with an empty range would return the nil-storage error instead of the contract-required `(false, nil)`.

Fix: reordered so empty-range short-circuits before nil-storage. `GuardSidecarBehindRaftLog` pre-filters the empty-gap case in production today, so this is contract conformance rather than a reachable bug — but pinning the order now avoids the surprise for any future caller (6C-2d main.go phase, 6D capability fan-out) that defensively calls scan with a same-index range on a maybe-nil engine.

`TestEncryptionScanner_NilStorageEmptyRange` pins the order. The existing `TestEncryptionScanner_NilStorageError` still passes because it uses a non-empty range (0, 5), so the new short-circuit doesn't interfere with the nil-storage error path it tests.

PR #784, opened 2026-05-17
feat(main): Stage 6C-2d — wire ErrSidecarBehindRaftLog startup guard into main.go

bootjp:
## Summary

Stage 6C-2d per the [PR #762 plan](https://github.com/bootjp/elastickv/pull/762) and the 6C sub-decomposition landed in PR #781 / #782 / #783.

Ties the encryption-side guard primitive ([PR #782 `GuardSidecarBehindRaftLog`](https://github.com/bootjp/elastickv/pull/782)) to the raftengine-side scanner ([PR #783 `Engine.EncryptionScanner()`](https://github.com/bootjp/elastickv/pull/783)) in `main.go`'s startup phase. **Completes the §9.1 `ErrSidecarBehindRaftLog` work end-to-end.**

### `main_encryption_startup_guard.go` (new)

| Function | Role |
|---|---|
| `checkSidecarBehindRaftLog(runtimes, defaultGroup, sidecarPath, encryptionEnabled)` | Locates the default-group runtime, type-asserts its engine to local `encryptionGapEngine` interface, reads the §5.1 sidecar, delegates to the inner helper |
| `runSidecarBehindRaftLogGuard(gapEngine, sidecarPath, defaultGroup)` | Per-engine inner half. Split out so tests use a stub instead of constructing a full `raftengine.Engine` |
| `chainEncryptionStartupGuard(prevErr, ...)` | Composes the guard with the existing `buildShardGroups` error path. Keeps `run()`'s cyclop budget intact |

### Lifecycle phase ordering

```
1. loadKEKAndRunStartupGuards   ← 6B-2 RPC gate + 6C-1/6C-2 flag/sidecar/exhaustion
2. buildShardGroups             ← opens every shard's engine
3. chainEncryptionStartupGuard  ← 6C-2d gap-coverage guard (NEW)
4. gRPC servers / accept loop
```

Step 3 runs in a **different lifecycle phase** from steps 1 — the guard requires the engine to be open and its applied index populated from WAL+snapshot replay.

### Skip conditions (`checkSidecarBehindRaftLog` returns nil)

- `encryptionEnabled` is false — no encryption opt-in, no gap to refuse on
- `sidecarPath` is empty — operator hasn't configured a sidecar location
- On-disk sidecar file absent — bootstrap hasn't committed yet
- Default-group runtime missing — no engine to query
- Engine not open — earlier startup-guard layer would have caught this

### Why scope is restricted to the default group

The §5.1 sidecar tracks a **single** `raft_applied_index`, advanced only by `WriteSidecar` calls in `ApplyBootstrap` / `ApplyRotation` on the default group's FSM. Other shards' Raft logs don't carry encryption FSM entries — running the guard against them would always report caught-up (gap == 0). `findDefaultGroupRuntime` silently skips non-matching runtimes.

## Test coverage (11 new tests)

| Test | Path |
|---|---|
| `_DisabledNoop` | `encryptionEnabled=false` → skip |
| `_NoSidecarPathNoop` | empty path → skip |
| `_SidecarAbsentNoop` | on-disk file missing → skip |
| `_SidecarStatError` | I/O error (NUL path) → wrapped, NOT `not-exist` |
| `_NoRuntimes` | no default-group runtime → skip |
| `_CaughtUp` | sidecar idx >= engine idx → pass, scanner never consulted |
| `_GapNotCovered` | gap exists, scanner says no → pass |
| `_GapCovered` | gap covers relevant entry → `ErrSidecarBehindRaftLog` |
| `_ScannerError` | scanner error → propagated wrapped, NOT `ErrSidecarBehindRaftLog` |
| `_PropagatesPrevError` | chain with non-nil prev → returns prev verbatim |
| `_NilPrevRunsGuard` | chain with nil prev → forwards to guard |

Tests use `stubGapEngine` + `stubScanner` to exercise the per-engine inner function directly, avoiding the full `raftengine.Engine` interface (which would require panic-stubs for ~30 unrelated methods).

## Posture matrix (before/after this PR)

| Scenario | pre-this-PR | post-this-PR |
|---|---|---|
| Sidecar caught up to engine | boot, encrypted | boot, encrypted |
| Sidecar behind, gap harmless (no §5.5 opcodes) | boot, encrypted | boot, encrypted |
| Sidecar behind, gap covers `OpBootstrap` / `OpRotation` | **boot, HaltApply on first post-cutover read with `unknown_key_id`** | refuse: `ErrSidecarBehindRaftLog` |
| Sidecar behind, gap covers only `OpRegistration` | boot, registry replays cleanly | boot (PR #782 narrowed predicate excludes 0x03) |

## Caller audit (per cron directive)

No public function signatures changed. The new functions in `main_encryption_startup_guard.go` are package-internal. `main.go run()` is the sole production caller; tests call the inner helpers directly.

The semantic effect of the wiring: nodes whose sidecar is behind by an encryption-relevant interval now refuse at startup instead of silently booting into a HaltApply later. This is a tighter failure boundary — operators see the §9.1 typed refusal pointing at `encryption resync-sidecar` rather than triaging downstream `unknown_key_id` errors.

## Stage 6B-2 / 6C-1 / 6C-2 / 6C-2b / 6C-2c invariants preserved

All earlier guards and their lifecycle phases unchanged. This PR composes existing primitives — no new sentinel, no new opcode, no new wire format.

## Five-lens self-review

1. **Data loss** — net-positive. Catches the partial-write-crash failure mode between an encryption-relevant Raft commit and the sidecar update. The guard primitive existed since PR #782 but was operator-inert until this wiring.
2. **Concurrency / distributed failures** — runs at startup before any goroutine fan-out / gRPC accept. No shared state introduced.
3. **Performance** — single sidecar read + single scanner call per process start on the happy path.
4. **Data consistency** — the only §9.1 mechanism that catches "Raft committed but sidecar missed it" partial writes. Without it the first post-cutover read would HaltApply.
5. **Test coverage** — 11 new tests cover every branch. Full main package test suite passes with no regressions.

## Test plan

- [x] `go test -race -timeout=180s .` — PASS (all main package tests + 11 new)
- [x] `go test -race -timeout=120s ./internal/raftengine/etcd/... ./internal/encryption/...` — PASS (no regressions)
- [x] `go build ./...` — PASS
- [x] `golangci-lint run .` — 0 issues (cyclop holds at 10 via the composition helper)
- [ ] Full Jepsen suite — not run; the guard is purely a startup refusal that runs before any Raft entry can be served

## Plan

**Stage 6C is now complete** (6C-1 ✅ + 6C-2 ✅ + 6C-2b ✅ + 6C-2c ✅ + 6C-2d this PR).

Next stages per PR #762:
- **6C-3** — `ErrNodeIDCollision` + `ErrLocalEpochRollback` (bundles with Stage 6D's capability fan-out)
- **6D** — `enable-storage-envelope` admin RPC + §7.1 Phase-1 cutover (**now UNBLOCKED** — all `ErrSidecarBehindRaftLog` work has shipped)
- **6E** — `enable-raft-envelope` admin RPC + §7.1 Phase-2 cutover (bundles 6C-4)
- **6F** — `--encryption-rotate-on-startup` ergonomics

## Summary by CodeRabbit

* **New Features**
  * Added encryption startup validation to ensure sidecar state is synchronized with the raft log during system initialization, preventing startup if consistency checks fail.

* **Tests**
  * Comprehensive test coverage for encryption startup guard behavior, including fast-skip scenarios, gap detection, and error propagation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/784?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bootjp:
@claude review

Stage 6C-2d wires the §9.1 `ErrSidecarBehindRaftLog` guard into `main.go`'s startup phase — the last load-bearing piece of the 6C ErrSidecarBehindRaftLog work (after PR #781 / #782 / #783).

Please validate:

1. **Default-group-only scope** — `findDefaultGroupRuntime` only matches the runtime whose `spec.id == defaultGroup`. Per the §5.1 sidecar contract, only that group's FSM advances `RaftAppliedIndex` (via `WriteSidecar` in `ApplyBootstrap` / `ApplyRotation`). Confirm running the guard against non-matching shards is correctly suppressed (silent skip, not error).

2. **Fast-skip ordering in `checkSidecarBehindRaftLog`** — five no-op conditions: `!encryptionEnabled` → `sidecarPath == ""` → sidecar absent → runtime missing → engine missing. `os.Stat` returning `os.ErrNotExist` returns nil; any other stat error wraps. Confirm no early-return swallows a legitimate I/O error.

3. **Lifecycle phase ordering** — the guard runs AFTER `buildShardGroups` (engine open + applied index populated from WAL+snapshot replay) and BEFORE `metricsRegistry.SetFSMApplySyncMode` / gRPC serving. This is a different phase from the 6C-1/6C-2 `CheckStartupGuards` (which run BEFORE engine open).

4. **`chainEncryptionStartupGuard` composition** — keeps `run()`'s cyclop budget at exactly 10. The helper short-circuits on non-nil `prevErr` (returns it verbatim) or forwards to `checkSidecarBehindRaftLog` on nil `prevErr`. Confirm `buildShardGroups` errors are NOT shadowed.

5. **Type assertion** — `engine.(encryptionGapEngine)` requires `AppliedIndex() uint64` + `EncryptionScanner() encryption.EncryptionRelevantScanner`. The production `etcd.Engine` satisfies this structurally (PR #783). Confirm the fmt.Errorf on type-assert failure surfaces enough context for an operator to triage (currently mentions the default group id and the missing methods).

6. **Caller audit** — only one production caller (`run()` in `main.go`). The three new functions are package-internal. No public signatures changed. Tests use stubs to exercise the per-engine inner function directly.

7. **Test coverage** — 11 new tests cover every branch: 5 fast-skip paths, 3 guard outcomes (caught-up, gap-not-covered, gap-covered, scanner-error), 2 chain-composition paths. Confirm the `_GapCovered` test would catch a regression where the guard accidentally always passed.

8. **Five-lens self-review** in the PR body — particularly the data-loss lens (catches partial-write between Raft commit + sidecar update) and the data-consistency lens (only §9.1 mechanism for the post-cutover `unknown_key_id` HaltApply scenario).

Verification done locally:
- `go test -race -timeout=180s .` — PASS (all main package tests + 11 new)
- `go test -race -timeout=120s ./internal/raftengine/etcd/... ./internal/encryption/...` — PASS (no regressions)
- `golangci-lint run .` — 0 issues

bootjp:
@claude review

Round-1 (commit `187bf310`): addressed gemini high (line 101) + codex P1 (line 125).

## gemini high — nil-engine fail-closed

`checkSidecarBehindRaftLog` returned nil when `engine` was nil, silently skipping the §9.1 guard. Changed to `pkgerrors.Errorf` so the lifecycle bug surfaces. Also upgraded the pre-existing `fmt.Errorf` on the interface-cast branch to `pkgerrors.Errorf` for stack-trace parity. Caller audit: only one production caller (`chainEncryptionStartupGuard` → main.go:359) which already treats non-nil return as startup abort.

## codex P1 — skip-when-zero gate

Confirmed via `grep -n RaftAppliedIndex internal/encryption/applier.go` that `writeBootstrapSidecar` / `writeRotationSidecar` never advance `sidecar.raft_applied_index`. Without the applier-side advancement, firing the guard against `(0, engine.applied]` would refuse startup on every restart of an encrypted cluster.

Resolution: skip the guard when `sidecar.RaftAppliedIndex == 0` with a WARN log indicating the applier-side advancement is a Stage 6C-2e follow-up. Fail-OPEN for this transitional window — matches the pre-6C-2d posture (no guard) so this PR does not regress any cluster. Once the applier advances the index in 6C-2e, the skip condition is naturally a no-op and the guard becomes active without code changes.

Pinned with `TestRunSidecarBehindRaftLogGuard_SidecarIndexZero_SkipsTransitionally` — uses RaftAppliedIndex=0 + stub scanner with hit=true and asserts the guard returns nil (pre-fix would have returned `ErrSidecarBehindRaftLog`).

Caller audit for the semantic change: `runSidecarBehindRaftLogGuard` has one caller, `checkSidecarBehindRaftLog`. The skip path joins the existing guard-passed nil-return semantics.

## Coderabbit Minor (docs:27)

"For each shard" wording — below the loop's P0/P1/high/critical/major threshold. Happy to address in a follow-up if reviewers prefer it in this PR.

bootjp:
@claude review

PR784 round-2 — addressed:

1. **codex P1** — load-bearing fix. The previous round (skip-when-zero gate at 187bf310) deferred the applier-side fix to a follow-up; this round ships the actual fix in the same PR:
   - Extended `EncryptionApplier.ApplyBootstrap` / `ApplyRotation` to receive a `raftIdx uint64` parameter.
   - New `raftengine.ApplyIndexAware` opt-in interface lets the etcd engine deliver `entry.Index` to the FSM via `SetApplyIndex` immediately before each `Apply`. Existing `StateMachine` implementations don't need to change.
   - `kvFSM` opts in and threads the index through `applyEncryption` → `applyBootstrap` / `applyRotation` → the applier.
   - `writeBootstrapSidecar` / `writeRotationSidecar` now call `advanceRaftAppliedIndex(sc, raftIdx)` inside the same `WriteSidecar` fsync that mutates `keys[]`. The helper enforces skip-on-zero (so callers that don't opt in never regress an existing index) and monotonic-only advancement (replay-safe).
   - Two regression tests pin the new behavior: `TestApplyBootstrap_PersistsRaftAppliedIndex` and `TestApplyRotation_PersistsRaftAppliedIndex`. A third (`TestApplyBootstrap_ZeroRaftIdxPreservesExisting`) pins the skip-on-zero policy.
   - The legacy-sidecar compat gate from 187bf310 stays in place — its comment is rewritten to describe its new role: handling sidecars that bootstrapped on a pre-fix binary. Fresh sidecars written by this PR's applier carry a non-zero index from the first Bootstrap, so the gate is naturally a no-op for them.

2. **gemini high** — converged on the same fix-closed posture for nil engine that round-1 (187bf310) already shipped. Both error sites use `pkgerrors.Errorf` for stack-trace parity; combined nil-engine check now reads:
   > `encryption startup guard: engine for default group %d is nil (buildShardGroups returned without an opened engine)`

3. **coderabbit minor** — design doc 6C-2d row reworded from "for each shard with encryption enabled" to "runs once on the default group"; the row now also documents the companion `EncryptionApplier` interface extension + `sidecar.RaftAppliedIndex` advancement that completes the §9.1 guard end-to-end.

### Caller audit (per cron directive — semantic changes)

- **`EncryptionApplier.ApplyBootstrap` / `ApplyRotation` signatures** now take `raftIdx uint64` as first param. Production callers updated: `kv/fsm_encryption.go` and the only impl `internal/encryption.Applier`. Test fake (`fakeApplier` in `kv/fsm_encryption_test.go`) updated to record `lastBootstrapIdx` / `lastRotationIdx`. No other implementations exist.

- **`raftengine.ApplyIndexAware`** is a new optional interface; existing `StateMachine` implementations that don't opt in continue to compile and work unchanged. Only `kvFSM` opts in.

- **`runSidecarBehindRaftLogGuard` nil-engine semantics**: returns error instead of nil. Sole production caller is `checkSidecarBehindRaftLog` → `chainEncryptionStartupGuard` → `main.go run()`, which already aborts on non-nil error from the chain helper.

### Five-lens self-review

1. **Data loss** — net-positive. The PR is now actually load-bearing for catching the partial-write window between an encryption-relevant Raft commit and the sidecar update; previously (post-187bf310) the guard was wired but dormant on fresh clusters too.
2. **Concurrency** — `SetApplyIndex` runs on the engine's serialised apply goroutine; `pendingApplyIdx` is plain field assignment. No race.
3. **Performance** — one uint64 store per Apply for the hook + one uint64 write inside an existing `WriteSidecar` fsync. Negligible.
4. **Consistency** — sidecar.RaftAppliedIndex now records the apply index inside the SAME crash-durable fsync as the keys[] mutation. The §9.1 invariant "sidecar reflects all encryption-relevant entries up to index N" now holds for newly-bootstrapped clusters.
5. **Test coverage** — 4 new applier-side tests + 1 main-side test + existing kvFSM dispatch tests extended via fakeApplier capturing threaded `raftIdx`. `go test -race` on `./kv/... ./internal/encryption/... ./internal/raftengine/... .` — all PASS. `golangci-lint run` on affected packages — 0 issues.

### What's still operator-inert

The compat gate remains for legacy sidecars bootstrapped on a pre-this-PR binary. To activate the guard on such clusters: run `encryption resync-sidecar` or rotate. Documented at the gate's WARN log line.

bootjp:
@claude review

Round-3 (commit `2a313276`): addressed the round-2 LGTM-with-observation by closing the FSM-level threading test gap.

Two new tests in `kv/fsm_encryption_test.go`:

- `TestApply_Bootstrap_ThreadsRaftIdx` — `SetApplyIndex(184273)` → `Apply(0x04 wireBytes)` → assert `applier.lastBootstrapIdx == 184273`
- `TestApply_Rotation_ThreadsRaftIdx` — same for the 0x05 rotation dispatch

These exercise the full `pendingApplyIdx → applyEncryption → applyBootstrap/applyRotation → applier.ApplyBootstrap/ApplyRotation` chain via the FSM dispatch path. The applier-level regression tests (`TestApplyBootstrap_PersistsRaftAppliedIndex`) call ApplyBootstrap directly and bypass the FSM — without these new tests, a future refactor that drops `f.pendingApplyIdx` from the `applyEncryption` argument list would compile and silently truncate the index to zero.

No production code changed in this commit. Uses the `lastBootstrapIdx` / `lastRotationIdx` fields already added to `fakeApplier` in round-2 (commit `487e21d6`).

Verification:
- `go test -race ./kv/` — PASS
- `golangci-lint run ./kv/` — 0 issues

PR #785, opened 2026-05-18
perf(redis): bounded Lua VM pool with configurable max-idle cap

bootjp:
## Summary
Replace the `sync.Pool`-backed `luaStatePool` with a buffered-channel pool sized at an operator-tunable cap. Adds `--redisLuaMaxIdleStates` flag (default 64) and `Drops()` / `Idle()` / `MaxIdle()` accessors for observability.

## Motivation
Recent triage of the 5-node production cluster surfaced the Lua pool as one source of erratic memory available across nodes — Pebble + Lua VM retention oscillates with GC, and the previous `sync.Pool` had no way to bound the steady-state floor. Symptoms:
- No operator-tunable upper bound on retained idle states (GC-driven only).
- No saturation signal: `Hits()`/`Misses()` showed reuse rate but not "pool full → put rejected."

## Behaviour change
- Steady-state memory floor is now bounded at `maxIdle * per-state-footprint` (~200 KiB per state empirically).
- Get/put surface is unchanged for the hot path (`adapter/redis_lua.go runLuaScript`); throughput is not capped — get() on empty pool still falls through to a fresh allocation.
- `put()` on full pool drops + closes the state (eagerly releases per-state resources) and increments `Drops`. `Drops > 0` is the "raise maxIdle" diagnostic.

## Tests
- All existing `TestLua_VMReuse*` / `TestLua_PoolSerial*` / `TestLua_PoolRecordsReuseVsAllocation` continue to pass.
- New `TestLua_PoolBoundedOverflow` locks in the cap invariant: 2*maxIdle puts retain exactly maxIdle, surplus increments Drops, draining produces exactly maxIdle hits.
- New `TestLua_PoolBoundedClampsZeroAndNegative` locks in the non-positive → default clamp so a misconfigured CLI doesn't silently disable pooling.

## Self-review (5 lenses)
1. **Data loss** — N/A. Pool drops only reset/closed `*lua.LState`s.
2. **Concurrency** — channel ops are O(1) atomic CAS; no new locks. `-race` clean.
3. **Performance** — channel ops ~ns, comparable to sync.Pool's per-P slabs. Hot-path allocation pattern unchanged.
4. **Data consistency** — N/A.
5. **Test coverage** — two new table-driven tests for new invariants; all existing pool tests retained.

## Test plan
- [x] `go test -race -count=1 -short ./adapter` — 564s, all green
- [x] `go vet ./adapter/` — clean
- [ ] CI green
- [ ] Production cluster: deploy with `--redisLuaMaxIdleStates=64` (default), verify `lua_pool_drops_total` metric (follow-up) stays at 0 under steady load; raise to 128 if Drops climbs

## Summary by CodeRabbit

* **New Features**
  * New command-line flag to configure the maximum number of idle Redis Lua VM states for better memory control.
  * Lua VM reuse now uses a bounded idle pool with deterministic overflow behavior to limit retained VMs.
  * Improved runtime diagnostics exposing current idle count, configured maximum, and overflow/drop metrics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/785?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bootjp:
@claude review

Round-1 fixes:

- **gemini medium (get)**: refactored to centralize ctxBinding assignment & return after the select per the snippet
- **gemini medium (put)**: added `len(p.idle) >= p.maxIdle` fast-path drop **before** `pls.reset()` so saturated EVAL paths skip the expensive globals/tables/metatables walk. Race-safety: select-based fallback retained for stale-snapshot races. Caller audit: only `runLuaScript` defers `luaPool.put(pls)` and doesn't touch `pls` after — no behavioural break.
- **golangci mnd (main.go:100)**: exported `adapter.DefaultLuaPoolMaxIdle` and use it as flag default. Single source of truth.
- **golangci gci**: `golangci-lint fmt` cleanup.

`golangci-lint run` 0 issues, race tests green.

bootjp:
@claude review

Round-2 fixes for your r1 findings:

- **Medium (godoc mis-attribution)**: deleted the duplicated luaStatePool narrative that was running into `DefaultLuaPoolMaxIdle` with no separator; security-invariant content now lives on `type luaStatePool struct` (where it belongs), const has its own clean doc.
- **Nit (unreachable nil guard)**: rewrote the comment to be explicit — "Intentionally-unreachable defence in depth" — and explain that put() already filters nil.
- **Low (stale sync.Pool references)**: updated all four comment clusters in `redis_lua_pool_test.go` to reflect channel-pool semantics. The actual non-determinism justifying the loop is t.Parallel sibling scheduling, not sync.Pool's GC eviction or -race per-P caching.

golangci-lint 0 issues, race tests green.

PR #786, opened 2026-05-18
docs(encryption): Stage 6D-1 proposed design — enable-storage-envelope cutover

bootjp:
## Summary

Doc-only PR. Stage 6D-1 per the [parent design's milestone table](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_04_29_partial_data_at_rest_encryption.md), now unblocked by PR #784 (6C-2d).

Per `CLAUDE.md`'s design-doc-first workflow: this doc lands BEFORE any 6D implementation so the decomposition, wire-format choice, RPC contract, capability fan-out helper, storage-layer toggle, bundled 6C-3 startup guards, and refusal posture can be reviewed as one self-contained record.

## What this ships

`docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md` — the Stage 6D as-implemented design. The parent design covers 6D abstractly in §6.6 + §7.1; this doc pins the concrete sub-PR decomposition, wire-format byte choice, and refusal posture.

## Sub-PR decomposition

| Sub-PR | Surface | Operator-visible? |
|---|---|---|
| **6D-1** | This design doc | No (doc-only) |
| **6D-2** | 6C-3 startup guards (`ErrNodeIDCollision` + `ErrLocalEpochRollback`) | No (startup-time guards) |
| **6D-3** | Capability fan-out helper in `internal/admin/` | No (library) |
| **6D-4** | Wire format: `RotateSubEnableStorageEnvelope = 0x01` + `ApplyRotation` sub-tag dispatch | No (writes only flag in sidecar) |
| **6D-5** | §6.2 storage-layer toggle: `PutAt` reads `StorageEnvelopeActive` | No (toggle stays false until 6D-6) |
| **6D-6** | `EnableStorageEnvelope` admin RPC + CLI + integration test | **Yes** — first user-visible cutover |

Each sub-PR is independently revertable. 6D-2 through 6D-5 land operator-inert (the cutover isn't reachable from any CLI until 6D-6 wires it).

## Highlights

- **Wire format**: new sub-tag `RotateSubEnableStorageEnvelope = 0x01` on the existing 0x05 OpRotation payload. No new opcode. Validation at both propose AND apply time: `Purpose == PurposeStorage`, `Wrapped == nil`, `DEKID == sidecar.Active.Storage`, `sidecar.StorageEnvelopeActive == false` (idempotency guard for retried RPCs).

- **RPC**: leader-only `EncryptionAdmin.EnableStorageEnvelope`. Followers return `FailedPrecondition` with leader hint (same as `RotateDEK`). Failure-mode table in §8 lists five new typed sentinels (`ErrEncryptionNotBootstrapped`, `ErrStorageEnvelopeAlreadyActive`, `ErrCapabilityCheckFailed`, plus existing `ErrEncryptionMutatorsDisabled` + `ErrNotLeader`).

- **Voters ∪ Learners fan-out** in `internal/admin/capability_fanout.go`. Fresh probing every call (no caching, per the parent design's "stale cached capability state cannot trigger a premature cutover" rationale). Default timeout = one `RaftElectionTimeout`.

- **6C-3 guards bundled into 6D-2**:
  - `ErrNodeIDCollision` — refuses startup if two members hash to the same 16-bit `node_id` (catches GCM nonce-reuse risk).
  - `ErrLocalEpochRollback` — refuses startup if sidecar's `local_epoch` is strictly less than the writer-registry record (catches sidecar-restored-from-old-backup).

- **§6.2 storage-layer toggle**: `store/mvcc_store.go::Put` reads `StorageEnvelopeActive` at write time. Pre-cutover versions stay cleartext; post-cutover writes are encrypted with `encryption_state = 0b01`; mixed versions within a single key are correct by construction per §5.4 of the parent doc.

## Five-lens self-review of THIS doc

1. **Data loss** — net-neutral. Cutover doesn't retire DEKs or delete values.
2. **Concurrency** — apply is deterministic; every replica flips `StorageEnvelopeActive` at the same apply index. Pre-flight fan-out is fresh per cutover.
3. **Performance** — per-Put adds one AES-GCM seal + 16-byte tag + 1-byte header (already benchmarked).
4. **Data consistency** — cutover entry's apply index is recorded in `RaftAppliedIndex` via the 6C-2d advancement; §9.1 guard then refuses startup if a later restart's sidecar is missing the cutover entry.
5. **Test coverage** — five new unit-test surfaces (one per sub-PR) + Jepsen extension at 6D-6.

## Open questions for the reviewer (§11)

1. **Sub-tag byte choice**: `0x01` (natural next value) vs `0x10` (group "cutover" sub-tags numerically away from "rotate" sub-tags). Doc picks `0x01`.
2. **Idempotency response code**: `AlreadyExists` (different CLI message) vs `OK` (same end state). Doc picks `AlreadyExists`.
3. **Fan-out helper location**: `internal/admin/` (matches existing `config.go` precedent) vs `internal/encryption/`. Doc picks `internal/admin/` because Stage 6E + the 6C-3 startup guard both reuse it.

If the reviewer disagrees with any of these, the discussion lives on this PR; subsequent implementation PRs will follow whatever this PR's review decides.

## Test plan

- [x] Doc renders cleanly on GitHub
- [x] Cross-references to parent design + 6C-2d PR #784 are valid
- [ ] No code changes; CI is doc-only

## Summary by CodeRabbit

* **Documentation**
  * Added design documentation describing an end-to-end cutover to enable cluster-wide storage envelope encryption. Covers leader-driven activation, pre-flight capability checks, idempotent retry semantics, a recorded cutover index for consistency, write-path behavior when encryption is active, failure/refusal modes, and a comprehensive testing strategy (unit/property/Jepsen).

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/786?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
+++ b/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
@@ -144,7 +144,7 @@ A rotation entry with `SubTag = 0x04` carries:
 | `SubTag` | `uint8` | `0x04` |
 | `Purpose` | `uint8` | MUST be `PurposeStorage` (`0x01`); any other value → halt apply with `ErrEncryptionApply` |
 | `DEKID` | `uint32` | MUST equal `sidecar.Active.Storage` at the leader at propose time; the applier re-checks at apply time |
-| `Wrapped` | `[]byte` | MUST be empty; the cutover does NOT add a new DEK |
+| `Wrapped` | `[]byte` | MUST be empty (`len(Wrapped) == 0`, NOT `Wrapped == nil` — see constraint #2 below for why); the cutover does NOT add a new DEK |
 | `ProposerRegistration` | `RegistrationPayload` | MUST have `DEKID = sidecar.Active.Storage`, `LocalEpoch` ≥ current registry record for that node (the proposer's epoch is expected to be at least the registry's current value; the strict-ahead invariant from §5.2 means after a normal boot it is strictly ahead, but a leader proposing the cutover on the same boot where it registered may legitimately be at equality, so the constraint accepts `=` as well as `>`). A proposer with `LocalEpoch < registry` is stale and the proposal is rejected. |
 
 Constraints (validated in both the
@@ -415,7 +415,7 @@ the rationale):
 | Sidecar missing or `Active.Storage == 0` | `FailedPrecondition` | new `ErrEncryptionNotBootstrapped` |
 | Already active (idempotent retry path) | `OK` | response's `was_already_active=true` and `applied_index = sidecar.StorageEnvelopeCutoverIndex` (the original cutover, §6.4). The OK code is intentional (see §3.2 step 5 + §11.2): unary gRPC drops the response body on non-OK status, so the idempotency contract must live on the success path. The CLI distinguishes the two outcomes by reading `was_already_active`. |
 | Capability fan-out failed | `FailedPrecondition` | new `ErrCapabilityCheckFailed` (lists offending node IDs) |
-| Apply halted (`SubTag` decode error, `Purpose != Storage`, `Wrapped != nil`, etc.) | `Internal` | existing `ErrEncryptionApply`. The duplicate-cutover-entry case is NOT in this row — §2.1 constraint #4 defines it as a benign consumed no-op (the cutover applier returns nil; nothing halts). |
+| Apply halted (`SubTag` decode error, `Purpose != Storage`, `len(Wrapped) > 0`, etc.) | `Internal` | existing `ErrEncryptionApply`. The duplicate-cutover-entry case is NOT in this row — §2.1 constraint #4 defines it as a benign consumed no-op (the cutover applier returns nil; nothing halts). |
 
 ## 4. Voters ∪ Learners capability fan-out
 
--- a/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
+++ b/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
@@ -56,8 +56,11 @@ atomic in the user-visible RPC.
   later by Stage 6E.
 - **FSM apply path**: extends `ApplyRotation` in
   `internal/encryption/applier.go` to handle the new sub-tag —
-  sets `sidecar.StorageEnvelopeActive = true` and persists via
-  the same `WriteSidecar` fsync that 6C-2d uses for
+  sets `sidecar.StorageEnvelopeActive = true` AND records the
+  applying raft index in a new `StorageEnvelopeCutoverIndex`
+  sidecar field (symmetric with the existing
+  `RaftEnvelopeCutoverIndex` field; see §6.4). Both writes land
+  inside the same `WriteSidecar` fsync that 6C-2d uses for
   `RaftAppliedIndex` advancement.
 - **§6.2 storage-layer toggle**: `store/mvcc_store.go` (and the
   underlying `store/lsm_store.go`) read `StorageEnvelopeActive`
@@ -219,6 +222,13 @@ message EnableStorageEnvelopeRequest {
 }
 
 message EnableStorageEnvelopeResponse {
+  // applied_index is the raft index of the cutover entry. On a
+  // freshly-successful call this is the entry the leader just
+  // proposed and waited to apply. On a retried call that returns
+  // AlreadyExists, this is the recorded
+  // sidecar.StorageEnvelopeCutoverIndex from the ORIGINAL
+  // cutover (see §6.4) — stable across arbitrary subsequent
+  // encryption-relevant Raft activity.
   uint64 applied_index = 1;
 
   // capability_summary records which (full_node_id) members were
@@ -226,7 +236,20 @@ message EnableStorageEnvelopeResponse {
   // Lets the operator see the membership view the leader used to
   // approve the cutover — useful for post-mortem if the leader's
   // route-catalog view diverges from the operator's expectation.
+  // Empty on AlreadyExists retries (the membership view of the
+  // original cutover is not retained).
   repeated CapabilityVerdict capability_summary = 2;
+
+  // cutover_index_unknown is set by the defensive branch
+  // described in §6.4 — only fires if a sidecar reports
+  // StorageEnvelopeActive=true with
+  // StorageEnvelopeCutoverIndex=0 (operationally impossible
+  // under normal apply; hedges against future schema rollback /
+  // hand-edited sidecar). On a healthy cluster this stays
+  // false; applied_index then carries
+  // sidecar.RaftAppliedIndex as a best-effort fallback. The
+  // CLI surfaces it as a warning.
+  bool   cutover_index_unknown = 3;
 }
 
 message CapabilityVerdict {
@@ -256,8 +279,12 @@ leader. Server-side sequence:
    "DEK bootstrap if necessary" reads as a separate explicit
    command in 6D.
 5. Verify sidecar.StorageEnvelopeActive == false — if already
-   active, return AlreadyExists with the recorded applied
-   index of the original cutover entry (idempotent retry).
+   active, return AlreadyExists with
+   sidecar.StorageEnvelopeCutoverIndex (the original cutover
+   entry's apply index, recorded by §6.4) as the response's
+   applied_index. This is stable across subsequent
+   encryption-relevant Raft entries that advance the generic
+   RaftAppliedIndex.
 6. Run the Voters ∪ Learners capability fan-out (§4). On any
    member reporting encryption_capable=false or unreachable
    within one election timeout, return
@@ -277,7 +304,7 @@ Server-side failure modes (each maps to a typed wrapped error):
 | Encryption disabled / KEK absent | `FailedPrecondition` | existing 6B `ErrEncryptionMutatorsDisabled` |
 | Caller is a follower | `FailedPrecondition` | existing `ErrNotLeader` |
 | Sidecar missing or `Active.Storage == 0` | `FailedPrecondition` | new `ErrEncryptionNotBootstrapped` |
-| Already active (idempotent retry path) | `AlreadyExists` | new `ErrStorageEnvelopeAlreadyActive` (carries the recorded apply index) |
+| Already active (idempotent retry path) | `AlreadyExists` | new `ErrStorageEnvelopeAlreadyActive` (response's `applied_index` = `sidecar.StorageEnvelopeCutoverIndex` from §6.4 — stable across subsequent encryption activity) |
 | Capability fan-out failed | `FailedPrecondition` | new `ErrCapabilityCheckFailed` (lists offending node IDs) |
 | Apply halted (`SubTag` decode, idempotency, etc.) | `Internal` | existing `ErrEncryptionApply` |
 
@@ -539,6 +566,64 @@ version. The `rewrite` migration command (deferred to Stage 9)
 backfills the old versions later; until then the mix is correct
 by construction.
 
+### 6.4 New sidecar field: `StorageEnvelopeCutoverIndex`
+
+Stage 6D-4 adds a new field to the §5.1 sidecar struct,
+symmetric with the existing `RaftEnvelopeCutoverIndex` reserved
+for Stage 6E:
+
+```go
+type Sidecar struct {
+    // ...existing fields...
+
+    StorageEnvelopeActive       bool   `json:"storage_envelope_active"`
+    StorageEnvelopeCutoverIndex uint64 `json:"storage_envelope_cutover_index"` // NEW in 6D-4
+    RaftEnvelopeCutoverIndex    uint64 `json:"raft_envelope_cutover_index"`    // reserved for 6E
+    RaftAppliedIndex            uint64 `json:"raft_applied_index"`             // 6C-2d
+}
+```
+
+**Apply-time semantics.** When `ApplyRotation` handles the
+`RotateSubEnableStorageEnvelope` sub-tag at raft index K, it
+writes:
+
+  - `sc.StorageEnvelopeActive = true`
+  - `sc.StorageEnvelopeCutoverIndex = K`
+  - `sc.RaftAppliedIndex` advanced to K via the existing
+    `advanceRaftAppliedIndex` helper shipped in 6C-2d
+
+All three writes land inside the single crash-durable
+`WriteSidecar` fsync. Atomicity is provided by the §5.1
+write-and-rename protocol — a crash mid-apply leaves either the
+pre-cutover sidecar or the post-cutover sidecar on disk, never
+a torn mix.
+
+**Idempotency contract for `AlreadyExists`.** The Stage 6D-6
+RPC handler reads `sc.StorageEnvelopeCutoverIndex` directly
+when returning `ErrStorageEnvelopeAlreadyActive` to a retried
+call. Because that field is set at apply time and never
+re-written by later encryption entries (only the cutover
+entry's apply sets it), the response carries the **original**
+cutover index regardless of how many subsequent
+encryption-relevant Raft entries have advanced
+`RaftAppliedIndex`. Operators and automation can use this as a
+stable idempotency token across arbitrary cluster activity
+following the original cutover.
+
+Older sidecars (pre-6D-4) decode the JSON with the field
+absent, in which case `StorageEnvelopeCutoverIndex` defaults to
+`0`. A `0` value on a sidecar where `StorageEnvelopeActive` is
+also false is the "cutover hasn't happened yet" baseline. A `0`
+value on a sidecar where `StorageEnvelopeActive == true` is
+operationally impossible (the apply path writes both fields
+together) but, defensively, the 6D-6 handler treats it as
+"original index unknown — return the current `RaftAppliedIndex`
+with a `cutover_index_unknown: true` flag in the response" so
+the cutover-completed posture is still reported to the
+operator. This branch should never fire in practice; it exists
+only to harden against a hypothetical sidecar-edit attack or
+a future schema rollback.
+
 ## 7. Decomposition into sub-PRs
 
 | Sub-PR | Surface | Land order | Notes |
@@ -546,7 +631,7 @@ by construction.
 | **6D-1** | This design doc | First | Doc-only; reviewable as a self-contained record |
 | **6D-2** | 6C-3 startup guards (`ErrNodeIDCollision` + `ErrLocalEpochRollback`) | Second | Cluster-wide guards bundled per parent doc §6C-3. No mutator changes. |
 | **6D-3** | Capability fan-out helper in `internal/admin/` | Third | Used by the cutover RPC (6D-6) and reused later by Stage 6E. Standalone unit-testable. |
-| **6D-4** | Wire format addition: `RotateSubEnableStorageEnvelope = 0x04` + `ApplyRotation` sub-tag dispatch (writes `StorageEnvelopeActive=true` in sidecar) | Fourth | No CLI, no §6.2 hookup yet. FSM-level testable with the existing applier_test patterns. Operator-inert at this point. |
+| **6D-4** | Wire format addition: `RotateSubEnableStorageEnvelope = 0x04` + `readRotationSubTag` whitelist (§2.2) + `ApplyRotation` sub-tag dispatch (writes `StorageEnvelopeActive=true` AND `StorageEnvelopeCutoverIndex=raftIdx` in sidecar; both writes inside one `WriteSidecar` fsync) | Fourth | No CLI, no §6.2 hookup yet. FSM-level testable with the existing applier_test patterns. Adds the new `StorageEnvelopeCutoverIndex uint64` field to the sidecar struct (§6.4). Operator-inert at this point. |
 | **6D-5** | §6.2 storage-layer toggle (PutAt reads `StorageEnvelopeActive`) | Fifth | The §4.1 envelope encoder is already in `lsm_store.go`; this PR wires the toggle in front of it. Independent unit tests. Still operator-inert until 6D-6. |
 | **6D-6** | `EnableStorageEnvelope` admin RPC + CLI command + integration test | Sixth | Composes 6D-3 + 6D-4 + 6D-5 into the user-visible cutover. End-to-end test exercises a single-node cluster doing Bootstrap → EnableStorageEnvelope → Put → read-back-via-envelope. |
 
--- a/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
+++ b/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
@@ -284,7 +284,12 @@ leader. Server-side sequence:
    entry's apply index, recorded by §6.4) as the response's
    applied_index. This is stable across subsequent
    encryption-relevant Raft entries that advance the generic
-   RaftAppliedIndex.
+   RaftAppliedIndex. See §6.4 for the
+   `cutover_index_unknown = true` defensive-branch fallback
+   when StorageEnvelopeCutoverIndex == 0 on a sidecar that
+   also reports StorageEnvelopeActive == true
+   (operationally impossible under normal apply but hedged
+   against future schema rollback / hand-edited sidecar).
 6. Run the Voters ∪ Learners capability fan-out (§4). On any
    member reporting encryption_capable=false or unreachable
    within one election timeout, return
@@ -682,9 +687,16 @@ until 6D-6 wires it).
   deduplicated.
 - **6D-4 (sub-tag dispatch)**: extend `applier_test.go` —
   invalid `Wrapped != nil` → halt apply. Invalid `Purpose` →
-  halt apply. Mismatched `DEKID` → halt apply. Happy path →
-  `StorageEnvelopeActive` flips true, `RaftAppliedIndex`
-  advances. Idempotent re-apply → no-op (already active).
+  halt apply. Mismatched `DEKID` → halt apply. Happy path
+  (apply at raft index K) → `StorageEnvelopeActive` flips
+  true, `StorageEnvelopeCutoverIndex == K` (the §6.4
+  idempotency token — assertion is load-bearing: without it,
+  the §3.2 / §6.4 `AlreadyExists` contract would silently
+  return 0 as the original cutover index), and
+  `RaftAppliedIndex` advances. Idempotent re-apply at a
+  later index → no-op (already active);
+  `StorageEnvelopeCutoverIndex` stays at K, NOT the
+  re-applying entry's index.
 - **6D-5 (storage toggle)**: stub the sidecar state at
   `StorageEnvelopeActive=true`, write a value, confirm the
   on-disk MVCC version header carries `encryption_state=0b01`
--- a/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
+++ b/docs/design/2026_05_18_proposed_6d_enable_storage_envelope.md
@@ -480,11 +480,18 @@ returns ErrCapabilityCheckFailed."
 ### 5.1 `ErrNodeIDCollision`
 
 **What it catches.** The §4.1 GCM nonce field is 16 bits of
-`node_id` || 16 bits of `local_epoch` || 32 bits of counter.
-`node_id` is derived as `xxhash(full_node_id) & 0xFFFF`. Two
-nodes whose `full_node_id`s happen to hash to the same 16-bit
-slice will reuse nonces under the same DEK with high probability
-within a single epoch — catastrophic for AES-GCM.
+`node_id` || 16 bits of `local_epoch` || 64 bits of counter.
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

Stage 6D-1 — doc-only PR. The encryption Stage 6 cutover that flips a cluster-wide encryption state from OFF to ON. This is the design proposal; implementation lands in subsequent sub-PRs (6D-2 through 6D-6 per §7 of the doc).

Please validate as a self-contained design record. Key things to check:

1. **Sub-PR decomposition** (§7) — six sub-PRs, each independently revertable, only 6D-6 is operator-visible. Is the split clean? Is there work that should bundle differently?

2. **Wire-format addition** (§2) — new sub-tag `RotateSubEnableStorageEnvelope = 0x01` on the existing 0x05 OpRotation payload. Five validation constraints documented (Purpose, DEKID, Wrapped, sidecar state, idempotency). Reuses the existing rotation pipeline rather than adding a new opcode.

3. **Admin RPC contract** (§3) — leader-only, 8-step server-side sequence including the Voters ∪ Learners pre-flight gate. Five typed sentinels for the failure modes. Failure-mode matrix in §8.

4. **Capability fan-out helper** (§4) — fresh probing per call (no caching), default timeout one election timeout, dedupes nodes that serve multiple groups, reused by Stage 6E AND the 6C-3 startup guard.

5. **6C-3 startup guards** (§5) — `ErrNodeIDCollision` and `ErrLocalEpochRollback`. Both gated on `encryptionEnabled`, both skip cleanly pre-bootstrap. Bundled into 6D-2 per the parent doc's rationale.

6. **§6.2 storage-layer toggle** (§6) — `PutAt` reads `StorageEnvelopeActive` at write time. Pre-cutover versions stay cleartext; post-cutover writes are encrypted; mixed versions per §5.4 of parent doc.

7. **Five-lens self-review** (§10) — data loss, concurrency, performance, consistency, test coverage.

8. **Open questions** (§11) — sub-tag byte choice, idempotency response code, fan-out helper location. Defensible defaults picked; reviewer can challenge.

9. **Cross-references** — the doc references PR #784 (6C-2d) as the prerequisite (now merged) and the parent design doc. Verify these are accurate.

10. **No code changes**. CI is doc-only.

bootjp:
@claude review

Round-1 (commit `abf42da9`): addressed gemini high (SubTag collision) + coderabbit Major (NodeIDCollision posture inconsistency).

## gemini high — SubTag byte conflict

The doc proposed `RotateSubEnableStorageEnvelope = 0x01` claiming existing `RotateSubRotateDEK = 0x00`. Verified against `internal/encryption/fsmwire/wire.go`:
- `RotateSubRotateDEK = 0x01` (shipped Stage 4)
- `0x04` already reserved in the comment block for `enable-storage-envelope (Stage 6, §7.1 Phase 1)`

Picking 0x01 would have collided with shipped RotateDEK and broken rolling upgrades (pre-6D apply would see the cutover entry as a malformed RotateDEK with `Wrapped == nil` and halt the FSM). Fix: declare `= 0x04` honoring the existing reservation. Updated §1 in-scope bullet, §2.1 wire-format block, §2.1 layout table, §10 milestone table, and §11.1 open-question.

## coderabbit Major — NodeIDCollision fan-out posture inconsistency

Three sections contradicted each other on whether `ErrNodeIDCollision` uses `GetCapability` fan-out:
- §1 in-scope bullet: "fan out to every Voters ∪ Learners"
- §4.2 caller contract: "reuses the same helper (CapabilityFanout)"
- §5.1 algorithm + rationale: purely local route-catalog walk, explicitly "NO fan-out" because the gRPC server is not up at startup

§5.1's posture is the only physically possible one (no gRPC server → no peer dial). Fix: reconciled §1 and §4.2 to match §5.1. `CapabilityFanout` is now exclusively the cutover-RPC handler's helper; the startup guard does NOT reuse it.

Docs-only change. No code semantics; no caller audit needed.

Remaining findings (gemini medium xxhash collision risk + 2 codex P2 about ErrLocalEpochRollback authority source and probe strategy) are below the loop's P0/P1/high/critical/major threshold and deferred. Happy to address in a follow-up if reviewers prefer them in this PR.

bootjp:
@claude review

PR786 round-2 (commit `542b24ba`): addressed all four findings.

## SubTag value: 0x01 → 0x04 (gemini high)

Initial doc proposed `0x02` (and the original draft had `0x01` which collided with `RotateSubRotateDEK`). Both were wrong — `internal/encryption/fsmwire/wire.go:108` already reserves `0x04 — enable-storage-envelope (Stage 6, §7.1 Phase 1)` in a comment block. Using any value other than `0x04` would have either collided with the existing RotateDEK byte (causing pre-6D apply paths to interpret a cutover entry as a malformed RotateDEK and halt the FSM — a rolling-upgrade-class regression) or required moving the reservation block (breaks the audit trail).

Doc now reflects `0x04` consistently in five locations (§1 scope, §2.1 declaration block + wire layout table, §7 sub-PR table, §11.1 open question). The rolling-upgrade hazard rationale is documented in §11.1.

## ErrNodeIDCollision posture: local-snapshot-only (coderabbit major + codex P2)

Three doc sections contradicted each other (§1 said fan-out, §4.2 said reuses CapabilityFanout, §5.1 said local-snapshot-only). All three now agree on local-snapshot-only. The §5.1 lifecycle rationale ("gRPC server not yet up at startup-guard phase") is the load-bearing reason; §1 and §4.2 cross-reference it.

## ErrLocalEpochRollback posture: local Pebble read, no RPC (codex P2)

§1 said the guard compares against "the leader's writer-registry record via ResyncSidecar"; §5.2 said it reads the local Pebble writer-registry directly. The writer-registry is local state on every node post-bootstrap (every node applies the same OpRegistration and OpBootstrap entries); no RPC is required at the startup-guard phase. §1 updated to match §5.2's correct algorithm.

## Collision-risk mitigation documented (gemini medium)

New "Collision probability and mitigation" subsection in §5.1 covers:
1. Re-rolling the colliding member's full_node_id (operator picks a different uint64; deterministic remap)
2. Coordinated ID assignment for new deployments (6D-2 will ship `elastickv-admin encryption probe-node-id --full-node-id=<u64>`)
3. Future widening of the 16-bit node_id field (out of scope; documented so future work knows the ceiling exists)

The birthday-bound risk is quantified: ~0.2% at N=32, ~3% at N=128, ~50% at N=300. For typical Raft clusters (3-7 voters + small learner pool) the probability is well under 0.01%.

## What's still doc-only

No code changes; this is the design-only PR. Wire-format declaration of `RotateSubEnableStorageEnvelope = 0x04` (along with the actual wire-format change to `wire.go`) lands in Stage 6D-4 per the §7 sub-PR table.

bootjp:
@claude review

Round-2 (commit `a64aa296`): addressed codex r2 P1 — ErrLocalEpochRollback strict-ahead threshold.

The proposed check rejected only when `sidecar < registry`, but the §4.1 nonce-monotonicity invariant requires `sidecar > registry` (strictly ahead) before this node resumes issuing nonces. An equality match means the sidecar's `local_epoch` hasn't advanced past the value the writer registry already saw the previous boot use — restarting on that sidecar would reissue `node_id || local_epoch` with the GCM counter reset to zero and recycle previously-used nonces under the same DEK.

Fix: tightened the §5.2 "How it works" step 3 to `sidecar <= registry → refuse`. Added a rationale paragraph explaining why `<=` (not `<`) is the right boundary: on a normal restart the applier increments the sidecar's epoch past whatever the registry recorded, so `sidecar > registry` is the happy-path invariant.

Docs-only change. No code semantics affected; no caller audit needed.

Remaining codex P2 findings (line 171 sub-tag wording in upgrade-compat note, line 369 collision probability exact percentages) are below the loop's P0/P1/high/critical/major threshold. Happy to address in a follow-up if reviewers prefer them in this PR.

bootjp:
@claude review

Round-3 (commit `20d8f618`) — two follow-ups:

## codex P2 — collision-probability arithmetic

Previous numbers in §5.1 were off by roughly a factor of 4:
- N=32: doc said 0.2%, actual ≈ 0.8% (per N·(N−1)/(2·65536))
- N=128: doc said 3%, actual ≈ 12.5%
- N=300: ~50% (this was approximately right; canonical √(2·d·ln 2) ≈ 302)

Fix: replaced the approximations with the proper birthday-bound expression N·(N−1)/(2·2¹⁶) and recomputed each anchor. Added clarification that small clusters (N≤8) stay under 0.05% so operators reading the section don't conflate the theoretical-N=300 risk with their actual deployment scale.

## Lost  fix from r1

My r1 fix replaced a stale `BootstrapAdmin` mutator name reference in §2.1 with `EnableStorageEnvelope` (the actual RPC name). The conflict resolution when rebasing onto the parallel claude[bot] commit took the remote's version which hadn't made that fix. Re-applied on this commit.

## codex P1 (ErrLocalEpochRollback strict-ahead) — already addressed

The parallel claude[bot] commit `a64aa296` (now in the branch via fast-forward) tightened the check from `sidecar < registry` to `sidecar <= registry` per the nonce-monotonicity contract. That commit is in the branch and is part of the round-2 changes.

## What's still doc-only

No code changes. Wire-format declaration of `RotateSubEnableStorageEnvelope = 0x04` ships in Stage 6D-4.

bootjp:
@claude review

Round-4 (commit `82504a51`): addressed codex P1 on round-3.

## codex P1 — test plan strict-ahead

The §9 test plan row for `ErrLocalEpochRollback` said "Sidecar matches → nil", which contradicted §5.2's strict-ahead semantics from round-2 (commit `a64aa296`). The §5.2 contract requires the guard to reject when `sidecar <= registry` (including equality), because the equality case would replay the same `(node_id, local_epoch)` prefix and reuse the GCM counter under the same DEK.

If implementers had followed the original test-plan row, the unit test for the equality case would have asserted nil and let a regression slip through.

Fix: rewrote the test-plan row to align with §5.2 — the equality case is now an expected-error case, and the no-registry-record-yet case is documented as nil (the freshly-joined-learner skip condition from §5.2).

## What's still doc-only

No code changes. The actual unit tests land with Stage 6D-2.

bootjp:
@claude review

Round-5 (commit `9ef584b1`): addressed all five carryover items claude flagged on round-4.

1. **§1 vs §5.2 `<=` threshold contradiction** — §1 `ErrLocalEpochRollback` bullet now reads `Refuse if sidecar <= registry (strict-ahead — see §5.2 for the equality-case rationale)`, matching §5.2 step 3.

2. **§2.2 stale `0x01` reference** — corrected to `0x04`. Added explicit clarification that the already-shipped `SubTag = 0x01` for `RotateSubRotateDEK` is unaffected.

3. **§11.3 stale `ErrNodeIDCollision` reuse claim** — dropped the claim from the open-question 3 justification and noted that the guard consults the local route-catalog snapshot directly. The Stage 6E reuse justification stands on its own.

4. **§7 table cross-reference typo** — 6D-3 row corrected from "cutover RPC (6D-5)" to "cutover RPC (6D-6)".

5. **§2.2 decoder path gap** — new "Two-site whitelist change required in 6D-4" paragraph documents both the decoder (`readRotationSubTag` in `wire.go:499`) AND the applier switch. The same constraint is flagged for future `rewrap-deks` / `retire-dek` / `enable-raft-envelope` additions so future sub-PRs don't repeat the gap.

No code changes; doc-only update.

bootjp:
@claude review

Round-6 (commit `1b7abc95`): addressed codex r5 P1 — flipped `ProposerRegistration.LocalEpoch` inequality direction in §2.1 rotation-entry table.

The constraint read `LocalEpoch ≤ current registry record`, which directly contradicts §5.2's strict-ahead nonce-safety invariant. Enforcing `≤` would either reject every healthy cutover after a normal restart (where the sidecar has advanced past the registry) or force the proposer into a circular refusal loop (regress the epoch to satisfy `≤`, then apply-time §5.2 fires `ErrLocalEpochRollback`).

Fix: flip the inequality to `≥` with a parenthetical explaining the equality case (same-boot proposer that just wrote the registry) is legitimate. A proposer with `LocalEpoch < registry` is stale and the proposal is rejected.

Docs-only change; no code semantics affected; no caller audit needed.

Remaining codex P2 (line 260, persist cutover index for idempotent retries) is below the loop's P0/P1/high/critical/major threshold.

bootjp:
@claude review

Round-6 (commit `82fd1055`): addressed both codex findings on round-5.

## codex P1 (line 145) — proposer LocalEpoch inequality

The original `<=` constraint was backwards (would reject healthy cutovers immediately after a normal restart). Parallel claude[bot] commit `1b7abc95` already pushed `>=` to the branch; that fix is now in the round-6 history. My commit reasoning ('exact equality') was also wrong — `>=` accommodates both the same-boot proposer (equality) and the post-restart proposer (strictly ahead) cases. The rebase took the remote's `>=` version.

## codex P2 (line 260) — stable cutover-index for AlreadyExists

Deferred by `1b7abc95` (below the P0/P1 threshold per that commit's note); addressed in this round-6 commit because the RPC contract becomes ambiguous without it. Changes:

- New §6.4 subsection defining `StorageEnvelopeCutoverIndex uint64` sidecar field symmetric with `RaftEnvelopeCutoverIndex`. Apply-time semantics document the three writes (StorageEnvelopeActive + StorageEnvelopeCutoverIndex + RaftAppliedIndex) landing inside one WriteSidecar fsync via §5.1 write-and-rename.
- §1 FSM-apply-path bullet now mentions both writes.
- §3.1 proto `EnableStorageEnvelopeResponse` has explicit comments on `applied_index` semantics (fresh-cutover vs AlreadyExists retry), and a new `cutover_index_unknown bool` field for the defensive 'StorageEnvelopeActive=true but cutover_index=0' branch (operationally impossible but hedges against schema rollback / hand-edited sidecar).
- §3.2 step 5 reads from `sidecar.StorageEnvelopeCutoverIndex` directly.
- §7 6D-4 row mentions both writes + the `readRotationSubTag` whitelist update.
- §8 idempotent-retry row updated to describe the AlreadyExists response as carrying the stable index.

The contract is now: `applied_index` on AlreadyExists is the ORIGINAL cutover entry's apply index, stable across arbitrary subsequent encryption activity. The CLI surfaces `cutover_index_unknown` as a warning if the defensive branch fires.

No code changes; doc-only update.

bootjp:
@claude review

Round-7 (commit `168b4f5a`): addressed both claude findings on round-6.

## claude P1 — §9 test plan 6D-4 row

Rewrote the happy-path assertion to cover all three writes from §6.4:
- `StorageEnvelopeActive = true`
- `StorageEnvelopeCutoverIndex == K` (with explicit "load-bearing" annotation — without this, the idempotency contract silently returns 0)
- `RaftAppliedIndex` advances

Also added the idempotent re-apply case: `StorageEnvelopeCutoverIndex` stays at K (the original), NOT the re-applying entry's index. This is the dual of the AlreadyExists contract from §6.4.

## claude minor — §3.2 step 5 cross-reference

Appended a cross-reference to §6.4's `cutover_index_unknown` defensive branch so the 6D-6 implementor reading the step-by-step server sequence doesn't need to discover the special case in another section.

No code changes; doc-only update.

bootjp:
@claude review

Round-8 (commit `091dc3ec`): addressed codex P1 from round-7.

## codex P1 — gRPC unary semantics drop body on non-OK

The spec required `EnableStorageEnvelope` to return `AlreadyExists` AND carry `applied_index` in the normal response, but unary gRPC's `(*Response, error)` shape only delivers the message when `err == nil`. Returning AlreadyExists sets a non-OK status and drops the response body, so the documented idempotency contract was un-implementable as written.

## Fix: OK with `was_already_active` discriminator

The idempotency contract now lives entirely on the success path:

- Fresh cutover: `OK`, `was_already_active=false`, `applied_index=K`, `capability_summary` populated
- Idempotent retry: `OK`, `was_already_active=true`, `applied_index=sidecar.StorageEnvelopeCutoverIndex`, `capability_summary` empty (the original cutover's view isn't retained)

Sites updated:
- §3.1 proto: new `was_already_active bool = 4` field with explicit comment about why OK rather than AlreadyExists
- §3.2 step 5: "return gRPC OK with was_already_active=true" + load-bearing note about unary gRPC dropping body on non-OK
- §3.2 outcome table: heading now "Server-side outcomes" (the idempotent-retry row is a success, not a failure)
- §6.4 idempotency contract: rewrote without AlreadyExists; cross-refs §11.2 for rationale
- §8 refusal-posture table: idempotent-retry row updated
- §9 test plan 6D-4: removed the AlreadyExists reference from the load-bearing note
- §10 self-review: "returns AlreadyExists" → "returns OK with was_already_active=true"
- §11.2 open question: rewrote to document the OK-with-bool choice and explain why AlreadyExists is un-implementable. Mentions the rejected third alternative (AlreadyExists + `status.Status` details proto attachment).

Plus an optimization: skip step 6 (capability fan-out) on the `was_already_active=true` path — the original cutover already passed the gate; re-running it would add latency to no-op calls.

No code changes; doc-only update.

bootjp:
@claude review

Round-9 (commit `f6b906aa`): closed out the two minor doc-consistency observations from round-8.

## claude minor — `cutover_index_unknown` scope explicit

Appended a paragraph to the §3.1 proto comment clarifying that `cutover_index_unknown` is ONLY relevant when `was_already_active=true`. On a fresh cutover, `applied_index` comes directly from the Raft apply result (not from reading `sidecar.StorageEnvelopeCutoverIndex`), so the defensive branch cannot fire.

## coderabbit minor — `cutover_index_unknown` in §9 test plan

Expanded the §9 6D-6 test plan to pin two new assertion blocks:
- Idempotent-retry happy path: stub sidecar with active+K, assert OK + was_already_active=true + applied_index=K + cutover_index_unknown=false + capability_summary empty
- Defensive branch: stub sidecar at active+0 (operationally impossible case), assert OK + was_already_active=true + cutover_index_unknown=true + applied_index = sidecar.RaftAppliedIndex

Both findings were below P0/P1 priority but the fixes are tiny and close out the doc-only iterations.

No code changes.

bootjp:
@claude review

Round-10 (commit `61836de6`): addressed codex r8 P1 — refuse missing-registry-row when cutover is active.

The §5.2 skip conditions silently passed nodes with no writer-registry row even after the cutover had fired. Pre-cutover this is correct (no nonces issued yet → no rollback anchor needed). Post-cutover a rowless node would start issuing GCM nonces from `local_epoch=0` with no anchor at all — directly defeating the §4.1 invariant the guard exists to enforce.

Fix: split the missing-registry-row behaviour on `StorageEnvelopeActive`:
- pre-cutover (false): allow (freshly-joined-learner case);
- post-cutover (true): refuse with `missing_registry_row=true` detail. Operator recovers by replaying `RegisterEncryptionWriter` (§4.1 case-2 idempotent).

Updated §5.2 "Missing-registry-row posture" subsection and §10 test plan for 6D-2 `ErrLocalEpochRollback`.

Docs-only change; no caller audit needed.

Remaining findings (coderabbit minor line 743 `cutover_index_unknown` test, codex P2 line 221 non-zero proposer_node_id) are below the loop's P0/P1/high/critical/major threshold.

bootjp:
@claude review

Round-11 (commit `6fe59f23`): closed out codex P2 from round-9.

## codex r9 P2 — non-zero `proposer_node_id` validation

The RPC contract validated only `proposer_local_epoch <= 0xFFFF` and did not require `proposer_node_id != 0`. Zero is a reserved sentinel across all mutator paths (BootstrapEncryption, RotateDEK, RegisterEncryptionWriter all reject it via `ErrZeroFullNodeID`). Accepting it on the cutover request would weaken the writer-registry collision and rollback invariants.

Three sites updated:
- §3.1 proto comment lists BOTH validation requirements with gRPC error code + wrapped sentinel
- §3.2 step 1 lists both checks explicitly; notes ApplyRotation re-validates at apply time
- §3.2 outcome table adds two new `InvalidArgument` rows (zero `proposer_node_id` + out-of-range `proposer_local_epoch`)

## codex r9 P1 — missing-registry-row when cutover is active

Already addressed in commit `61836de6` (parallel claude[bot] commit, fast-forwarded into this branch).

No code changes; doc-only update.

bootjp:
@claude review

Round-12 (commit `767b5d9f`): addressed codex r10 P1 ×2 — concurrent cutover races on §2.1 constraints #3 and #4.

## codex P1 — DEK race halts apply (line 164)

The constraint `DEKID == sidecar.Active.Storage at apply time` was routed through `ErrEncryptionApply` (halt-on-apply) on mismatch. If `RotateDEK` commits between cutover propose and apply, the active DEK advances and the cutover's DEKID no longer matches; halting the apply loop on this normal admin race would escalate a routine retry into a cluster-stopping condition.

Fix: §2.1 constraint #3 now documents the apply path consuming the entry as a benign no-op (advance `RaftAppliedIndex`, record `ErrCutoverDEKIDStale` outcome on §6.4 detail ride-along) and the mutator surfacing it as `FailedPrecondition: ErrCutoverDEKIDStale` to the operator with a retry hint. Retryable precondition, not halt.

## codex P1 — concurrent-overlap idempotency (line 169)

The §6.4 stable-token guarantee was breakable by two overlapping `EnableStorageEnvelope` calls: both could pass pre-checks and commit, and the second call's applied_index observation would differ from the first's.

Fix: §2.1 constraint #4 now states the two layers closing the race:
- **Apply-side no-overwrite**: `StorageEnvelopeCutoverIndex` is set on the FIRST apply only; subsequent applies consume the entry without re-writing.
- **Propose-side mutator lock**: cutover-RPC handler serializes overlapping admin calls on a per-default-group mutator lock; `RotateDEK` and `EnableStorageEnvelope` share the same lock so they cannot interleave between propose and apply on a single leader.

Docs-only change; no caller audit needed.

bootjp:
@claude review

Round-13 (commit `1e34a00c`): addressed codex r12 P1 — aligned §10 6D-4 test plan with the §2.1 non-halting apply contract.

The test plan still said "Mismatched `DEKID` → halt apply", contradicting §2.1 constraint #3 added in r11 (DEKID mismatch is a benign stale-cutover consumed without halt). Test plan now mirrors §2.1: `RaftAppliedIndex` advances, `StorageEnvelopeActive` stays false, entry recorded as `ErrCutoverDEKIDStale` on the §6.4 detail ride-along.

Docs-only; no caller audit needed.

bootjp:
@claude review

Round-14 (commit `4c5b09e5`): addressed codex r13 P1 — `node_id` derivation must match the shipped `uint16(full_node_id)` narrowing.

The §5.1 guard algorithm derived `node_id` as `xxhash(full_node_id) & 0xFFFF`, but the existing writer-registry keying and nonce-prefix logic already uses `uint16(full_node_id)` (low 16 bits of the FNV-1a 64-bit hash) — verified in `internal/encryption/applier.go:51` and `adapter/encryption_admin.go:64`. A guard hashing differently from the runtime would check a different 16-bit space than the one actually used for nonce uniqueness: real collisions would slip past (silent nonce reuse) and false collisions would be reported (refuse-to-boot on safe configs).

Fix: replaced every `xxhash(...) & 0xFFFF` in §5.1 with `uint16(full_node_id)`. Touched "What it catches" (added the must-match-shipped-narrowing constraint), "How it works" step 2, "Collision probability", and mitigation #2. Also corrected the §5.1 nonce-layout description to 16/16/64 (was 16/16/32) per codex P2 at line 483 — the counter is 64 bits to support ~2^64 writes per (DEK, local_epoch).

Docs-only; no caller audit needed.

bootjp:
@claude review

Round-15 (commit `fff0aae2`): closed out two codex P2 stale-text findings from round-14.

## codex P2 (line 201) — stale AlreadyExists retry-path

§2.1 concurrency analysis paragraph still pointed at "§3.2 AlreadyExists path" after round-6 moved retries to OK + was_already_active. Rewrote to reference "§3.2 idempotent-retry path (gRPC OK with was_already_active=true)".

## codex P2 (line 404) — idempotency in halt-apply row

§3.2 outcome table row lumped "idempotency" under ErrEncryptionApply halt cases. The round-12 decision (§2.1 constraint #4) defines duplicate cutover entries as a benign consumed no-op. Rewrote the row to enumerate actual halt causes (decode error, Purpose != Storage, Wrapped != nil, etc.) and explicitly excluded the duplicate-entry case with a forward-reference to §2.1.

Both were P2 (below cron's P0/P1/major threshold) but tiny one-line fixes that close out the doc-only iterations.

No code changes.

bootjp:
@claude review

Round-16 (commit `1e10a69f`): closed out both carryover items from claude r15.

## §5.2 cross-reference §6.4 → §6.2

`StorageEnvelopeActive` is defined in §6.2 (write-path toggle), not §6.4 (which is the adjacent `StorageEnvelopeCutoverIndex` field). Updated the §5.2 cross-reference with a parenthetical clarifying both fields' locations so a reader who lands in §6.4 by mistake still finds the related field.

## `ErrZeroFullNodeID` / `ErrLocalEpochOutOfRange` not Go symbols

Verified — `grep` returns zero hits in `.go` files. `adapter/encryption_admin.go:554–565` uses inline string errors for the same checks. Picked **Option B** (consistent with existing path): each error is described as an inline `InvalidArgument` carrying a specific message, matching the existing `RotateDEK` / `BootstrapEncryption` convention.

Three sites updated:
- §3.1 proto comment — both error checks described as inline `InvalidArgument` with message text
- §3.2 step 1 — both checks reject with inline  carrying the specified message
- §3.2 outcome table — both rows describe inline `InvalidArgument` and reference the existing inline-error convention

A future cleanup could introduce shared sentinels and migrate all three sites; that's documented as out of scope for 6D.

No code changes; doc-only update.

bootjp:
@claude review

(Retry — previous round-16 review encountered an error at https://github.com/bootjp/elastickv/actions/runs/26011881679 .)

Round-16 (commit `1e10a69f`): closed out both carryover items from claude r15:

1. **§5.2 cross-reference §6.4 → §6.2** for `StorageEnvelopeActive`. Added a parenthetical clarifying both fields' locations.

2. **`ErrZeroFullNodeID` / `ErrLocalEpochOutOfRange` not Go symbols** — picked Option B (inline InvalidArgument with message, matching existing RotateDEK / BootstrapEncryption convention). Three sites updated: §3.1 proto comment, §3.2 step 1, §3.2 outcome table.

No code changes; doc-only update.

bootjp:
@claude review

Round-17 (commit `37520e2f`): addressed codex r16 P1 — validate empty wrapped by length, not nil.

The §2.1 constraint #2 read `Wrapped == nil` for the cutover's empty-wrapped requirement. But the wire decoder (`internal/encryption/fsmwire/wire.go::readLenU32Bytes`) materializes zero-length payloads as an allocated empty slice (`[]byte{}`), not `nil`. Implementing as written would classify every valid cutover entry with `wrapped_len=0` as malformed and halt-apply the cluster.

Fix: changed §2.1 constraint #2 to `len(Wrapped) == 0` with an explanatory note pinning the decoder shape. Also updated §10 6D-4 test row to assert `len(Wrapped) > 0 → halt apply" (not `Wrapped != nil`), with an explicit note that `Wrapped == []byte{}` from the wire decoder must NOT halt.

Docs-only; no caller audit needed.

bootjp:
@claude review

Round-17 cleanup (commit `f6e25364`): the parallel claude[bot] commit `37520e2f` addressed the same codex P1 (len(Wrapped) vs nil) but only updated §2.1 constraint #2; two propagation sites were left stale. Re-applied:

- §2.1 table row (line 147): added the "`len(Wrapped) == 0`, NOT `Wrapped == nil`" annotation with cross-reference to constraint #2
- §3.2 outcome table halt-apply row (line 418): "`Wrapped != nil`" → "`len(Wrapped) > 0`"

The §9 test-plan row was already updated by the remote commit to use the length-based check.

The previous two claude review attempts (round-16 + retry) encountered job errors; this push is a chance to re-trigger.

No code changes; doc-only update.

bootjp:
@claude review

bootjp:
@claude review

Round-19 (commit `a990e7a9`): addressed the last P1 carryover from round-18.

## claude r18 P1 — §3.2 outcome table missing `ErrCutoverDEKIDStale` row

First flagged in R12, fixed in §2.1 + §9 in R13, but the outcome-table row was silently dropped from tracking. A 6D-6 implementor reading only the outcome table would have routed DEKID mismatch to the "Apply halted" catch-all (`ErrEncryptionApply`), reintroducing the cluster-stopping halt that §2.1 constraint #3 was specifically written to prevent.

Fix: added one row between "Capability fan-out failed" and "Apply halted":

> DEKID stale at apply (`RotateDEK` raced) | `FailedPrecondition` | new `ErrCutoverDEKIDStale` — benign no-op at FSM: `RaftAppliedIndex` advances, `StorageEnvelopeActive` stays false, no halt. Handler surfaces `FailedPrecondition` to operator with retry hint. See §2.1 constraint #3.

Also updated the "Apply halted" row to explicitly exclude both benign-no-op cases (constraint #3 + constraint #4).

Per the claude r18 verdict, this is the last P1 from the 17-round history. No code changes; doc-only update.

bootjp:
@claude review

bootjp:
@claude review

PR #789, opened 2026-05-18
fix(redis): NOTLEADER prefix for gRPC-wrapped leader errors (closes Jepsen run 26035515694 crash)

bootjp:
## Summary
Fix Redis worker crashes seen in scheduled Jepsen run [26035515694](https://github.com/bootjp/elastickv/actions/runs/26035515694/job/76532418404) — many workers died with `IllegalArgumentException: Don't know how to create ISeq from: clojure.lang.ExceptionInfo` because the carmine error carried `:prefix :rpc`, which `jepsen-io/redis`'s `with-exceptions` does not catch.

PR #753 introduced the `NOTLEADER` prefix and `isTransientLeaderRedisError` classifier (typed via `errors.Is`). After gRPC boundary the typed chain is stripped and the error becomes the literal string `"rpc error: code = Unknown desc = leader not found"` — the typed classifier missed it.

This PR adds a **suffix-match fallback** to `isTransientLeaderRedisError`, mirroring `kv.hasTransientLeaderPhrase`. The phrase set is closed and `HasSuffix`-based — the same load-bearing discipline used in `kv/coordinator.go` to avoid misclassifying a user-controlled key like `"key: not leader: write conflict"`.

## Why this matters for the run-26035515694 "Analysis invalid" verdict
The run also showed apparent list-ordering anomalies (e.g. `[1 2 3 4 6 7 8 9 11 12 13 10 …]`). Workers crashing caused many transactions to be marked `:type :info` (outcome unknown), which inflates the checker's solution space and can produce spurious anomalies. Fixing the worker crash is the prerequisite — after this lands, a re-run will tell whether the list-ordering signal is real or an artifact.

## Caller audit
`isTransientLeaderRedisError` has one caller: `writeRedisError`. The change is **strictly additive** — errors that were previously classified `true` still are; errors with a transient-leader suffix on a stripped chain are now also `true`. No error loses the prefix. Carmine semantics improve from `:prefix :rpc` (uncaught → worker crash) to `:prefix :notleader` (caught by `with-exceptions` → mapped to `:type :fail :error :notleader`).

## Self-review (5 lenses)
1. **Data loss** — N/A (Redis error reply prefix only).
2. **Concurrency** — pure read; -race clean.
3. **Performance** — error path only; one `strings.ToLower` + 4 `HasSuffix` per error reply.
4. **Data consistency** — wire format only; sentinels unchanged.
5. **Test coverage** — six new classifier cases (4 gRPC-wrapped variants + 1 false-positive negative), one new wire-reply case.

## Test plan
- [x] `go test -race -count=1 -short ./adapter` — 566s, all green
- [x] `go vet ./adapter` — clean
- [x] `golangci-lint run --config=.golangci.yaml ./adapter` — 0 issues
- [ ] CI green
- [ ] Re-run scheduled Jepsen Redis workload to confirm worker crashes resolved + check whether list anomaly persists

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of transient Redis leadership errors, including cases where errors are wrapped by transport layers and lose their original type. Added a safe suffix-matching fallback to correctly classify leader-related errors while avoiding misclassification of user-controlled text.

* **Tests**
  * Extended regression tests to cover wrapped/substring leadership error scenarios and ensure correct error-prefixing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/789?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -355,12 +355,58 @@ func isTransientLeaderRedisError(err error) bool {
 	if err == nil {
 		return false
 	}
-	return errors.Is(err, ErrLeaderNotFound) ||
+	if errors.Is(err, ErrLeaderNotFound) ||
 		errors.Is(err, ErrNotLeader) ||
 		errors.Is(err, kv.ErrLeaderNotFound) ||
 		errors.Is(err, raftengine.ErrNotLeader) ||
 		errors.Is(err, raftengine.ErrLeadershipLost) ||
-		errors.Is(err, raftengine.ErrLeadershipTransferInProgress)
+		errors.Is(err, raftengine.ErrLeadershipTransferInProgress) {
+		return true
+	}
+	// Suffix fallback for gRPC-wrapped sentinels. When the coordinator
+	// forwards a request to a remote leader via the operational gRPC
+	// service and that leader returns ErrLeaderNotFound, the status
+	// interceptor flattens the error to "rpc error: code = Unknown
+	// desc = leader not found"; the typed sentinel chain is stripped
+	// at the wire boundary, so errors.Is misses it. The Jepsen Redis
+	// workload (scheduled run 26035515694) saw workers crash with
+	// `:prefix :rpc` because the un-prefixed `"rpc error: …"` string
+	// reached Carmine. Match the same closed phrase set
+	// kv.hasTransientLeaderPhrase uses, with the same HasSuffix
+	// guard: free-form Contains would misclassify a user-controlled
+	// key like "key: not leader: write conflict" as transient.
+	return hasTransientLeaderSuffix(err.Error())
+}
+
+// redisLeaderErrorPhrases mirrors kv.leaderErrorPhrases (the kv
+// package keeps it unexported). Any new transient-leader phrase the
+// kv layer treats as retryable should also flip a NOTLEADER prefix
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

Round-1 fix:
- **gemini medium (alloc)**: replaced `strings.ToLower(msg) + HasSuffix` with `strings.EqualFold(msg[len(msg)-len(phrase):], phrase)` per the gemini snippet. Avoids the full-message copy on error path (cockroachdb stack-traces can be multi-KB). Caller audit clean — internal helper, same return contract.

Tests green, lint 0 issues.

bootjp:
@claude review

Round-2: closed the sentinel-pinning gap you noted.

Added `TestHasTransientLeaderSuffix_PinsSentinels` in `adapter/redis_error_prefix_test.go` — calls `hasTransientLeaderSuffix(sentinel.Error())` against all six transient-leader sentinels (adapter.ErrLeaderNotFound / ErrNotLeader, kv.ErrLeaderNotFound, raftengine.ErrNotLeader / ErrLeadershipLost / ErrLeadershipTransferInProgress) to catch a future rename. Mirrors the comment at `kv/coordinator.go:529`. Tests green, lint 0 issues.

PR #790, opened 2026-05-19
backup: Redis zset encoder (Phase 0a)

bootjp:
## Summary

Adds the Redis sorted-set (zset) encoder for the Phase 0 logical
snapshot decoder (`docs/design/2026_04_29_proposed_snapshot_logical_decoder.md`,
lines 334-335). Mirrors the hash/list/set encoders shipped in
#725/#755/#758. After this lands, Phase 0a's remaining Redis work
is the stream encoder + HLL TTL sidecar; the per-adapter encoders
themselves will be complete.

Output shape per the design:

```json
{"format_version": 1,
 "members": [{"member": "alice", "score": 100}],
 "expire_at_ms": null}
```

Members sorted by raw byte order (not by score) so `diff -r` between
two dumps with the same logical contents stays line-stable across
score-only mutations. Scores emit as JSON numbers for finite values,
ZADD-conventional `"+inf"` / `"-inf"` strings for ±Inf (`json.Marshal`
rejects non-finite floats). NaN scores fail closed at intake — Redis's
ZADD rejects NaN at the wire level, so a NaN at backup time indicates
corruption.

Wire layout (mirrors `store/zset_helpers.go`):
- `!zs|meta|<userKeyLen(4)><userKey>` → 8-byte BE Len
- `!zs|mem|<userKeyLen(4)><userKey><member>` → 8-byte IEEE 754 score
- `!zs|scr|<userKeyLen(4)><userKey><sortableScore(8)><member>` →
  silently discarded (secondary index; `!zs|mem|` is the source of
  truth)
- `!zs|meta|d|<userKeyLen(4)><userKey><commitTS(8)><seqInTxn(4)>` →
  silently skipped (delta family; same policy as hash/set encoders)

TTL routing: `!redis|ttl|<userKey>` for a registered zset folds into
the JSON `expire_at_ms` field, matching the set/list/hash inlining,
so a restorer replays ZADD + EXPIRE in one shot.

## Self-review (5 lenses)

1. **Data loss** — `!zs|mem|` is the source of truth; `!zs|scr|` and
   delta records are intentional silent skips with audit notes.
   Empty zsets (Len=0 but meta seen) still emit a file because
   `TYPE k → zset` is observable to clients.
2. **Concurrency / distributed** — `RedisDB` is sequential per scope
   (matches the existing per-DB encoder contract); no shared state.
3. **Performance** — per-zset state in a map, flushed once at Finalize.
   Bounded by `maxWideColumnItems` on the live side; sort is O(n log n)
   on member-name bytes. Identical cost shape to hash/set encoders.
4. **Data consistency** — JSON field order pinned via struct tags
   (not map). Inf score uses `json.RawMessage` so the same `score`
   key emits as either a number or a string. NaN fails closed.
5. **Test coverage** — 17 table-driven tests under
   `internal/backup/redis_zset_test.go`:
   - round-trip basic / empty / TTL inlining
   - binary member via base64 envelope
   - delta-key skip (both `HandleZSetMeta` entry + `parseZSetMetaKey` guard)
   - `HandleZSetMetaDelta` explicit entry
   - `HandleZSetScore` silent-discard
   - malformed meta length / overflow / MaxInt64 boundary
   - malformed member-value length / NaN rejection
   - ±Inf string-form serialization
   - members-without-meta still emits file
   - duplicate-members latest-wins (ZADD semantics)

## Caller audit (per `/loop` standing instruction)

Semantics-changing edit: new `case redisKindZSet:` branch in
`HandleTTL` (redis_string.go:310). Purely additive — the new branch
fires only when `zsetState()` has previously registered the key.
No existing handler maps to `redisKindZSet`, so no prior call site
changes behavior. Verified:

```
grep -n 'redisKindZSet' internal/backup/
# internal/backup/redis_string.go:88
# internal/backup/redis_string.go:310
# internal/backup/redis_zset.go:166
```

Three references, all new in this PR. No legacy caller assumes the
prior behavior.

## Test plan

- [x] `go test -race ./internal/backup/` → ok
- [x] `golangci-lint run ./internal/backup/...` → 0 issues
- [x] `go build ./...` → ok
- [x] `go vet ./internal/backup/...` → ok

## Summary by CodeRabbit

* **New Features**
  * Added Redis sorted-set (ZSET) backup/restore with deterministic JSON files, binary-safe member encoding, raw-byte member sorting, duplicate "latest-wins" semantics, and ±Inf score serialization as strings.
  * TTL handling: per-key TTLs can be inlined into ZSET/Set outputs; TTLs arriving before typed rows are buffered and drained when the key is first seen.
  * Orphan-TTL buffering with configurable byte-cap, fail-closed overflow behavior, and an option to disable buffering.

* **Tests**
  * Extensive test suite covering ZSET semantics, TTL buffering/drain, legacy-blob reconciliation, byte-cap edge cases, and overflow warnings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/790?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/backup/redis_hash.go
+++ b/internal/backup/redis_hash.go
@@ -237,7 +237,11 @@ type hashFieldRecord struct {
 	Value json.RawMessage `json:"value"`
 }
 
-func marshalHashJSON(st *redisHashState) ([]byte, error) {
+// nolint comment lives at the function head: dupl pairs this with
+// marshalZSetJSON, which carries the rationale (parallel design-spec
+// wrappers that can't collapse into a shared helper without breaking
+// JSON field-order determinism). See redis_zset.go:marshalZSetJSON.
+func marshalHashJSON(st *redisHashState) ([]byte, error) { //nolint:dupl // see comment above + redis_zset.go
 	// Sort by raw byte order for deterministic output across runs.
 	names := make([]string, 0, len(st.fields))
 	for name := range st.fields {
--- a/internal/backup/redis_set.go
+++ b/internal/backup/redis_set.go
@@ -114,12 +114,24 @@ func (r *RedisDB) HandleSetMetaDelta(_, _ []byte) error { return nil }
 // setState lazily creates per-key state. Mirrors the hash/list
 // kindByKey-registration pattern so HandleSetMeta, HandleSetMember,
 // and the HandleTTL back-edge all agree on the kind.
+//
+// On first registration we drain any pendingTTL for the user key.
+// `!redis|ttl|<k>` lex-sorts BEFORE `!st|...` (because `r` < `s`),
+// so in real snapshot order the TTL arrives FIRST; HandleTTL parks
+// it in pendingTTL, and this function inlines it into the set's
+// `expire_at_ms`. Without this drain step, every TTL'd set would
+// restore as permanent — a latent bug in PR #758 surfaced by codex
+// on PR #790. Phase 0a tests added in the same PR pin the ordering.
 func (r *RedisDB) setState(userKey []byte) *redisSetState {
 	uk := string(userKey)
 	if st, ok := r.sets[uk]; ok {
 		return st
 	}
 	st := &redisSetState{members: make(map[string]struct{})}
+	if expireAtMs, ok := r.claimPendingTTL(userKey); ok {
+		st.expireAtMs = expireAtMs
+		st.hasTTL = true
+	}
 	r.sets[uk] = st
 	r.kindByKey[uk] = redisKindSet
 	return st
--- a/internal/backup/redis_string.go
+++ b/internal/backup/redis_string.go
@@ -584,7 +584,15 @@ func (r *RedisDB) Finalize() error {
 	// key whose type prefix we don't recognise (a future Redis
 	// type added on the live side without a backup-encoder update).
 	r.orphanTTLCount += len(r.pendingTTL)
-	if r.warn != nil && r.orphanTTLCount > 0 {
+	// Codex P2 (PR #790 r13): also warn on pendingTTLOverflow even
+	// when orphanTTLCount is zero. Overflow entries fail-closed at
+	// parkUnknownTTL with ErrPendingTTLBufferFull and never reach
+	// pendingTTL, so they don't contribute to orphanTTLCount. If
+	// the caller swallows the HandleTTL error and continues, the
+	// previous condition (`orphanTTLCount > 0` only) would suppress
+	// the visibility signal for dropped expirations. Treat overflow
+	// as an independent reason to warn.
+	if r.warn != nil && (r.orphanTTLCount > 0 || r.pendingTTLOverflow > 0) {
 		fields := []any{
 			"count", r.orphanTTLCount,
 			"hint", "TTL records whose user key never appeared in a typed record — possible store corruption or an unknown type prefix",
--- a/internal/backup/redis_string.go
+++ b/internal/backup/redis_string.go
@@ -73,22 +73,27 @@ var ErrRedisInvalidStringValue = cockroachdberr.New("backup: invalid !redis|str|
 var ErrRedisInvalidTTLValue = cockroachdberr.New("backup: invalid !redis|ttl| value")
 
 // ErrPendingTTLBufferFull is returned by HandleTTL when an
-// unknown-kind TTL arrives but the pendingTTL buffer is already at
-// pendingTTLCap. The encoder fails closed here rather than silently
-// counting the TTL as an orphan because in real Pebble snapshot
-// order (`!redis|ttl|` lex-sorts before `!st|`/`!stream|`/`!zs|`),
-// the dropped entry would likely belong to a valid wide-column key
-// that arrives later — losing its expire_at_ms would produce a
-// restored database with non-expiring data that the source
-// snapshot's clients expected to expire. Codex P1 on PR #790
-// round 5.
+// unknown-kind TTL arrives but the pendingTTL buffer's BYTE
+// budget (pendingTTLBytesCap) is already exhausted. The encoder
+// fails closed here rather than silently counting the TTL as an
+// orphan because in real Pebble snapshot order (`!redis|ttl|`
+// lex-sorts before `!st|`/`!stream|`/`!zs|`), the dropped entry
+// would likely belong to a valid wide-column key that arrives
+// later — losing its expire_at_ms would produce a restored
+// database with non-expiring data that the source snapshot's
+// clients expected to expire.
 //
-// Recovery: raise WithPendingTTLCap above the snapshot's count of
-// unmatched-at-intake TTLs, or set the cap to 0 to explicitly opt
-// into the lossy counter-only mode (callers that prefer not to
-// see this error must accept that orphan-counted entries will be
-// dropped without inlining into wide-column state).
-var ErrPendingTTLBufferFull = cockroachdberr.New("backup: pendingTTL buffer at cap; raise WithPendingTTLCap or accept orphan-counter mode via WithPendingTTLCap(0)")
+// The budget is in BYTES (not entries) because Redis user keys can
+// be up to 1 MiB each; an entry-count cap of N still permits N
+// MiB of accumulated key bytes, which defeats the OOM protection
+// on adversarial snapshots with large keys. Codex P1 on PR #790
+// round 5 (entry-count -> byte-budget on round 6).
+//
+// Recovery: raise WithPendingTTLByteCap above the snapshot's
+// expected cumulative byte cost of unmatched-at-intake TTLs, or
+// set the byte cap to 0 to explicitly opt into the lossy
+// counter-only mode.
+var ErrPendingTTLBufferFull = cockroachdberr.New("backup: pendingTTL byte budget exhausted; raise WithPendingTTLByteCap or accept orphan-counter mode via WithPendingTTLByteCap(0)")
 
 // redisKeyKind tracks which Redis-type prefix introduced a user key, so that
 // when a later !redis|ttl|<K> record arrives we know whether to write its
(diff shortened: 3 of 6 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

Round-7 (commit `7f28222e`): codex P2 — short-circuit duplicate-key parkUnknownTTL.

## codex P2 — byte double-count on duplicate userKey

When `parkUnknownTTL` received the same userKey twice before claim, the map write overwrote (latest-wins) but `pendingTTLBytes` was incremented on every call. At byte-budget boundary, a duplicate would spuriously fire `ErrPendingTTLBufferFull`.

Fix: hoist `string(userKey)` to the top, short-circuit on already-buffered key (overwrite + return nil, no byte change). `cost` is only charged when a NEW key is added.

## Caller audit

`grep -rn 'parkUnknownTTL\b' --include='*.go'` → exactly one caller (HandleTTL's redisKindUnknown branch, redis_string.go:442). Error semantics for callers unchanged; the change strictly broadens the success path.

## Test added

`TestRedisDB_PendingTTLDuplicateKeyDoesNotDoubleCountBytes` pins:
- Replaying same key: bytes stays at 10, map size stays at 1, value updated (latest-wins).
- At exact cap (20/20 bytes): duplicate succeeds, overflow stays at 0.
- Extracted `pendingTTLState`/`capturePendingTTLState`/`assertPendingTTLState` helpers to keep the test under cyclop=10.

`go test -race ./internal/backup/` passes; `golangci-lint` clean.

bootjp:
@claude review

Round-8 (commit `f8ce906e`): codex P1 — raise default pendingTTL cap from 64 MiB to 1 GiB.

## codex P1 — large-but-valid snapshots hard-fail at default cap

Because `!redis|ttl|` lex-sorts BEFORE `!st|`/`!stream|`/`!zs|`, ALL expiring wide-column TTLs are parked in pendingTTL before any wide-column row arrives to drain them. Peak buffer size = sum((userKeyLen + 8) for all expiring wide-column keys). A real Redis with 20M expiring wide-column keys × 50-byte user keys peaks at ~1.16 GiB. The old 64 MiB default would hard-fail with `ErrPendingTTLBufferFull` on healthy snapshots — and there's no in-repo caller raising the ceiling (`WithPendingTTLByteCap` is test-only), so the default IS the policy.

## Fix

Raise `defaultPendingTTLBytesCap` to 1 GiB (`1 << 30`). Sizing rationale documented inline:
- Tolerates a 20M-expiring-wide-column-key workload at 50-byte average key size.
- Still a hard OOM ceiling — adversarial snapshots with TB of 1 MiB keys still fail closed.
- Typical dump-host budget headroom is 16+ GiB, so 1 GiB is a small fraction.

## Caller audit

Constant value change only — no semantics shift on `parkUnknownTTL` or any caller. Same fail-closed-at-cap behavior; just the threshold moved.

## Test added

`TestRedisDB_DefaultPendingTTLByteCapSurvives1GiBLoad` pins the constant invariant and verifies NewRedisDB applies it. Sentinel guard against accidental regression.

`go test -race ./internal/backup/` passes; `golangci-lint` clean.

bootjp:
@claude review

Round-9 (commit `060ce8a9`): codex P2 — preserve legacy zset members on meta-only.

## codex P2 — data-loss on meta-only snapshots

`HandleZSetMeta` was calling `markZSetWide(st)` which dropped any legacy `!redis|zset|<k>` members as soon as a `!zs|meta|<k>` row appeared, even without any `!zs|mem|` rows. But the live read (`adapter/redis_compat_helpers.go:610-621`) switches to wide-column ONLY when the `!zs|mem|` scan finds at least one row — a meta-only snapshot continues to serve the legacy blob to live clients. The encoder dropping legacy members on meta-alone produced empty/stale zsets on restore (actual user-visible data loss).

Although codex labelled this P2, the substance is data-loss for legitimate meta-without-members snapshot shapes.

## Fix

Remove `r.markZSetWide(st)` from `HandleZSetMeta`. Only `HandleZSetMember` triggers supersedence — matching the live read's "switch only on member-row presence" semantic.

## Caller audit

`grep -rn 'markZSetWide\|HandleZSetMeta\b' --include='*.go'` confirms markZSetWide has exactly one remaining caller (HandleZSetMember, redis_zset.go:198). Tests combining meta + member still trigger supersedence via the member row.

## Test changes

- `TestRedisDB_ZSetMetaAloneEvictsLegacyBlob` (asserted the OLD buggy behavior) → renamed and inverted to `TestRedisDB_ZSetMetaAlonePreservesLegacyBlob`. Seeds 2-member legacy blob, emits meta-only, asserts both members survive.
- `TestRedisDB_ZSetWideColumnSupersedesLegacyBlob` (member row triggers supersedence) and `TestRedisDB_ZSetLegacyBlobAfterWideRowsIsDropped` still pass.

`go test -race ./internal/backup/` passes; `golangci-lint` clean.

bootjp:
@claude review

Round-11 (commit `6f1f0716`): claude r9/r10 carryover items addressed.

## 3 carryover items

### Medium — `HandleZSetLegacyBlob` docstring scan-order claim (carried R3-R10)

Wrong claim: "`!redis|zset|` sorts BEFORE `!redis|ttl|`". Actual: 'z'=0x7A > 't'=0x74 on the diverging byte after the shared `!redis|` prefix, so **`!redis|ttl|` sorts BEFORE `!redis|zset|`** in natural Pebble order. The code was always correct (uses `claimPendingTTL` to drain the pre-arrived TTL); only the doc misled. Same wrong claim in `TestRedisDB_ZSetLegacyBlobWithInlineTTL` docstring — both corrected.

### Medium — sizing claim says 20M but cap fits only ~18.5M (carried R9)

Both `defaultPendingTTLBytesCap` and `ErrPendingTTLBufferFull` docstrings said 1 GiB cap accommodates a 20M-key workload "peaking at ~1.16 GiB". But 1.16 GiB > 1 GiB → operators trusting the guidance would hit `ErrPendingTTLBufferFull`. Corrected: cap holds **~18.5M expiring wide-column keys** at 50-byte avg (1,073,741,824 / 58 ≈ 18,513,652). Constant value unchanged.

### Low — missing test for natural Pebble order on legacy blobs (carried R3-R10)

Added `TestRedisDB_ZSetLegacyBlobTTLArrivesBeforeBlob`: TTL first → parked in pendingTTL → `HandleZSetLegacyBlob` → `zsetState()` drains via `claimPendingTTL`. Pins both the `expire_at_ms` round-trip and member survival. `TestRedisDB_ZSetLegacyBlobWithInlineTTL` remains for the reverse-order fast-path (now correctly documented as non-natural).

## Caller audit

Doc-only changes for (1) and (2). Test addition for (3) is purely additive. No semantics change anywhere; no caller-side audit needed.

`go test -race ./internal/backup/` passes; `golangci-lint` clean.

bootjp:
@claude review

Round-12 (commit `13039134`): claude r2-r10 base64-envelope assertion now actually checks the payload.

## Low — non-functional base64 assertion (carried R2-R10)

Previous check at `TestRedisDB_ZSetBinaryMemberUsesBase64Envelope`:

```go
if envelope["base64"] == "" { ... }
```

`envelope["base64"]` is `any`. Comparing `any` to `""` with `==` is never equal for any non-nil string — so any payload (including a corrupt one) passed. The test name claimed it verifies the round-trip; it didn't.

## Fix

Type-assert to string, base64-decode, bytes.Equal against the original input `[0x80, 0xff, 0x01]`. Extracted `assertBase64EnvelopeDecodesTo` helper to keep the test under cyclop=10.

`go test -race ./internal/backup/` passes; `golangci-lint` clean.

bootjp:
@claude review

Round-13 (commit `ab7ee2ce`): claude nits — consolidated gosec nolints + renamed contradictory test.

## Nits per CLAUDE.md "Avoid //nolint — refactor instead"

**(1) Consolidated 5 `//nolint:gosec` -> 1.** All five sites did the same `uint32(len(userKey))` cast for the wideColKeyLen prefix. Extracted `appendUserKeyLenBE32(out []byte, userKey string) []byte` helper with an explicit bounds check (panics on theoretical overflow) and a single nolint annotation. Same fix for `encodeZSetMetaValue` (added non-negative bounds check on `memberCount int64 -> uint64`).

**(2) Renamed `TestRedisDB_ZSetLegacyBlobRejectsMalformedKey` -> `...RejectsMissingPrefix`.** Previous docstring contradicted itself ("empty user key is valid" then "use a key without prefix to trigger failure"). Rewrote to accurately describe what the test pins: missing-prefix routing through ErrRedisInvalidZSetLegacyBlob.

`go test -race ./internal/backup/` passes; `golangci-lint` clean.

bootjp:
@claude review

Round-14 (commit `c9d49e28`): codex P2 — Finalize warns on overflow-only.

## codex P2 — overflow signal suppressed when orphanTTLCount==0

`Finalize`'s warn sink fired only when `orphanTTLCount > 0`. But `pendingTTLOverflow` entries fail-closed at `parkUnknownTTL` and never reach `pendingTTL`, so they don't contribute to `orphanTTLCount`. A caller that swallowed `ErrPendingTTLBufferFull` would lose the visibility signal for dropped expirations.

## Fix

Predicate widened: `orphanTTLCount > 0 || pendingTTLOverflow > 0`. The warning's existing field set already carries `pending_ttl_buffer_overflow` and `pending_ttl_buffer_bytes_cap` when overflow > 0, so operators get the same diagnostics in the overflow-only path.

## Caller audit

`grep -rn 'orphanTTLCount\b|pendingTTLOverflow\b'` confirms single writer for each counter (`parkUnknownTTL`, `Finalize`) and a single reader (`Finalize` warn). Predicate-only change broadens emission; no caller depends on the absence of the warning.

## Test added

`TestRedisDB_FinalizeWarnsOnOverflowOnlyEvenWhenOrphanCountZero` pins the overflow-only path: byte_cap=16, two TTL inserts (second fails closed), drain via HandleSetMember, then Finalize. Asserts `redis_orphan_ttl` fires AND `pending_ttl_buffer_overflow` field present. Extracted `warnFieldsContain` helper to stay under cyclop=10.

`go test -race ./internal/backup/` passes; `golangci-lint` clean.

PR #793, opened 2026-05-20
feat(encryption): Stage 6D-3 — CapabilityFanout helper (Voters ∪ Learners)

bootjp:
## Summary

- Adds `internal/admin/capability_fanout.go` — the §4.1 helper that the 6D-6 cutover RPC will call to verify every (voter ∪ learner) of every Raft group is reachable and reports `encryption_capable=true` before the cutover entry is proposed.
- Dedupes by `FullNodeID` per §4.1 (a node serving multiple groups is probed exactly once), fans dials concurrently under one `context.WithTimeout` (default = one RaftElectionTimeout per §4.3), surfaces dial / RPC failure as `Reachable=false` verdicts so the operator-facing error can name the unreachable peer.
- `Result.OK = true` iff every verdict has `Reachable && EncryptionCapable` — no partial-success mode per §4.3.
- Adds the §9 table-driven unit tests (every-capable, one-not-capable, one-unreachable, duplicate-nodes-deduplicated) + three input-validation cases (nil `DialFunc`, non-positive timeout, empty snapshot). 7 tests total, all passing.
- One deliberate deviation: the result struct is `CapabilityFanoutResult` rather than the design doc's `FanoutResult` to avoid colliding with the unrelated `admin.FanoutResult` already in `keyviz_fanout.go` (same package).
- Design-doc lifecycle: renamed `2026_05_18_proposed_6d_enable_storage_envelope.md` → `2026_05_18_partial_6d_enable_storage_envelope.md` and updated the doc to record what shipped (6D-1, 6D-2, 6D-3) and what remains (6D-4, 6D-5, 6D-6).
- Operator-inert: no in-tree callers yet. 6D-6 wires it.

## Self-review (5 lenses)

1. **Data loss** — N/A; helper is read-only.
2. **Concurrency / distributed failures** — fan-out is goroutine-per-target with a single `context.WithTimeout` governing the deadline. Result aggregation guarded by `sync.Mutex`. Dedup phase runs before goroutines launch so the loop-var-capture race is avoided (`m` passed by value as the goroutine arg). No partial-success mode — a single unreachable learner is a hard "no" per §8.
3. **Performance** — concurrent dials bounded by `len(unique FullNodeID)`; no quadratic fan-out. Maps + result slice preallocated. `uint64ToDecimal` avoids `fmt` for the dedup-key hot path.
4. **Data consistency** — N/A; synchronous probe, no side effects. Per §4.2 deliberately does NOT cache — every cutover probes fresh.
5. **Test coverage** — §9 table-driven cases all pin. Dedup contract asserted by counting per-address dial calls (`stubDial.calls`). Input-validation cases pin the bad-input refusal posture.

## Test plan

- [x] `go test ./internal/admin/...` — all 7 `TestCapabilityFanout_*` tests pass
- [x] `golangci-lint run internal/admin/...` — 0 issues
- [ ] Reviewer approval (standalone helper; no callers wired yet, so no Jepsen suite applies)

## Summary by CodeRabbit

* **Documentation**
  * Updated Stage 6D milestone status with completed sub-milestones (6D-1 through 6D-3) and outlined remaining work items (6D-4 through 6D-6).

* **Infrastructure Updates**
  * Added concurrent capability verification helper to support cluster-wide node readiness assessment.

* **Tests**
  * Added comprehensive unit tests covering capability verification scenarios including success, failures, deduplication, and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/793?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/design/2026_05_18_partial_6d_enable_storage_envelope.md
+++ b/docs/design/2026_05_18_partial_6d_enable_storage_envelope.md
@@ -2,11 +2,35 @@
 
 | Field | Value |
 |---|---|
-| Status | proposed |
+| Status | partial — 6D-1 (doc), 6D-2 (startup guards), 6D-3 (capability fan-out helper) shipped; 6D-4, 6D-5, 6D-6 remain |
 | Date | 2026-05-18 |
 | Parent design | [`2026_04_29_partial_data_at_rest_encryption.md`](2026_04_29_partial_data_at_rest_encryption.md) |
 | Blockers (now satisfied) | 6B (KEK plumbing), 6C-1 / 6C-2 (startup guards), 6C-2d (`ErrSidecarBehindRaftLog` wiring) |
 | Bundles | 6C-3 (`ErrNodeIDCollision` + `ErrLocalEpochRollback` cluster-wide guards) |
+
+### Shipped milestones (lifecycle: partial)
+
+- **6D-1** (doc) — this file. Landed with the initial proposal commit.
+- **6D-2** (startup guards) — `ErrNodeIDCollision` primitive
+  (`internal/encryption/node_id_collision.go`), `ErrLocalEpochRollback`
+  primitive (`internal/encryption/local_epoch_rollback.go`), and the
+  `probe-node-id` admin CLI subcommand. Landed in PR #788.
+- **6D-3** (capability fan-out helper) — `CapabilityFanout` in
+  `internal/admin/capability_fanout.go` with the table-driven §9
+  unit tests. The §4.1 `FanoutResult` type was renamed
+  `CapabilityFanoutResult` to disambiguate from the unrelated
+  `admin.FanoutResult` already shipped in `keyviz_fanout.go`; the
+  contract is otherwise unchanged.
+
+### Open milestones
+
+- **6D-4** — `RotateSubEnableStorageEnvelope = 0x04` wire-format
+  addition + FSM apply-path dispatch + `StorageEnvelopeCutoverIndex`
+  sidecar field.
+- **6D-5** — §6.2 storage-layer toggle (`PutAt` consults
+  `StorageEnvelopeActive`).
+- **6D-6** — `EnableStorageEnvelope` admin RPC + CLI command +
+  integration test composing 6D-3 + 6D-4 + 6D-5.
 
 ## 0. Why this doc exists
 
--- a/docs/design/2026_05_18_partial_6d_enable_storage_envelope.md
+++ b/docs/design/2026_05_18_partial_6d_enable_storage_envelope.md
@@ -8,7 +8,7 @@
 | Blockers (now satisfied) | 6B (KEK plumbing), 6C-1 / 6C-2 (startup guards), 6C-2d (`ErrSidecarBehindRaftLog` wiring) |
 | Bundles | 6C-3 (`ErrNodeIDCollision` + `ErrLocalEpochRollback` cluster-wide guards) |
 
-### Shipped milestones (lifecycle: partial)
+## Shipped milestones (lifecycle: partial)
 
 - **6D-1** (doc) — this file. Landed with the initial proposal commit.
 - **6D-2** (startup guards) — `ErrNodeIDCollision` primitive
@@ -22,7 +22,7 @@
   `admin.FanoutResult` already shipped in `keyviz_fanout.go`; the
   contract is otherwise unchanged.
 
-### Open milestones
+## Open milestones
 
 - **6D-4** — `RotateSubEnableStorageEnvelope = 0x04` wire-format
   addition + FSM apply-path dispatch + `StorageEnvelopeCutoverIndex`
--- a/internal/admin/capability_fanout.go
+++ b/internal/admin/capability_fanout.go
@@ -1,0 +1,241 @@
+package admin
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	pb "github.com/bootjp/elastickv/proto"
+	pkgerrors "github.com/cockroachdb/errors"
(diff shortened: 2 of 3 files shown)


bootjp:
@claude review

Round-1 (commit `b46d1f5e`) addressed three reviewer findings from the initial round:

## codex P1 / gemini high / coderabbit major — fail-closed on empty address

Pre-fix: `addCapabilityFanoutDedupTarget` returned early when `member.Address == ""`, silently dropping malformed catalog rows from `Result.Verdicts`. The cutover RPC could then see `OK=true` against an unprobed peer.

Fix: admit every row into the dedup map; `probeCapability` classifies empty-Address members as `Reachable=false` with a new `errCapabilityFanoutMalformedMember` sentinel. The caller-facing error can now name the misconfigured member.

Sibling case (FullNodeID=0 AND Address=""): synthetic ordinal-based dedup key so two malformed rows surface as two separate verdicts rather than collapsing into one and hiding the second misconfiguration.

Regression tests added per CLAUDE.md test-then-fix convention:
- `TestCapabilityFanout_EmptyAddressFailsClosed`
- `TestCapabilityFanout_FullyMalformedRowsNotCollapsed`

Both confirmed failing pre-fix, passing post-fix.

Caller audit: `addCapabilityFanoutDedupTarget` and `probeCapability` are file-private; sole callers are also in this file; no external propagation. The public-facing semantic change ("Result.Verdicts now includes malformed-row entries as Reachable=false") is strictly safer than the pre-fix silent drop, and `capabilityFanoutAllOK` already requires `Reachable=true`.

## coderabbit minor — MD001 heading-level skip

Promoted `### Shipped milestones` and `### Open milestones` to H2 in the partial design doc.

## gemini medium — preallocate dedup map

`buildCapabilityFanoutDedupSet` now pre-sizes the dedup map with the upper-bound member count.

## Verification

- `go test -run TestCapabilityFanout ./internal/admin/...` — all 9 tests pass (7 pre-existing + 2 new regressions)
- `golangci-lint run internal/admin/...` — 0 issues

bootjp:
@claude review

Round-2 (commit `f2bbdaa3`): addressed codex r2 P1.

## codex r2 P1 — nil-client guard against buggy DialFunc

Pre-fix: if a `DialFunc` returns `(nil, nil, nil)` — no error but also no client — `probeCapability` would panic on `client.GetCapability(...)` and crash the admin RPC path. The fix is a defensive nil-client check that converts the contract violation into a `Reachable=false` verdict with a new `errCapabilityFanoutBadDialer` sentinel.

Regression test per CLAUDE.md test-then-fix:
- `TestCapabilityFanout_NilClientFromDialFailsClosed` — injects a buggy dialer that returns `(nil, nil, nil)`. Confirmed panicking on the pre-fix code (goroutine traceback through `probeCapability:264`); passes on the fix.

Caller audit: `probeCapability` is file-private, sole caller `runCapabilityFanoutProbes` is also in this file. No external propagation. The semantic change is "nil client from a buggy dialer now produces a `Reachable=false` verdict instead of a panic" — strictly safer than the pre-fix crash.

## Verification

- `go test -run TestCapabilityFanout ./internal/admin/...` — all 10 tests pass (8 pre-existing + 1 new regression + 1 from round-1)
- `golangci-lint run internal/admin/...` — 0 issues

bootjp:
@claude review

Round-3 (commit `438482f3`): addressed two codex r3 findings.

## codex r3 P1 — reject capability replies from unexpected node IDs

Pre-fix: `probeCapability` accepted any successful `GetCapability` response, even when the responder's `full_node_id` differed from the snapshot's expected one (stale routing, shared address). One healthy node could satisfy probes for another, and `Result.OK` could become true despite the intended member never being verified.

Fix: when both expected and response `full_node_id`s are non-zero and disagree, the verdict is `Reachable=false` with a new `errCapabilityFanoutMismatchedResponder` sentinel.

Regression test: `TestCapabilityFanout_MismatchedResponderIDFailsClosed`.

## codex r3 P2 — honor timeout even when a probe ignores ctx

Pre-fix: `runCapabilityFanoutProbes` did `wg.Wait()` unconditionally. A buggy DialFunc/client that ignored ctx cancellation would hang the helper indefinitely, violating the §4.3 "Returns within `timeout`" contract.

Fix: pre-seed every dedup-key slot with a `Reachable=false` verdict carrying `errCapabilityFanoutProbeTimeout`. Each goroutine overwrites its slot when its probe finishes. The orchestrator selects on `min(done, ctx.Done())` and returns whatever slot values are present.

Goroutine cleanup: goroutines that ignore ctx continue running. The helper documents this — the contract is "the HELPER returns within timeout", not "every spawned goroutine returns within timeout". A dialer that ignores ctx is a caller-side bug; the helper's job is to prevent that bug from hanging the admin RPC path.

Regression test: `TestCapabilityFanout_TimeoutBoundsHelperReturn` — the pre-fix code hung the entire 30s test process; the fix returns within 500ms (5x budget for CI margin).

## Caller audit

`probeCapability`, `runCapabilityFanoutProbes`, and the two new sentinels are all file-private. The public-facing semantic changes are: (1) stale-routing scenarios previously slipping through as `OK=true` now produce `OK=false`, and (2) hung probes that hung the helper now produce `Reachable=false` verdicts. Both strictly safer. `capabilityFanoutAllOK` already requires `Reachable=true` so no caller-side change is needed.

## Verification

- `go test -race -run TestCapabilityFanout -timeout 60s ./internal/admin/...` — all 12 tests pass (10 pre-existing + 2 new regressions) under `-race`
- `golangci-lint run internal/admin/...` — 0 issues

PR #794, opened 2026-05-20
adapter/sqs: AdminPeekQueue backend (Phase 3)

bootjp:
## Summary

Phase 3 of `docs/design/2026_05_16_proposed_admin_purge_queue.md`: backend `AdminPeekQueue` method on `*adapter.SQSServer`.

- Returns up to `opts.Limit` currently-visible messages (`visible_at <= now`); pure read, no receive-count bump, no receipt-handle minting, no visibility-timer change.
- Partitioned FIFO support via rotated sequential scanning across partitions, reusing the receive path's `receiveFanoutCounters` for fairness.
- Versioned JSON + base64url cursor pins `Generation`, `StartPartition`, `Partition`, and `LastKey`; stale-generation cursors are rejected so a purge between pages forces the SPA to refresh from the front.
- `AdminRole.canRead()` — peek requires non-zero role; read-only principals can triage, purge stays on `canWrite()`.
- Pre-existing `goconst` regression in `sqs_messages.go` (the validator switch's inline `"String"` / `"Number"` literals) fixed by extracting `sqsAttributeBaseTypeString` / `sqsAttributeBaseTypeNumber` alongside the existing `sqsAttributeBaseTypeBinary`.

**Caller audit (semantic-change rule):** no existing function signatures changed. `canRead` is a new method; `AdminPeekQueue` is a new entrypoint. The validator switch's constant extraction is a literal-for-named rename (byte-identical wire behaviour).

## Behavior change

- New `AdminPeekQueue` admin method. Not yet wired to an HTTP handler (Phase 4) — this PR only adds the adapter surface.
- No change to SigV4 paths, `AdminPurgeQueue`, or `AdminDescribeQueue`.

## Risk

Low. Pure-read entry; the only writes I touch are the constant rename in `sqs_messages.go` (byte-identical to the prior inline literals). The cursor codec is hard-capped at 512 bytes after encoding; oversize cursors surface as `ErrAdminSQSValidation` from the decoder.

## Self-review (5 passes)

1. **Data loss** — Read-only entry. The constant extraction in `sqs_messages.go` is a literal-for-named rename with no behaviour change.
2. **Concurrency** — Leader-only via `isVerifiedSQSLeader`. Single `nextTxnReadTS` threaded into `loadQueueMetaAt`, `ScanAt`, and per-row `GetAt`, so the page is a consistent MVCC snapshot. The cursor's stored Generation is checked against the live meta on every continuation page — a purge between pages produces `ErrAdminSQSValidation` rather than rows from a purged generation.
3. **Performance** — `O(opts.Limit)` `ScanAt` round-trips + per-row `GetAt`. Hard caps: `Limit <= 100`, body <= 256 KiB. Cursor is bounded at 512 bytes. No hot-path allocations.
4. **Consistency** — Cursor encodes `Generation`; mismatch surfaces as `ErrAdminSQSValidation`. Partition rotation uses the same `receiveFanoutCounters` the receive path uses, so peek's partition fairness matches the data plane's. Data records are read via `sqsMsgDataKeyDispatch`, so partitioned FIFO data keys are routed to the same partition the vis-index entry was found under.
5. **Test coverage** — 22 new tests (happy path / no-bump-on-peek / truncation / Limit clamp / cursor round-trip / stale-gen / four malformed-cursor classes / read-only allowed / role-less denied / missing queue / empty name / delayed-message hidden / FIFO attribute projection / typed MessageAttribute (String+Binary) round-trip / partitioned FIFO pagination / cursor codec round-trip+empty+oversize / preparePeekCursor fresh+stale / clampPeekLimit+clampPeekBodyBytes truth tables / projectPeekedAttributes nil-safety / canRead truth table). `go test -race` clean.

## Test plan

- [x] `go test -race -count=1 ./adapter/...` (targeted: `TestAdminPeekQueue*`, `TestPeekCursorCodec*`, `TestClampPeek*`, `TestPreparePeekCursor*`, `TestEncodePeekCursor*`, `TestProjectPeekedAttributes*`, `TestAdminRole_CanRead`, `TestAdminPurge*`, `TestAdminQueueSummary*`, `TestSQSServer_PartitionedFIFO*`, `TestSQSServer_Send*`)
- [x] `make lint` clean on `./adapter/...` (0 issues)
- [ ] CI on this PR

## Out of scope (follow-ups)

- **Throttle integration**: dedicated per-queue admin-peek bucket (`bucketActionAdminPeek` + `resolveActionConfig` explicit case + `*adminPeekThrottledError` + `*PeekThrottledError`). Leader-only + `Limit <= 100` already bound the steady-state cost; the dedicated bucket adds a per-queue cap that should land alongside the SPA wiring so the metric has a real consumer.
- **Phase 4**: HTTP handler at `GET /admin/api/v1/sqs/queues/{name}/messages`, bridge in `main_admin.go`, integration tests.
- **`principalForReadSensitive` live `RoleStore` re-check** (design doc Goal 8): blocked on the wider `RoleStore` plumbing — neither `AdminPurgeQueue` nor `AdminDeleteQueue` do a live re-check today; peek will inherit that pattern when the broader work lands.

## Summary by CodeRabbit

* **New Features**
  * Admin API to peek SQS queue messages without modifying them: pagination with continuation tokens, configurable body truncation while reporting original size, full message attribute visibility, supports standard and FIFO queues.
* **Bug Fix / Permissions**
  * Read-only admin roles are now allowed to perform non-destructive admin reads (peek).
* **Polish**
  * Improved message-attribute typing/handling for more reliable attribute projection.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/794?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/adapter/dynamodb_admin.go
+++ b/adapter/dynamodb_admin.go
@@ -83,6 +83,16 @@ const (
 // future "delete-only" tier reads consistently across the package.
 func (r AdminRole) canWrite() bool { return r == AdminRoleFull }
 
+// canRead reports whether the role authorises non-destructive but
+// sensitive reads — e.g. SQS AdminPeekQueue, which exposes message
+// bodies / attributes. Both AdminRoleReadOnly and AdminRoleFull
+// satisfy this gate; the zero value (unauthenticated / role-less
+// principal) does not. List / Describe paths use the looser
+// session-auth gate because their output is queue metadata already
+// shown on the SPA list page; peek is divergent because the payload
+// is the message bodies themselves.
+func (r AdminRole) canRead() bool { return r == AdminRoleReadOnly || r == AdminRoleFull }
+
 // AdminPrincipal is the authentication context every admin write
 // entrypoint takes. The adapter re-evaluates authorisation against
 // this principal *itself* — it does not trust the caller to have
--- a/adapter/sqs_admin_peek.go
+++ b/adapter/sqs_admin_peek.go
@@ -233,7 +233,7 @@ func (s *SQSServer) AdminPeekQueue(
 	if !exists {
 		return nil, "", ErrAdminSQSNotFound
 	}
-	cursor, err = preparePeekCursor(cursor, meta, s.nextReceiveFanoutStart(name, meta.PartitionCount))
+	cursor, err = preparePeekCursor(cursor, meta, s.peekStartPartition(name, cursor, meta))
 	if err != nil {
 		return nil, "", err
 	}
@@ -246,6 +246,26 @@ func (s *SQSServer) AdminPeekQueue(
 		return nil, "", err
 	}
 	return rows, encoded, nil
+}
+
+// peekStartPartition returns the starting partition for a fresh peek
+// walk, or 0 when a starting partition is not needed (continuation
+// pages already carry one in the cursor; non-partitioned queues only
+// have partition 0).
+//
+// The fanout counter is consulted ONLY on the first page of a peek
+// walk. nextReceiveFanoutStart increments the per-queue counter on
+// every call regardless of whether the returned value is consumed;
+// advancing it on every continuation page would perturb the same
+// counter ReceiveMessage reads for partition fairness, reintroducing
+// fixed-stride aliasing (Codex r2 P1: e.g. PartitionCount=4 with 3
+// peek pages between receives lands every receive on the same
+// partition, starving the other three).
+func (s *SQSServer) peekStartPartition(queueName string, cursor *peekCursor, meta *sqsQueueMeta) uint32 {
+	if cursor != nil || meta.PartitionCount <= 1 {
+		return 0
+	}
+	return s.nextReceiveFanoutStart(queueName, meta.PartitionCount)
 }
 
 // preparePeekCursor builds the effective cursor for this call. On the
--- a/adapter/sqs_admin_peek.go
+++ b/adapter/sqs_admin_peek.go
@@ -1,0 +1,469 @@
+package adapter
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"strings"
+	"time"
+
+	"github.com/bootjp/elastickv/store"
+	"github.com/cockroachdb/errors"
+	json "github.com/goccy/go-json"
+)
+
+// AdminPeekedAttribute mirrors the typed shape SQS uses for
+// MessageAttribute values — DataType (e.g. "String", "Number",
+// "Binary", "String.MyCustom") plus the value in the appropriate
+// representation. The earlier draft used map[string]string here,
+// which would have flattened the typed attribute set stored in
+// sqsMessageRecord.MessageAttributes and silently dropped binary
+// payloads + the DataType discriminator (Codex r11 on the design
+// doc). Operators triaging a DLQ need both — a message routed there
+// because of an attribute-encoding mismatch is invisible if peek
+// only surfaces stringified values.
+type AdminPeekedAttribute struct {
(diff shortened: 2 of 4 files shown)


bootjp:
@claude review

bootjp:
@claude review

Addressed Codex r1 P1 in 9bfaacfa:
- preparePeekCursor now rejects out-of-range StartPartition / Partition with ErrAdminSQSValidation. Non-partitioned queues (PartitionCount <= 1) require both fields to be 0.
- Closes the unbounded-loop DoS vector: walkPeek's '(Partition+1) mod PartitionCount == StartPartition' termination never fired for an out-of-range StartPartition, looping ScanAt forever.
- Regression test runs under a 5-second deadline so any future loss of the bounds check terminates rather than hanging CI; four unit tests pin the truth table.
- Caller audit: walkPeek's partition advance ((p+1) mod PartitionCount) is the only other cursor producer; it yields values in [0, PartitionCount) by construction.

bootjp:
@claude review

Addressed Codex r2 P1 in 3e6841e0:
- AdminPeekQueue no longer calls nextReceiveFanoutStart on continuation pages. New peekStartPartition helper consults the counter only when (a) cursor is nil (first page) AND (b) PartitionCount > 1.
- Closes the fanout-stride aliasing: peek pagination no longer perturbs the partition-rotation counter ReceiveMessage uses for fairness.
- Regression test reads receiveFanoutCounters directly after the first page, drives 5 continuation calls, and asserts the counter is unchanged.
- Caller audit: nextReceiveFanoutStart still bumps once per receive (unchanged) and once per fresh peek walk (new), never on continuations.

bootjp:
@claude review

Addressed r3 in fc35ee51:
- Codex P2 (sqs_admin_peek.go:268): peekStartPartition / walkPeek / preparePeekCursor now key off effectivePartitionCount(meta), aligning peek with the data plane. perQueue FIFO queues now walk only partition 0 instead of N-1 guaranteed-empty partitions. New end-to-end test TestAdminPeekQueue_PerQueueFIFOCollapsesToOnePartition + cursor codec test TestPreparePeekCursor_PerQueueCollapse pin the contract.
- CodeRabbit (test order-coupling): HappyPath now uses set-based assertion via assertPeekRowsAsSet — vis-index entries with the same visible_at millisecond tie-break on random message_id, so order-sensitive assertion was timing-flaky.
- CodeRabbit (counter type assertion panic): extracted loadFanoutCounter helper that fails the test on bad type / nil instead of panicking on counter.Load().
- Caller audit: effectivePartitionCount is the same helper receiveMessage uses for its fanout loop, so peek and receive agree on partition usage.

bootjp:
@claude review

Addressed r4 in ddb4e6ec:
- Codex P2 (sqs_admin_peek.go:401, LastKey forgery → read amplification): preparePeekCursor now validates cursor.LastKey has the expected visibility-index prefix (queue+partition+gen). Three forgery classes covered: arbitrary bytes / foreign-queue LastKey / mismatched-partition LastKey. Caller audit: signature gained queueName; one non-test caller (AdminPeekQueue) and 10 test sites updated.
- CodeRabbit Major (AdminPeekedMessage JSON tags missing): added snake_case json tags matching the design doc §3.5 wire shape; omitempty on GroupID / DeduplicationID / Attributes so empty values are dropped, not emitted as null/'""'. Two new tests pin the wire shape: TestAdminPeekedMessage_JSONWireFormat + TestAdminPeekedMessage_JSONOmitsEmptyAttributes.

PR #804, opened 2026-05-22
feat(encryption): Stage 6D-4 - cutover wire byte + apply dispatch + sidecar field

bootjp:
## Summary

- Stage 6D-4 lands the FSM-level half of the storage-envelope cutover per `docs/design/2026_05_18_partial_6d_enable_storage_envelope.md` §2.1 / §2.2 / §6.4. Operator-inert: no CLI, no §6.2 PutAt toggle — those land in 6D-5 / 6D-6.
- Promote the `0x04` reservation to a declared constant `RotateSubEnableStorageEnvelope`, widen `readRotationSubTag` to a whitelist switch (the §2.2 "two-site whitelist change required" property), and dispatch the new sub-tag in `ApplyRotation`.
- Add `Sidecar.StorageEnvelopeCutoverIndex uint64` field (§6.4). The fresh cutover apply flips `StorageEnvelopeActive`, records the cutover index, and advances `RaftAppliedIndex` inside one `WriteSidecar` fsync. Stale-DEKID (§2.1 #3 race against `RotateDEK`) and already-active duplicates are benign no-ops that consume the entry without halting — the §6.4 stable idempotency token is preserved across duplicates.

## Test plan

- [x] `go test -race -timeout 180s ./internal/encryption/... ./kv/... ./store/...` — all green
- [x] `golangci-lint --config=.golangci.yaml run` (new-from-rev origin/main) — 0 issues
- [x] `TestRotation_EnableStorageEnvelope_RoundTrip` — wire shape pins
- [x] `TestRotation_SubTagConstantPinned` — guards against silent byte drift (would break rolling-upgrade compatibility)
- [x] `TestApplyRotation_EnableStorageEnvelope_FreshSuccess` — bootstrap → cutover → assert sidecar flip, cutover index, `RaftAppliedIndex` advance, `Active.Storage` unchanged, proposer registration row inserted; parametrised over `Wrapped = nil` AND `Wrapped = []byte{}` to pin the length-based guard
- [x] `TestApplyRotation_EnableStorageEnvelope_RejectsNonStoragePurpose` — §2.1 #1 halt
- [x] `TestApplyRotation_EnableStorageEnvelope_RejectsNonEmptyWrapped` — §2.1 #2 halt
- [x] `TestApplyRotation_EnableStorageEnvelope_RejectsProposerDEKMismatch` — proposer DEK pinning halt
- [x] `TestApplyRotation_EnableStorageEnvelope_StaleDEKID_BenignNoOp` — `RotateDEK` races; cutover apply consumes the entry without halting
- [x] `TestApplyRotation_EnableStorageEnvelope_AlreadyActive_IdempotentNoOp` — duplicate cutover at a higher raftIdx; original `StorageEnvelopeCutoverIndex` preserved

## Self-review (5 lenses)

1. **Data loss** — no. The cutover apply touches only sidecar fields; all writes go through the existing `WriteSidecar` crash-durable protocol. Stale-DEKID and already-active no-ops still `WriteSidecar` so the `RaftAppliedIndex` advance is durable; without that, a crash + replay would re-apply the entry.
2. **Concurrency / distributed failures** — the §2.1 #3 race posture is the central concurrency invariant. The fresh-success and benign-no-op paths are exhaustive over the (`Active.Storage`, `StorageEnvelopeActive`) sidecar state at apply time. Apply is single-threaded per FSM so no internal locking is needed.
3. **Performance** — one extra `ReadSidecar` + `WriteSidecar` per cutover apply; cutover is one-shot per cluster so this is irrelevant to steady-state. Length-based check on `Wrapped` is O(1).
4. **Data consistency** — §6.4 atomicity holds (all three sidecar field writes inside one `WriteSidecar` fsync). The two-site whitelist invariant is pinned by a comment in `readRotationSubTag` and by `TestRotation_SubTagConstantPinned`. The §2.1 #4 idempotency contract holds (original `StorageEnvelopeCutoverIndex` preserved across duplicate applies, exercised by the regression test).
5. **Test coverage** — 6 new applier tests cover every §2.1 constraint + each apply-path outcome (fresh success / stale no-op / already-active no-op). 2 new wire tests pin the constant and the round-trip. `Wrapped = nil` vs `[]byte{}` is parametrised so the length-based check cannot regress to a `== nil` check.

Refs: design doc §2.1 / §2.2 / §6.4 (Stage 6D-4 row of §7 decomposition table).

### DIFF
--- a/docs/design/2026_05_18_partial_6d_enable_storage_envelope.md
+++ b/docs/design/2026_05_18_partial_6d_enable_storage_envelope.md
@@ -2,7 +2,7 @@
 
 | Field | Value |
 |---|---|
-| Status | partial — 6D-1 (doc), 6D-2 (startup guards), 6D-3 (capability fan-out helper) shipped; 6D-4, 6D-5, 6D-6 remain |
+| Status | partial — 6D-1 (doc), 6D-2 (startup guards), 6D-3 (capability fan-out helper), 6D-4 (cutover wire + apply dispatch) shipped; 6D-5, 6D-6 remain |
 | Date | 2026-05-18 |
 | Parent design | [`2026_04_29_partial_data_at_rest_encryption.md`](2026_04_29_partial_data_at_rest_encryption.md) |
 | Blockers (now satisfied) | 6B (KEK plumbing), 6C-1 / 6C-2 (startup guards), 6C-2d (`ErrSidecarBehindRaftLog` wiring) |
@@ -21,12 +21,19 @@
   `CapabilityFanoutResult` to disambiguate from the unrelated
   `admin.FanoutResult` already shipped in `keyviz_fanout.go`; the
   contract is otherwise unchanged.
+- **6D-4** (cutover wire + apply dispatch) —
+  `RotateSubEnableStorageEnvelope = 0x04` constant
+  (`internal/encryption/fsmwire/wire.go`), two-site whitelist
+  (`readRotationSubTag` decoder + `ApplyRotation` switch in
+  `internal/encryption/applier.go`), `StorageEnvelopeCutoverIndex`
+  sidecar field, and the apply-path semantics per §2.1 / §6.4:
+  malformed entries halt via `ErrEncryptionApply`, stale-DEKID and
+  already-active outcomes are benign no-ops that advance
+  `RaftAppliedIndex` without halting. Operator-inert until 6D-5
+  wires the §6.2 storage-layer toggle.
 
 ## Open milestones
 
-- **6D-4** — `RotateSubEnableStorageEnvelope = 0x04` wire-format
-  addition + FSM apply-path dispatch + `StorageEnvelopeCutoverIndex`
-  sidecar field.
 - **6D-5** — §6.2 storage-layer toggle (`PutAt` consults
   `StorageEnvelopeActive`).
 - **6D-6** — `EnableStorageEnvelope` admin RPC + CLI command +
--- a/internal/encryption/applier.go
+++ b/internal/encryption/applier.go
@@ -548,15 +548,39 @@ func advanceRaftAppliedIndex(sc *Sidecar, raftIdx uint64) {
 	}
 }
 
-// ApplyRotation implements §5.2 / §5.4 rotation apply. Stage 6B-1
-// handles only the RotateSubRotateDEK sub-tag (rotate the storage
-// or raft DEK to a new key_id). Other sub-tags (rewrap-deks,
-// retire-dek, enable-storage-envelope, enable-raft-envelope) land
-// in later stages and return ErrEncryptionApply for now so the
+// ApplyRotation implements §5.2 / §5.4 rotation apply. The sub-tag
+// dispatches the entry to the per-variant handler:
+//
+//   - RotateSubRotateDEK — install a new wrapped DEK and re-point
+//     the Active slot for its Purpose (Stage 6B-1).
+//   - RotateSubEnableStorageEnvelope — one-shot storage-layer
+//     cutover (Stage 6D-4). Flips StorageEnvelopeActive and records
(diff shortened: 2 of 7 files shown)


bootjp:
@claude review

Round-1 (commit `74a504c8`): addressed two gemini-code-assist medium findings.

## gemini medium #1 — `applyEnableStorageEnvelope` registration ordering

Pre-fix: `WriteSidecar` ran before `ApplyRegistration` in the fresh-success path. A crash between the two would leave `StorageEnvelopeActive=true` with the proposer's writer-registry row missing, and FSM replay would short-circuit on the §2.1 #4 already-active no-op — permanently skipping the registry insert.

Fix: reorder so `ApplyRegistration` runs first. It is idempotent on same-epoch via the existing §4.1 case 2-idempotent path, so a crash + replay re-enters the fresh-success branch and re-runs both operations cleanly. The §6.4 atomicity invariant still holds: all three sidecar field writes (`StorageEnvelopeActive`, `StorageEnvelopeCutoverIndex`, `RaftAppliedIndex`) land inside one `WriteSidecar` fsync.

Regression test: `TestApplyRotation_EnableStorageEnvelope_RegistrationBeforeSidecar` injects `setEr` on the registry store and asserts the sidecar's cutover state stays untouched (no premature flip) when the registration insert fails.

## gemini medium #2 — sidecar version bump 1 → 2

Pre-fix: adding `StorageEnvelopeCutoverIndex` while keeping `SidecarVersion = 1` would let a downgraded pre-6D-4 binary silently drop the field on its next `WriteSidecar`, breaking the §6.4 idempotency-token contract.

Fix: bump `SidecarVersion = 2`, add `MinReadableSidecarVersion = 1` so a freshly-upgraded post-6D-4 node still reads the legacy pre-6D-4 sidecar on first boot. `WriteSidecar` emits the current version, so the very next mutation auto-upgrades the file; a subsequent downgrade refuses to boot via the version check — preventing the silent drop.

Regression test: `TestReadSidecar_AcceptsLegacyVersion1` pins the legacy read + auto-upgrade-on-write path. `TestReadSidecar_RejectsUnknownVersion` updated to test `{0, 3, 42, -1}` (version 2 is now valid; version 1 covered by the legacy-read test).

## Caller audit

- `ApplyRegistration` is also called from `applyRotateDEK` and the bootstrap path; both keep their existing ordering. The reorder is local to `applyEnableStorageEnvelope`.
- The only non-test code reading `Sidecar.Version` is `sidecar.go` itself (ReadSidecar version check, WriteSidecar version override). No external callers compare the field.

## Verification

- `go test -race -timeout 180s ./internal/encryption/... ./kv/...` — all green
- `golangci-lint --new-from-rev=origin/main` — 0 issues

bootjp:
@claude review

Round 2 (commit `f6cad1f7`, doc-only):

**Status:**
- gemini medium #1 (registration-before-sidecar): resolved in r1 commit `74a504c8`
- gemini medium #2 (sidecar version bump): resolved in r1 commit `74a504c8`
- codex P1 (`Reapply proposer registration on idempotent cutover`): the original report was filed against commit `3eb4555c77`, before the registration-before-sidecar reordering. The crash window codex described ("crash after WriteSidecar but before registration insert") is **only reachable under the old ordering**; with the r1 reordering it is structurally impossible. Claude-bot r1 explicitly confirmed this resolution:

> *"The 'crash after WriteSidecar but before registration insert' scenario codex described was contingent on the old ordering; with registration-first it is impossible."*

**This commit:** applies the one cosmetic suggestion from claude-bot r1 — a comment on the already-active no-op branch documenting *why* it can safely skip `ApplyRegistration` (the registration-first invariant). No code change. The commit also gives codex a fresh push to re-scan, so the P1 should auto-resolve on its next pass.

`go test -race ./internal/encryption/...` passes; `golangci-lint` clean.

PR #812, opened 2026-05-23
feat(encryption): Stage 6D-6a - EnableStorageEnvelope admin RPC + server method

bootjp:
## Summary

- Stage 6D-6 (final 6D milestone) is sliced into 6D-6a/b/c to match the per-PR review cadence of 6D-3/4/5. This PR lands **6D-6a: proto + server method + server-level tests**. 6D-6b adds the CLI; 6D-6c adds main.go production wiring + end-to-end integration test.
- Proto adds `EnableStorageEnvelope` RPC + `EnableStorageEnvelopeRequest` / `Response` + `CapabilityVerdict` messages per design §3.1. Uses OK gRPC status with `was_already_active` boolean discriminator rather than `AlreadyExists` — unary gRPC drops the response body on non-OK status, so the §6.4 idempotency contract MUST live on the success path.
- `adapter/encryption_admin.go` ships the server method composing the §3.2 sequence: leader gate → input validation → sidecar read → bootstrap gate → idempotent-retry short-circuit (§6.4) → capability fan-out (6D-3) → propose `RotateSubEnableStorageEnvelope` through Raft (6D-4 wire) → post-apply re-read discriminating fresh-success vs. stale-DEKID race.
- New `WithEncryptionAdminCapabilityFanout` option wires the §4 pre-flight. Without it, `EnableStorageEnvelope` refuses with `FailedPrecondition` — matches the proposer / leaderView posture for other mutator RPCs. The 6D-6c main.go wiring closes over the live route snapshot + DialFunc + timeout.

## Test plan

- [x] `go test -race -timeout 240s -run TestEncryptionAdmin_ ./adapter/...` — all green
- [x] `golangci-lint --new-from-rev=origin/main` — 0 issues
- [x] `_RejectsOnFollower` — leader gate + hint embedding
- [x] `_RejectsWithoutSidecarPath`, `_RejectsWithoutCapabilityFanout` — wiring gates
- [x] `_RejectsZeroProposerNodeID`, `_RejectsOversizedLocalEpoch` — §3.1 input validation
- [x] `_RejectsNotBootstrapped` — bootstrap gate with hint
- [x] `_IdempotentRetry` — sidecar already active → OK + `was_already_active=true`, applied_index = original `StorageEnvelopeCutoverIndex`, NO fan-out, NO propose
- [x] `_DefensiveCutoverIndexUnknown` — active sidecar with index=0 (hand-edit hedge) → `CutoverIndexUnknown=true`, fallback to `RaftAppliedIndex`
- [x] `_RejectsOnCapabilityRefusal` — names the refusing node
- [x] `_RejectsOnFanoutError` — fan-out helper itself erroring → FailedPrecondition
- [x] `_HappyPath` — end-to-end fresh cutover with simulated apply effect; wire-level decode confirms `RotateSubEnableStorageEnvelope` + §2.1 constraints
- [x] `_StaleDEKIDRace` — simulated apply does NOT flip sidecar (RotateDEK raced) → FailedPrecondition with retry hint
- [x] `_CapabilitySummaryProjection` — per-row wire-shape pin for the internal `CapabilityVerdict` → `proto.CapabilityVerdict` mapping

## Caller audit (semantic change)

The new RPC is purely additive: no existing method's signature or semantics change. `WithEncryptionAdminCapabilityFanout` is a new option; existing test fixtures that don't wire it leave `s.capabilityFanout=nil`, which only affects `EnableStorageEnvelope` (refuses) — other mutator RPCs are unaffected. `proto.UnimplementedEncryptionAdminServer` carries the `EnableStorageEnvelope` stub returning `codes.Unimplemented`, so test doubles that don't override stay unbroken.

## Self-review (5 lenses)

1. **Data loss** — no. The server method reads the sidecar, proposes via Raft, then re-reads. The 6D-4 applier is the durable writer; this PR only reads pre- and post-apply.
2. **Concurrency / distributed failures** — the propose-then-re-read sequence relies on the apply being visible on the leader's sidecar by the time Propose returns. The Raft proposer's contract (proposeEncryptionEntry returns after the FSM apply completes locally on the leader) is the load-bearing invariant; existing `RotateDEK` / `BootstrapEncryption` paths rely on the same.
3. **Performance** — one extra sidecar read pre- and post-Propose per cutover. Cutover is one-shot per cluster lifetime; the cost is negligible.
4. **Data consistency** — the §6.4 idempotency contract is preserved: on retry, `applied_index` reports the ORIGINAL `StorageEnvelopeCutoverIndex` (not the retry's Raft index). `CapabilitySummary` is intentionally empty on retries so a caller cannot accidentally re-use the original cutover's membership view. The §2.1 #3 stale-DEKID race is surfaced as `FailedPrecondition` with a retry hint (not Aborted; the design pins `FailedPrecondition`).
5. **Test coverage** — 12 server-level tests across the §3.2 sequence + idempotency contract + race discrimination + wire shape. The `applyingProposer` helper pins the §6.4 / §2.1 #3 discrimination logic — the central correctness-critical branch.

Refs: design doc §3.1 / §3.2 / §6.4 (6D-6a slice of the §7 decomposition table).

## Summary by CodeRabbit

* **New Features**
  * Added admin API for encrypted storage envelope management, enabling cluster storage format cutover.

* **Tests**
  * Comprehensive test coverage for envelope functionality, including validation, idempotency, and error handling.

* **Documentation**
  * Design documentation updated to mark server implementation milestone as complete.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/812?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/adapter/encryption_admin.go
+++ b/adapter/encryption_admin.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"runtime/debug"
 	"strconv"
+	"sync"
 
 	"github.com/bootjp/elastickv/internal/admin"
 	"github.com/bootjp/elastickv/internal/encryption"
@@ -38,6 +39,23 @@ type EncryptionAdminServer struct {
 	// is what threads the route-snapshot builder + DialFunc +
 	// timeout into this closure. Other mutator RPCs are unaffected.
 	capabilityFanout CapabilityFanoutFn
+	// cutoverMu serializes concurrent EnableStorageEnvelope calls
+	// per design §2.1 #4 ("The cutover-RPC mutator serializes
+	// overlapping calls on the propose side"). Without it, two
+	// concurrent cutover calls could both observe
+	// StorageEnvelopeActive=false, both propose, and the loser
+	// would assemble a freshCutoverResponse with
+	// was_already_active=false but the FIRST cutover's
+	// applied_index — violating the §6.4 fresh-success contract
+	// (coderabbit Major on PR812). The design also calls for a
+	// shared lock with RotateDEK so a RotateDEK cannot interleave
+	// between a cutover's propose and apply; that piece is left
+	// to a follow-up because RotateDEK has no existing lock and
+	// extending serialization to it would change a hot-path
+	// mutator's semantics — see the §2.1 #3 stale-DEKID benign
+	// no-op for the FALLBACK that handles the un-serialized
+	// RotateDEK/cutover interleave today.
+	cutoverMu sync.Mutex
 	pb.UnimplementedEncryptionAdminServer
 }
 
@@ -676,6 +694,14 @@ func (s *EncryptionAdminServer) RotateDEK(ctx context.Context, req *pb.RotateDEK
 // true with cutover-index mismatch ⇒ another cutover landed
 // concurrently, treat as idempotent success.
 func (s *EncryptionAdminServer) EnableStorageEnvelope(ctx context.Context, req *pb.EnableStorageEnvelopeRequest) (*pb.EnableStorageEnvelopeResponse, error) {
+	// Serialize concurrent cutover RPCs (design §2.1 #4 + PR812
+	// coderabbit Major). The lock spans the entire precheck →
+	// fan-out → propose → postcheck sequence so a second
+	// overlapping call sees StorageEnvelopeActive=true at its
+	// precheck and takes the §6.4 idempotent-retry short-circuit
+	// rather than re-proposing.
+	s.cutoverMu.Lock()
+	defer s.cutoverMu.Unlock()
 	preSidecar, earlyResp, err := s.cutoverPrecheck(ctx, req)
 	if err != nil {
 		return nil, err
@@ -761,6 +787,18 @@ func (s *EncryptionAdminServer) runCutoverFanout(ctx context.Context) (admin.Cap
 		return admin.CapabilityFanoutResult{}, capabilityFanoutErrorToStatus(err)
 	}
 	if !result.OK {
+		// Codex P2 round-2 on PR812: the production fan-out
+		// helper can return (result, nil) with OK=false when ctx
+		// expires mid-probe — it synthesizes Reachable=false
+		// verdicts and returns the result rather than erroring
+		// out. In that case, classifying the outcome as a
+		// configuration refusal (FailedPrecondition) hides the
+		// real transport-layer cancellation/deadline from
+		// retry-aware clients. Check ctx.Err() first so the
+		// gRPC status code matches what the caller observed.
+		if ctxErr := ctx.Err(); ctxErr != nil {
+			return admin.CapabilityFanoutResult{}, capabilityFanoutErrorToStatus(ctxErr)
+		}
 		return admin.CapabilityFanoutResult{}, grpcStatusErrorf(codes.FailedPrecondition,
 			"encryption: capability check refused cutover (%s)", capabilityRefusalSummary(result))
 	}
--- a/adapter/encryption_admin.go
+++ b/adapter/encryption_admin.go
@@ -903,6 +903,14 @@ func freshCutoverResponse(sc *encryption.Sidecar, proposedIdx uint64, fanoutResu
 		// index alongside the active flag. A zero here means
 		// the post-apply read raced (unlikely, but the §6.4
 		// fallback exists for hand-edited sidecars).
+		// CutoverIndexUnknown stays false on this branch — the
+		// proto field is "only meaningful when
+		// was_already_active=true" (§3.1), and here we know the
+		// correct applied_index: it is proposedIdx (the Raft
+		// entry this call just committed). The idempotent-retry
+		// path is the only context where the original cutover
+		// index is irrecoverable, hence its CutoverIndexUnknown
+		// signal. Claude bot informational on PR812.
 		appliedIndex = proposedIdx
 	}
 	return &pb.EnableStorageEnvelopeResponse{
--- a/adapter/encryption_admin.go
+++ b/adapter/encryption_admin.go
@@ -748,7 +748,19 @@ func (s *EncryptionAdminServer) EnableStorageEnvelope(ctx context.Context, req *
 // A plain sync.Mutex would block past the caller's deadline,
 // breaking RPC cancellation semantics (codex P2 round-3 on
 // PR812).
+//
+// The explicit ctx.Err() check before the select is load-bearing:
+// Go's select picks uniformly at random among ready cases (it
+// does NOT prioritize ctx.Done()), so an already-canceled ctx
+// paired with a free semaphore would coin-flip between acquiring
+// the slot (and running precheck / fan-out / propose against an
+// aborted caller) and the cancellation return. Checking
+// ctx.Err() first turns the cancellation into a deterministic
+// short-circuit (codex P1 round-4 on PR812).
 func (s *EncryptionAdminServer) acquireCutoverSemaphore(ctx context.Context) error {
+	if err := ctx.Err(); err != nil {
+		return cutoverSemaphoreErrorToStatus(err)
+	}
 	select {
 	case s.cutoverSem <- struct{}{}:
 		return nil
--- a/adapter/encryption_admin.go
+++ b/adapter/encryption_admin.go
@@ -9,6 +9,7 @@ import (
 	"runtime/debug"
 	"strconv"
 
+	"github.com/bootjp/elastickv/internal/admin"
 	"github.com/bootjp/elastickv/internal/encryption"
 	"github.com/bootjp/elastickv/internal/encryption/fsmwire"
 	"github.com/bootjp/elastickv/internal/raftengine"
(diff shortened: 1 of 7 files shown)


bootjp:
@claude review

Round-1 (commit `55be2579`): addressed codex P2 finding on context-cancellation propagation.

## codex P2 — preserve context errors from capability fan-out

Pre-fix: `runCutoverFanout` wrapped every err from `s.capabilityFanout(ctx)` as FailedPrecondition. When the request deadline expires or the client cancels mid-pre-flight, the client received FailedPrecondition (a configuration-failure signal) instead of Canceled / DeadlineExceeded (the transport-layer cancellation signal). This breaks retry logic that switches on the gRPC code.

Fix: new `capabilityFanoutErrorToStatus` helper maps `context.Canceled` and `context.DeadlineExceeded` to their native gRPC codes (`codes.Canceled` / `codes.DeadlineExceeded`) and preserves FailedPrecondition for everything else. Symmetric with the existing `proposeErrorToStatus` pattern.

Regression test: `TestEncryptionAdmin_EnableStorageEnvelope_PreservesContextCancellation` table-driven across `{context.Canceled, context.DeadlineExceeded}` asserts the gRPC status code matches.

## Caller audit

`runCutoverFanout` is called only from `cutoverPostcheck` which threads the returned error back to gRPC. Code mapping is now narrower (transport-layer errors get their native codes; non-transport errors keep FailedPrecondition). No other internal caller.

## Verification

- `go test -race -timeout 60s -run TestEncryptionAdmin_EnableStorageEnvelope ./adapter/...` — all green
- `golangci-lint --new-from-rev=origin/main` — 0 issues

bootjp:
@claude review

bootjp:
@claude review

Round-2 (commit `c4084495`): addressed claude bot round-1 P2 finding + informational comment.

## claude P2 — `capabilityRefusalSummary` Reachable=false branch untested

Pre-fix: `_RejectsOnCapabilityRefusal` only exercised the `EncryptionCapable=false` branch of `capabilityRefusalSummary`. The `Reachable=false` branch (which produces 'unreachable member …' in the status detail) had no regression coverage.

Fix: convert the test to table-driven across `{not_capable, unreachable}` fixtures. Each case asserts FailedPrecondition + refusing node id + refusal reason label in the status detail.

## claude informational — `freshCutoverResponse` defensive fallback comment

Added a clarifying comment explaining why `CutoverIndexUnknown` stays false on the fresh-success defensive branch (because `proposedIdx` IS the known index — only the idempotent-retry path has an irrecoverable original cutover index).

## claude informational — catch-all in `capabilityRefusalSummary`

Noted but not addressed: the 'fan-out reported OK=false with no per-member refusal' catch-all is structurally unreachable under the current `CapabilityFanout` contract (OK=false ⇒ at least one failing verdict). It's a defensive fallback message, not a correctness path.

## Caller audit

No semantic change. Test refactor widens fixture coverage; production code change is a comment addition. No caller impact.

## Verification

- `go test -race -timeout 60s -run TestEncryptionAdmin_EnableStorageEnvelope ./adapter/...` — all green
- `golangci-lint --new-from-rev=origin/main` — 0 issues

bootjp:
@claude review

Round-3 (commit `a45eeac7`): addressed coderabbit Major + codex P2 + coderabbit quick-win + minor.

## coderabbit Major — serialize concurrent EnableStorageEnvelope calls

Pre-fix: gRPC handlers run concurrently. Two cutover RPCs could both observe `StorageEnvelopeActive=false`, both propose, and the loser would assemble a `freshCutoverResponse` with `was_already_active=false` but the FIRST cutover's `StorageEnvelopeCutoverIndex` — violating the §6.4 fresh-success contract.

Fix: `cutoverMu sync.Mutex` on `EncryptionAdminServer`. `EnableStorageEnvelope` acquires it at the top, releases on defer. The lock spans the entire precheck → fan-out → propose → postcheck so a second overlapping call sees `StorageEnvelopeActive=true` at its precheck and takes the §6.4 idempotent-retry path.

The §2.1 #4 'shared lock with RotateDEK' is left to a follow-up — extending serialization to RotateDEK changes a hot-path mutator's semantics, and §2.1 #3 stale-DEKID benign no-op handles the un-serialized RotateDEK/cutover interleave today. Documented in the `cutoverMu` doc comment.

Regression test: `_SerializesConcurrentCutovers` fires 4 concurrent callers via a release barrier, asserts `proposer.calls == 1` and exactly one caller sees `was_already_active=false` while the rest hit the §6.4 idempotent-retry shape.

## codex P2 — preserve ctx codes when fan-out returns OK=false

Pre-fix: when production `CapabilityFanout` hits ctx-expired mid-probe it returns `(result, nil)` with `OK=false` (synthesizing Reachable=false verdicts). `runCutoverFanout` then classified the outcome as FailedPrecondition, hiding the transport-layer Canceled/DeadlineExceeded.

Fix: in the OK=false branch, check `ctx.Err()` first and route through `capabilityFanoutErrorToStatus` if set.

Regression test: `_PreservesContextCancellationOnFanoutNotOK` table-driven across {canceled, deadline_exceeded}; fan-out fixture returns `(notOK, nil)`; assertion confirms the gRPC code matches the ctx error.

## coderabbit quick-win — idempotent-retry tests must not wire fan-out

Pre-fix: `_IdempotentRetry` and `_DefensiveCutoverIndexUnknown` both wired a successful fan-out fixture. A regression that re-ordered `cutoverPrecheck` so the fan-out fired BEFORE the §6.4 short-circuit would silently pass.

Fix: new `failOnCallCapabilityFanout(t)` helper that `t.Errorf`s if invoked. Both idempotent-retry tests now use it — any re-ordering regression trips the fixture.

## coderabbit minor — design doc obsolete entry

Removed the original combined '6D-6' milestone line; it's been superseded by 6D-6a (shipped) + 6D-6b/c (open).

## Caller audit

`EnableStorageEnvelope` is the only mutator that acquires `cutoverMu`; RotateDEK / BootstrapEncryption / RegisterEncryptionWriter remain un-serialized at the RPC layer (their FSM apply-path conflict detection is unchanged). No production caller of `EnableStorageEnvelope` exists yet (CLI lands in 6D-6b); only server-level unit tests, which exercise the new locked path.

## Verification

- `go test -race -timeout 60s -run TestEncryptionAdmin_EnableStorageEnvelope ./adapter/...` — all green (the -race build also exercises the 4-goroutine serialization test)
- `golangci-lint --new-from-rev=origin/main` — 0 issues

bootjp:
@claude review

Round-4 (commit `f93bea7e`): addressed codex P2 round-3.

## codex P2 round-3 — honor ctx deadlines before blocking on cutover mutex

Pre-fix: `EnableStorageEnvelope` took `cutoverMu (sync.Mutex)` unconditionally. A concurrent caller with a short ctx deadline could block on `Lock()` well past its own cancellation/deadline if the first caller stalled in fan-out or propose — breaking gRPC cancellation semantics and tying up admin handlers.

Fix: replaced `sync.Mutex` with a capacity-1 channel semaphore (`cutoverSem chan struct{}`). New `acquireCutoverSemaphore` uses `select { case sem <- struct{}{}: ...; case <-ctx.Done(): ... }` so an expired ctx returns immediately with the matching gRPC code. New `cutoverSemaphoreErrorToStatus` helper maps the ctx error with status-detail wording distinct from the fan-out path so operators can tell which stage the wait fired in.

`NewEncryptionAdminServer` eagerly initializes the buffered chan (a nil chan would block forever on send, defeating the ctx-aware acquire).

Regression test: `_HonorsCtxDeadlineWaitingOnMutator` drives the first call into a fan-out fixture that blocks indefinitely (holds the semaphore); the second call uses an already-expired ctx and asserts the RPC returns within 500ms with `DeadlineExceeded`.

## Caller audit

`acquireCutoverSemaphore` is called only from `EnableStorageEnvelope`; release is symmetric via `defer`. No other code path touches `cutoverSem`. The round-3 `_SerializesConcurrentCutovers` test still covers mutual exclusion; the new test covers ctx-aware acquisition.

## Verification

- `go test -race -timeout 60s -run TestEncryptionAdmin_EnableStorageEnvelope ./adapter/...` — all green
- `golangci-lint --new-from-rev=origin/main` — 0 issues

bootjp:
@claude review

Round-5 (commit `367de59d`): addressed codex P1 round-4 on `acquireCutoverSemaphore`.

## codex P1 — pre-canceled ctx + free semaphore was non-deterministic

Pre-fix: `acquireCutoverSemaphore` entered a 2-case select (send to `cutoverSem` vs `ctx.Done()`) without first checking `ctx.Err()`. When ctx was already canceled AND the semaphore had capacity, both select cases were immediately ready, and Go's select picks uniformly at random among ready cases — so the acquire could non-deterministically take the send branch and the RPC would proceed into precheck / fan-out / propose against an aborted caller.

Fix: add an explicit `ctx.Err()` check before the select. The select still handles the during-wait cancellation case; the up-front check turns the pre-canceled case into a deterministic short-circuit.

Regression test: `_PreCanceledCtxDeterministicShortCircuit` table-driven across `{canceled, deadline_exceeded}`. Each case constructs a freshly-created server (semaphore empty, send slot available) and runs 100 iterations with a pre-canceled ctx. The fan-out fixture is `failOnCallCapabilityFanout`, which `t.Errorf`'s if invoked — so any iteration that lost the random select coin flip and advanced into precheck would trip the test. Without the explicit `ctx.Err()` check, the loop would have flapped on roughly half the iterations.

## Caller audit

`acquireCutoverSemaphore` is called only from `EnableStorageEnvelope`. The change narrows behaviour: a pre-canceled ctx now ALWAYS returns the matching gRPC code, rather than sometimes running the cutover sequence to completion. Strictly safer.

## Verification

- `go test -race -timeout 60s -run TestEncryptionAdmin_EnableStorageEnvelope ./adapter/...` — all green (the new 100-iteration test passes deterministically)
- `golangci-lint --new-from-rev=origin/main` — 0 issues

PR #813, opened 2026-05-23
internal/admin: HTTP handlers for DynamoDB item RPCs (Phase 3a)

bootjp:
## Summary

**Phase 3a** of `docs/design/2026_05_22_implemented_admin_data_browser.md` §3.3 — exposes the Phase 2a (PR #805) `*adapter.DynamoDBServer` item-level admin RPCs through the admin HTTP API.

## Surface

```
GET    /admin/api/v1/dynamo/tables/{name}/items                — scan
GET    /admin/api/v1/dynamo/tables/{name}/items/{key-b64url}   — get
PUT    /admin/api/v1/dynamo/tables/{name}/items/{key-b64url}   — put
DELETE /admin/api/v1/dynamo/tables/{name}/items/{key-b64url}   — delete
```

`{key}` is base64-url-encoded JSON of the primary-key `map[string]AdminAttributeValue`. Base64-url passes the path validator's `%`-ban cleanly so arbitrary key bytes (non-ASCII, slashes, etc.) traverse without ambiguity.

## Routing

`dispatchPerTableRoute` adopts the SQS handler's 6-step procedure (escape-aware, dot-segment-rejecting, dispatch on segment count). `EscapedPath()` — not `Path` — is consulted so percent-encoded segments (`%2F`, `%2E%2E`, `%252F`) are rejected before decoding. The pre-Phase-3 `/tables/{name}` describe/delete branch is unchanged except for the more uniform 400 invalid_path classification on a bare trailing slash (previously 404 via empty-name passthrough — two existing tests updated).

## Authorization

| Operation | Gate |
|---|---|
| Scan / Get | `principal.Role.AllowsRead()` (RoleReadOnly or RoleFull) — payload is item content, matches SQS peek precedent |
| Put / Delete | `principal.Role.AllowsWrite()` (RoleFull only) via existing `principalForWrite` |

## Sentinels

| Sentinel | Maps to |
|---|---|
| `ErrItemsForbidden` | 403 |
| `ErrItemsNotLeader` | 503 |
| `ErrItemsTableNotFound` | 404 |
| `ErrItemsValidation` | 400 |

`dynamoTablesBridge.translateAdminItemsError` translates the adapter sentinels (`ErrAdminForbidden` / `NotLeader` / `DynamoNotFound` / `DynamoValidation`) onto these.

## Caller audit (semantic-change rule)

- `TablesSource` interface grew 4 methods (`AdminScanItems` / `AdminGetItem` / `AdminPutItem` / `AdminDeleteItem`). All implementations (production bridge + 4 test fixtures) updated in lockstep — the Go compiler enforces this.
- `dispatchPerTableRoute` changes the classification of bare `/tables/` (trailing slash, empty name) from 404 to 400 `invalid_path`. Two tests updated to reflect the semantically-clearer code. The SPA never builds that URL so no production caller depends on the 404 shape.

## Risk

Low. The handler is a pure wrapper over the existing Phase 2a admin RPCs. No new write codepath. The bridge in `main_admin.go` translates wire types and sentinels — both sides are isomorphic structs.

## Self-review (5 passes)

1. **Data loss** — Pure dispatch + bridge layer; no new write path. The adapter's existing chunked-write / Raft-dispatch path is unchanged.
2. **Concurrency** — Read paths re-use the adapter's pinned-readTS lifecycle. Write paths go through the same `*WithRetry` helpers the SigV4 path uses.
3. **Performance** — Scan paginates at `defaultAdminItemScanLimit=25` (cap 100). Key decoding is O(1) base64 + O(N) JSON parse on a primary-key map (usually <10 attributes). No new hot-path allocations.
4. **Consistency** — Body/path-key reconciliation ensures clients can't accidentally `PUT` a key disagreement (full-key body alongside a different `{key}` URL).
5. **Test coverage** — 19 new unit tests covering routing, key decoding, scan/get/put/delete happy paths, RBAC, and sentinel→status translation.

## Test plan

- [x] `go test -race -count=1 -timeout=60s ./internal/admin/...` — passes
- [ ] CI

## Phase plan

- Phase 2a: DynamoDB item RPCs — merged (#805)
- Phase 2b: S3 object RPCs — merged (#811)
- **Phase 3a: this PR (DynamoDB item HTTP)**
- Phase 3b: S3 object HTTP handlers + bridge — next
- Phase 4: SPA DynamoDetail Items tab
- Phase 5: SPA S3Detail Objects tab + Upload
- Phase 6: Doc rename `_proposed_` → `_implemented_`

## New Features
* Item-level operations for DynamoDB tables: scan items with pagination support, retrieve individual items, write items, and delete items via the admin API.
* Enhanced path validation and error handling with appropriate HTTP status codes and retry guidance for sensitive operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/813?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/admin/dynamo_handler.go
+++ b/internal/admin/dynamo_handler.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"log/slog"
 	"net/http"
+	"net/url"
 	"sort"
 	"strconv"
 	"strings"
@@ -282,18 +283,24 @@ const (
 
 func (h *DynamoHandler) dispatchPerTableRoute(w http.ResponseWriter, r *http.Request) {
 	// EscapedPath() exposes percent-encoded segments verbatim so the
-	// `%`-rejection in isValidAdminPathSegment catches things like
-	// %2F (encoded /), %2E%2E (encoded ..), %252F (double-encoded).
-	// r.URL.Path is pre-decoded and would let those slip through.
+	// decode+validate step in decodeAndValidatePathSegment can
+	// distinguish raw `/` (path separator) from `%2F` (a literal `/`
+	// the operator URL-escaped) — both must be rejected as segment
+	// content. r.URL.Path is pre-decoded and would silently turn
+	// `%2F` into a real `/`, collapsing two distinct attack classes
+	// into one.
 	escaped := r.URL.EscapedPath()
 	rest := strings.TrimPrefix(escaped, pathPrefixDynamoTables)
-	segments := dropSingleTrailingEmpty(strings.Split(rest, "/"))
-	for _, seg := range segments {
-		if !isValidAdminPathSegment(seg) {
+	rawSegments := dropSingleTrailingEmpty(strings.Split(rest, "/"))
+	segments := make([]string, 0, len(rawSegments))
+	for _, seg := range rawSegments {
+		decoded, ok := decodeAndValidatePathSegment(seg)
+		if !ok {
 			writeJSONError(w, http.StatusBadRequest, "invalid_path",
-				"path segment is empty, contains %, or is a dot-segment")
+				"path segment is empty, malformed percent-encoding, contains an encoded slash, or is a dot-segment")
 			return
 		}
+		segments = append(segments, decoded)
 	}
 	switch len(segments) {
 	case dynamoRouteSegmentsTable:
@@ -311,7 +318,13 @@ func (h *DynamoHandler) dispatchPerTableRoute(w http.ResponseWriter, r *http.Req
 				"unknown sub-resource on this table")
 			return
 		}
-		h.handleItemResource(w, r, segments[0], segments[2])
+		// The {key} segment is base64-url-encoded by the SPA, so the
+		// decoded form here equals the raw form (base64-url alphabet
+		// is a subset of unreserved URL chars). Pass rawSegments[2]
+		// to handleItemResource so its existing base64 decoder runs
+		// against the wire form — a future format change in the SPA
+		// (e.g. using padded base64) wouldn't drift the contract.
+		h.handleItemResource(w, r, segments[0], rawSegments[2])
 	default:
 		writeJSONError(w, http.StatusNotFound, "unknown_endpoint",
 			"too many path segments")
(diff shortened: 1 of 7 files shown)


bootjp:
@claude review

Phase 3a — DynamoDB item HTTP handlers + bridge for the Phase 2a (#805) item RPCs. 19 unit tests; `golangci-lint` clean; `go test -race ./internal/admin/...` passes.

Self-review notes are in the PR body. Phase 3b (S3 object HTTP) is next on a separate branch.

bootjp:
@claude review

Round 1 addresses Gemini's three medium-priority findings:

- AdminAttributeValue.MarshalJSON preserves nil-vs-empty L/M distinction
- handleItemPut now uses decodeAdminItemBody with NUL-byte + trailing-data + DisallowUnknownFields defence-in-depth (mirrors decodeCreateTableRequest)
- attributeValuesEqual switched to bytes.Equal / slices.Equal / slices.EqualFunc; manual helpers removed

4 new failing-first tests added per CLAUDE.md test-then-fix convention.

go test -race ./internal/admin/... — passes
golangci-lint — clean

bootjp:
@claude review

Round 2 addresses three Codex findings:

- P1: accept percent-encoded table segments — `isValidAdminPathSegment` → `decodeAndValidatePathSegment` (decodes, then rejects decoded `/`, `.`, `..`). Tables named e.g. "foo bar" or "café" are now reachable through describe / delete / items.
- P2: leader-churn classification on `translateAdminItemsError` (was 500, now 503).
- P2: documented the schema-awareness boundary in `assertItemBodyMatchesPathKey` — body cannot declare extra primary-key columns the URL key doesn't name. The principled fix (plumb URL key into AdminPutItem so the adapter can compare) crosses the Phase-2a boundary and is tracked separately.

4 new tests, all failing-first per CLAUDE.md. `go test -race ./internal/admin/...` passes, lint clean.

bootjp:
@claude review

Round 3 addresses three drifts you found in r2:

- **Bug 1 (real)**: `writeItemsError` 503 case now sets `Retry-After: 1` and uses canonical `leader_unavailable` code. Audited all callers of both code strings — every other admin 503 (s3_handler, sqs_handler, forward_server, dynamo_handler) already emits `leader_unavailable`; only items shipped the non-canonical string. New failing-first test `TestDynamoItems_NotLeader503HasRetryAfterAndCanonicalCode` pins the contract.
- **Minor: 500 → 401**: `principalForItemRead` returns 401 for missing session principal, matching `principalForWrite`. Test `TestDynamoItems_Scan_RejectsAnonymous` updated.
- **Minor: sentinel-string leakage**: `writeItemsError` 403 / 503 cases now use fixed operator-facing strings instead of `err.Error()`, matching `writeTablesError`. The 400 `ErrItemsValidation` case keeps `err.Error()` because the bridge translates with operator-facing messages there.

Caller audit clean: both helpers are file-local and callers only check `ok=false` / `err != nil` boolean; no caller branches on the written body or status.

bootjp:
@claude review

Round 4 addresses the follow-up you flagged in r3: `translateAdminItemsError` for `ErrAdminDynamoValidation` now wraps the adapter error via `fmt.Errorf("...: %w", ...)` so the operator-facing reason (legacy-key migration hint, etc.) reaches the HTTP 400 body. `errors.Is(out, admin.ErrItemsValidation)` still matches so the handler's match arm fires unchanged.

Caller audit: 4 bridge methods, all forward via the TablesSource interface. Handlers use `errors.Is` for matching (transparent through %w); only the 400 case passes `err.Error()` through to the response body, so only the ValidationError arm needed the wrap.

New failing-first test `TestTranslateAdminItemsError_ValidationPreservesAdapterMessage` pins both invariants.

bootjp:
@claude review

Round 5 lands the principalForItemRead JWT-first fix in parallel with PR #814 r3's principalForObjectRead fix — same defence-in-depth gap, same fix shape. JWT role check fires unconditionally before the optional live-store re-validation; the live store can only further constrain, never relax.

bootjp:
@claude review

Round 6 adds the missing regression test you flagged in r5: `TestDynamoItems_Scan_JWTNoneRoleRejectedWhenLiveStoreGrantsRead` exercises the JWT-bypass scenario (JWT Role="", live store grants RoleReadOnly) and asserts 403. Mirrors `TestDynamoHandler_CreateTable_LiveRoleRevocation` on the write side.

Re. CodeRabbit's "Preserve adapter validation details" major from before r4 — that's exactly what r4 fixed (`translateAdminItemsError` now wraps via `fmt.Errorf %w`). The bot reviewed an earlier snapshot.

bootjp:
@claude review

Round 7 closes the dot-segment table-name orphaning Codex flagged as P2. `validateCreateTableRequest` now rejects `.` and `..` symmetrically with the existing `/` rejection. Same orphaning class — a table created with one of those names would hit 400 invalid_path on every per-table URL afterward.

New failing-first test `TestDynamoHandler_CreateTable_RejectsDotSegmentNames` covers both single-dot and double-dot. Caller audit: `validateCreateTableRequest` is only called by `decodeCreateTableRequest`; the added rejection is strictly fail-closed.

bootjp:
Re. the Codex P2 on `decodeAdminItemBody` line 289 — this is a false positive. Empirical verification with the actual decoder library (`goccy/go-json`):

```
input="{\"a\":1}"          decode=<nil>  More=false
input="{\"a\":1}\n"         decode=<nil>  More=false   ← trailing newline accepted
input="{\"a\":1}   "         decode=<nil>  More=false   ← trailing spaces accepted
input="{\"a\":1}\n\n\n"    decode=<nil>  More=false   ← multiple newlines accepted
input="{\"a\":1} \t\n"      decode=<nil>  More=false   ← mixed whitespace accepted
input="{\"a\":1}{\"b\":2}" decode=<nil>  More=true    ← actual trailing JSON correctly rejected
input="{\"a\":1} {\"b\":2}" decode=<nil>  More=true   ← whitespace + JSON correctly rejected
```

`goccy/go-json`'s `More()` peeks past whitespace internally before deciding whether more JSON tokens exist; pure trailing whitespace returns `false`. The pretty-printed / `curl --data-binary` cases the finding mentions all decode cleanly today. Test exists in `TestDynamoItems_Put_RejectsTrailingJSONData` which only fails when there's an actual second JSON value.

Closing this finding as not-a-bug. (Codex may have been thinking of stdlib `encoding/json` which has different semantics.)

bootjp:
@claude review

bootjp:
@claude review

Round 8 addresses Codex P2 ([#3293505226](https://github.com/bootjp/elastickv/pull/813#discussion_r3293505226)): `dispatchPerTableRoute` was passing `rawSegments[2]` for the items {key} route, so a client that URL-escapes padded base64's '=' to '%3D' (encodeURIComponent does this) had the base64 decoder reject the value with 'illegal base64 data at input byte 23' — defeating the documented padded-base64 fallback in `decodeAdminItemKeySegment2`.

Fix: pass `segments[2]` (percent-decoded). Raw-form SPA keys still work because the decoded form equals the raw form for the raw-base64-url alphabet (A-Z a-z 0-9 - _, no reserved chars).

New failing-first test `TestDynamoItems_Get_AcceptsURLEncodedPaddedBase64Key` pins the round-trip; pre-fix 400, post-fix 200.

Rebased onto the latest main (includes #818's Redis flake fix), so CI should now run cleanly.

bootjp:
@claude review

Round 9 addresses Codex P2 (empty/nil decoded `next_cursor` silently masquerading as 'start over').

Fix: `parseAdminScanItemsQuery` now rejects `len(cursor) == 0` after decode — symmetric with `decodeAdminItemKeySegment`'s existing len-zero guard on the URL {key} segment. Closes the duplicate-page-0 / paging-loop trap when a malformed client cursor decodes to `null` or `{}`.

New failing-first test `TestDynamoItems_Scan_RejectsEmptyDecodedCursor` covers 'null', '{}', and padded-base64 '{}'. Caller audit: `parseAdminScanItemsQuery` is only called from `handleItemsCollection`; new branch follows the existing ok=false pattern.

PR #814, opened 2026-05-23
admin: HTTP handlers for S3 object RPCs (Phase 3b)

bootjp:
## Summary

**Phase 3b** of `docs/design/2026_05_22_proposed_admin_data_browser.md` §3.3 — exposes the Phase 2b (PR #811) `*adapter.S3Server` object-level admin RPCs through the admin HTTP API.

## Surface

```
GET    /admin/api/v1/s3/buckets/{name}/objects                  — list
GET    /admin/api/v1/s3/buckets/{name}/objects/{key-b64url}     — download
PUT    /admin/api/v1/s3/buckets/{name}/objects/{key-b64url}     — upload
DELETE /admin/api/v1/s3/buckets/{name}/objects/{key-b64url}     — delete
```

`{key}` is the base64-url-encoded raw object key.

## Body wire shapes

| Op | Request | Response |
|---|---|---|
| LIST | — | JSON `AdminObjectListing` |
| GET | — | `application/octet-stream` (raw bytes) + security headers |
| PUT | `application/octet-stream` (raw bytes) | — |
| DELETE | — | — |

## Security headers on GET

Defence-in-depth against a hostile upload rendering in the admin origin:

| Header | Value | Why |
|---|---|---|
| `X-Content-Type-Options` | `nosniff` | Block MIME sniffing |
| `Content-Security-Policy` | `sandbox; default-src 'none'` | Strip script execution |
| `Content-Disposition` | `attachment; filename="..."` | Force download dialog |
| `Cache-Control` | `no-store` | No intermediary caching |

Filename is sanitised — control chars / quote / backslash → `_` so the quoted-string form is never broken by an attacker-controlled key.

## Upload limits

| Limit | Where | Outcome |
|---|---|---|
| 100 MiB | `http.MaxBytesReader` in `handleObjectPut` | 413 payload_too_large |
| Adapter cap | `ErrAdminUploadTooLarge` via bridge | 413 payload_too_large |

Both paths produce the same 413 response shape so the SPA retry contract is uniform.

## Authorization

| Operation | Gate |
|---|---|
| List / Get | `principal.Role.AllowsRead()` (RoleReadOnly or RoleFull) |
| Put / Delete | `principal.Role.AllowsWrite()` (RoleFull only) |

Read gate matches the existing SPA-facing precedent (`principalForItemRead` on Dynamo). Write gate reuses `S3Handler.principalForWrite` (live-role re-check + RoleFull).

## Sentinels

| Sentinel | Maps to |
|---|---|
| `ErrObjectsForbidden` | 403 |
| `ErrObjectsNotLeader` | 503 + Retry-After: 1 |
| `ErrObjectsBucketNotFound` | 404 |
| `ErrObjectsNotFound` | 404 |
| `ErrObjectsValidation` | 400 |
| `ErrObjectsUploadTooLarge` | 413 |

`bucketsBridge.translateAdminObjectsError` translates the adapter sentinels (`ErrAdminForbidden` / `ErrAdminNotLeader` / `ErrAdminBucketNotFound` / `ErrAdminObjectNotFound` / `ErrAdminUploadTooLarge` / `ErrAdminInvalidBucketName` / `ErrAdminInvalidObjectKey` / `ErrAdminInvalidContinuationToken`) onto these.

## Routing

`servePerBucket` now recognises three sub-tree shapes under `/buckets/{name}/`:

1. `/acl` PUT (existing)
2. `/objects` GET — list (new)
3. `/objects/{key}` GET/PUT/DELETE — resource (new)

The objects branch is matched first via `splitBucketObjectsRoute`. Since `validateS3BucketName` already forbids `/` in bucket names, a bucket literally named `objects` cannot collide with the describe-by-name path.

Existing `/buckets/{name}` describe/delete and `/buckets/{name}/acl` PUT behaviour are unchanged.

## Caller audit (semantic-change rule)

- `BucketsSource` interface grew 4 methods (`AdminListObjects` / `AdminGetObject` / `AdminPutObject` / `AdminDeleteObject`). All implementations updated in lockstep:
  - `bucketsBridge` (production, `main_admin.go`)
  - `stubBucketsSource` (tests, `internal/admin/s3_handler_test.go`) — stubs return "not implemented" so existing bucket-tier tests fail loudly if they accidentally exercise the object methods
  - `dummyBucketsSource` (tests, `main_admin_forward_test.go`) — panicking stubs
- The Go compiler enforces this.

## Risk

Low. The handler is a pure wrapper over the existing Phase 2b admin RPCs. No new write codepath. The bridge in `main_admin.go` translates wire types and sentinels — both sides are isomorphic structs.

## Self-review (5 passes)

1. **Data loss** — Pure dispatch + bridge layer; no new write path on the storage side. The adapter's existing chunked-write / Raft path is unchanged.
2. **Concurrency** — Read body Close is in a defer; client cancellation mid-stream logs at warn and is not treated as error.
3. **Performance** — List paginates at default 100 (cap 1000). PUT streams to the adapter via `http.MaxBytesReader` without buffering the full body in the admin layer. GET streams from the adapter via `io.Copy`.
4. **Consistency** — Empty lists serialize as `"objects":[]` (not null) so the SPA distinguishes empty bucket from missing bucket. Security headers are unconditional on GET.
5. **Test coverage** — 18 new tests covering routing / key decoding / list / get with headers / get 404 / put round-trip / put 413 (both `http.MaxBytesReader` and adapter-sentinel paths) / delete idempotence / RBAC / sentinel translation / `objectKeyBasename`.

## Test plan

- [x] `go test -race -count=1 -timeout=120s ./internal/admin/...` — passes
- [ ] CI

## Phase plan

- Phase 2a: DynamoDB item RPCs — merged (#805)
- Phase 2b: S3 object RPCs — merged (#811)
- Phase 3a: DynamoDB item HTTP — open (#813)
- **Phase 3b: this PR (S3 object HTTP)**
- Phase 4: SPA DynamoDetail Items tab
- Phase 5: SPA S3Detail Objects tab + Upload
- Phase 6: Doc rename `_proposed_` → `_implemented_`

### DIFF
--- a/internal/admin/buckets_source.go
+++ b/internal/admin/buckets_source.go
@@ -3,6 +3,7 @@ package admin
 import (
 	"context"
 	"errors"
+	"io"
 )
 
 // BucketsSource is the in-process surface the admin S3 handler
@@ -47,6 +48,33 @@ type BucketsSource interface {
 	// when objects remain (the dashboard cannot force a recursive
 	// delete; the SPA renders the 409 with a hint to clean up first).
 	AdminDeleteBucket(ctx context.Context, principal AuthPrincipal, name string) error
+
+	// Object-level admin RPCs (Phase 3b — design §3.1.2). The handler
+	// in s3_objects_handler.go drives these for the
+	// /buckets/{name}/objects{,/{key}} sub-tree. Implementations live
+	// in main_admin.go (bucketsBridge) and bridge to the adapter
+	// package's Admin*Object methods.
+
+	// AdminListObjects returns a bounded page of objects under a
+	// prefix. Delimiter folds pseudo-folders into CommonPrefixes
+	// (only emitted when Options.Delimiter is set).
+	AdminListObjects(ctx context.Context, principal AuthPrincipal, bucket string, opts AdminListObjectsOptions) (AdminObjectListing, error)
+	// AdminGetObject streams one object's body + metadata. The caller
+	// MUST close the returned body to release the read-tracker pin.
+	// (body, meta, nil) on success; (nil, AdminObject{}, err) on any
+	// failure. ErrObjectsNotFound flags the object-absent case.
+	AdminGetObject(ctx context.Context, principal AuthPrincipal, bucket, key string) (io.ReadCloser, AdminObject, error)
+	// AdminPutObject creates-or-replaces one object from a streaming
+	// body. ContentType is the operator-supplied Content-Type header
+	// (empty falls back to the adapter's defaultS3ContentType).
+	// Returns ErrObjectsUploadTooLarge if the body exceeds the admin
+	// per-object cap.
+	AdminPutObject(ctx context.Context, principal AuthPrincipal, bucket, key string, body io.Reader, contentType string) error
+	// AdminDeleteObject removes one object. Idempotent — a missing
+	// object surfaces as success (mirrors the SigV4 DeleteObject
+	// contract). Returns ErrObjectsBucketNotFound when the bucket
+	// itself is absent.
+	AdminDeleteObject(ctx context.Context, principal AuthPrincipal, bucket, key string) error
 }
 
 // BucketSummary is the bucket-level DTO the SPA receives. The JSON
--- a/internal/admin/s3_handler.go
+++ b/internal/admin/s3_handler.go
@@ -209,10 +209,11 @@ func (h *S3Handler) dispatchACLSubresource(w http.ResponseWriter, r *http.Reques
 //   - bucket name must be non-empty and slash-free (otherwise
 //     reject via the standard 404 fallthrough)
 //   - the literal "objects" sub-resource must be present
-//   - for the per-object route, the residual segment must not be
-//     empty (a trailing `/objects/` with no key is the collection
-//     shape, but we surface that as a 404 since the SPA never
-//     constructs it and a typo should be loud)
+//   - a residual that is empty OR has a leading slash is the
+//     objects route: empty returns (name, "", true) so the
+//     dispatcher routes to handleObjectsCollection (same shape
+//     as /buckets/{name}/objects); a leading-slash residual
+//     returns (name, key, true) for the per-object route
 //
 // The {key} validator (no `%`, no dot-segments) runs in the
 // object handler's segment decoder rather than here so the route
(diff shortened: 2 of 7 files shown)


bootjp:
@claude review

Phase 3b: S3 object HTTP handlers + bridge.

- 4 new routes: list, get (octet-stream + security headers), put (octet-stream + 100 MiB cap), delete
- Bridge in main_admin.go with sentinel translation
- 18 new tests covering routing / RBAC / sentinels / security headers / 413 paths

go test -race ./internal/admin/... — passes
golangci-lint — clean

bootjp:
@claude review

Round 1 addresses Codex P1: `r.URL.Path` was pre-decoded so `%2F` in the bucket segment collapsed into a literal `/`, rerouting the request to a different bucket (confused-deputy class). Fixed by switching `servePerBucket` to `r.URL.EscapedPath()`.

New failing-first test `TestS3Objects_RoutingRejectsEncodedSlashInBucket` covers GET/PUT/DELETE object + GET list under the bypass URL. `go test -race ./internal/admin/...` passes, lint clean.

bootjp:
@claude review

Round 2 addresses three items from your r1 review:

- **Medium**: stale comments in `splitBucketObjectsRoute` and `dispatchObjectsSubtree` rewritten to match actual collection-route behavior on empty residual / sub.
- **Low**: nil-body guard added to `handleObjectGet` so a misbehaving adapter (nil body, nil err) surfaces a 500 + log instead of panicking on `defer body.Close()`.
- **Low**: `TestS3Objects_Put_Rejects413OnOversizedBody` switched from a 100 MiB `bytes.Repeat` allocation to `io.LimitReader(infiniteX{}, cap+1)` — same MaxBytesReader trip-point, ~1 KiB memory footprint.

The validation-error-format inconsistency (object tier wraps via `errors.Wrap(ErrObjectsValidation, ...)` vs bucket tier's `*ValidationError`) is left as-is — both surface the operator-facing message, the wrapped-sentinel string is a minor SPA cosmetic issue I'd rather not change without coordinating with the bucket-tier path.

bootjp:
@claude review

Round 3 addresses the principalForObjectRead asymmetry you flagged in r2: the JWT role gate now fires unconditionally before the optional live-store re-validation, matching principalForWrite's defence-in-depth pattern (PR #669). A JWT with role=none can no longer slip through reads via a live-store grant.

Caller audit: 2 callers (handleObjectsCollection, handleObjectGet), both bail on ok=false without inspecting the written response. Strictly fail-closed change.

The parallel principalForItemRead in dynamo_items_handler.go has the same asymmetry; that fix is being applied on PR #813 to keep the two PR diffs scoped.

bootjp:
@claude review

Round 4 documents the object-tier leader-forwarding gap Codex flagged as P1 in r3.

Scope decision: defer to a follow-up PR. The objection is real, but the proper fix isn't a one-liner:

- GetObject and PutObject stream bodies up to 100 MiB — the existing unary `AdminForward` proto (`bytes Payload`) doesn't accommodate that. Streaming gRPC RPCs are required.
- That's ~5 packages of churn: proto regen, server handlers, client methods, bridge, tests.

The current 503 + Retry-After: 1 + SPA retry-via-leader-probe escape hatch matches what the SQS admin handler (writeQueuesError) ships with today; bucket-tier forwarding landed in a later slice after the initial bucket HTTP shipped, on the same trajectory.

I've added a comment on `writeObjectsError` explaining the intentional deferral so a future reader doesn't assume it was missed. Open to discussion if you think this should block the PR.

PR #815, opened 2026-05-23
web/admin: Items tab on DynamoDetail (Phase 4)

bootjp:
## Summary

**Phase 4** of `docs/design/2026_05_22_proposed_admin_data_browser.md` — adds the Items tab to the DynamoDB detail page so operators can scan / view / add / edit / delete items through the admin SPA. Consumes the Phase-3a HTTP surface from PR #813.

## Stacked on

This branch is built on `feat/admin-http-dynamo-items` (PR #813). Merge order: #813 → #814 (Phase 3b) → this PR. CI will compile cleanly here because the Phase-3a server endpoints live in the parent branch.

## Capabilities

| Role | Allowed |
|---|---|
| `read_only` | scan + view item JSON |
| `full` | add + edit + delete (plus everything above) |

## UI

- New "Items" card on `DynamoDetail` with a paginated table.
- Page-size selector (25 / 100) mirrors the server's default + cap.
- "Refresh" reloads the current page; "Next page →" advances via the server's `last_evaluated_key`.
- Per-row "Open" button opens a single modal that switches between view / edit / add modes.
- Delete uses a two-stage confirm (button row collapses to "Delete this item?") so a click cannot accidentally destroy a row.

## API client

`web/admin/src/api/client.ts` grows:

- `AdminAttributeValue` / `AdminItem` / `AdminScanItemsResult` / `AdminScanItemsOptions` types (mirror the Go handler shapes, including the empty-L/M preservation fixed in PR #813 r1).
- `scanItems` / `getItem` / `putItem` / `deleteItem` methods.
- `encodeAdminItemKey` helper: UTF-8 → bytes → btoa → url-safe alphabet, padding stripped. Non-ASCII keys and binary blobs round-trip without ambiguity.

## Server-side rules honoured

- URL key derives from the body's primary-key attributes (`extractPrimaryKey`). The server's `assertItemBodyMatchesPathKey` is authoritative; this client-side extraction only guards against the operator forgetting the partition key entirely.
- Modal `busy` gates Esc / backdrop dismissal so a half-completed save cannot be orphaned mid-network.

## Self-review (5 passes)

1. **Data loss** — No SPA-side write path bypasses the server. Body / URL-key check happens on both sides.
2. **Concurrency** — `AbortController` wired through every fetch. Modal `busy` blocks dismissal.
3. **Performance** — Scan pages bounded. No N+1 per-row Describe calls. Cursor stack lets Refresh reload the current page without restarting the scan.
[...]

PR #818, opened 2026-05-23
test(redis): widen TTL window on ExpiredKey_BecomesInvisible to fix CI flake

bootjp:
## Summary

Fixes the recurring `TestRedis_ExpiredKey_BecomesInvisible` CI flake that has been blocking unrelated admin PRs (#813, #814, #815, #816) all day.

## Root cause

The test sets a key with `PX 200` (200ms TTL) then immediately reads it back:

```go
require.NoError(t, rdb.Do(ctx, "SET", "expiry:short", "v", "PX", "200").Err())

got, err := rdb.Get(ctx, "expiry:short").Result()
require.NoError(t, err)                                          // ← fails: redis.Nil
require.Equal(t, "v", got, "key must be visible before expiry")
```

The `SET` goes through a 3-node Raft cluster: client → leader proposes → quorum apply → leader applies → response. Under `-race` on CI runners this round-trip can take 100–250ms. The TTL clock starts at the moment the leader applies, so by the time `SET` returns OK on a slow CI runner, the key may have <50ms of TTL left. The immediately-following `GET` then loses the race and hits `redis.Nil`.

That this is the bug — not anything in any of #813-#816 — is locked in by the cross-PR failure pattern: the flake fired today on PRs touching only `internal/admin/` and `web/admin/`, with zero code overlap with `adapter/`.

## Fix

`PX 200 → PX 2000`. 2s is comfortably past the worst observed SET-ack latency (~250ms on CI) while keeping the test under a 7s wall-clock budget (2s TTL + 5s `Eventually` deadline, raised from 1s).

The test's intent — verify expired keys become invisible to subsequent GETs — is preserved exactly. Only the timing constants change; the assertions are unchanged.

## Self-review (5 passes)

1. **Data loss** — none. Test-only constants.
2. **Concurrency** — closes a wall-clock vs. Raft-apply-latency race in the test fixture.
3. **Performance** — test now takes ~3s instead of ~250ms locally; still well under the 30s package timeout.
4. **Consistency** — the broader test still asserts both halves (visible before expiry, invisible after expiry) of the TTL contract.
5. **Test coverage** — no change to coverage; same assertions, friendlier timing.

## Test plan

- [x] `go test -race -count=3 -timeout=180s -run TestRedis_ExpiredKey_BecomesInvisible ./adapter/` — passes 3/3 (10.792s)
- [ ] CI

## Context

Discovered while running the PR review loop on #813-#817. The Claude bot's Round 8 review on #813 marked the PR as "Ready to merge once CI passes" — only the Redis flake was blocking. This fix unblocks #813 (and the parallel admin PRs) without touching any admin code.

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of the Redis TTL compatibility test by adjusting timing parameters to reduce intermittent test failures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/818?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/adapter/redis_ttl_compat_test.go
+++ b/adapter/redis_ttl_compat_test.go
@@ -186,15 +186,33 @@ func TestRedis_ExpiredKey_BecomesInvisible(t *testing.T) {
 	rdb := redis.NewClient(&redis.Options{Addr: nodes[0].redisAddress})
 	defer func() { _ = rdb.Close() }()
 
-	// Set a key with a very short TTL.
-	require.NoError(t, rdb.Do(ctx, "SET", "expiry:short", "v", "PX", "200").Err())
+	// PX=2000ms (was 200ms). The previous 200ms TTL races against
+	// the SET → GET round-trip latency on a 3-node Raft cluster
+	// under -race on CI runners: SET proposes through Raft, waits
+	// for quorum apply, and then returns OK, which can take
+	// 100–250ms on a slow runner. By the time the "visible before
+	// expiry" GET fires, the wall-clock-based TTL has already
+	// fired and the GET returns redis.Nil — a recurring flake
+	// observed across PRs #813/#814/#815/#816 today, all of which
+	// share no code with the Redis adapter.
+	//
+	// 2s is comfortably past the worst observed SET-ack latency
+	// (~250ms on CI) while still letting the "must be gone after
+	// expiry" loop's 5s deadline (raised below) catch the expiry
+	// well before the test deadline.
+	const initialTTL = 2 * time.Second
+	require.NoError(t, rdb.Do(ctx, "SET", "expiry:short", "v", "PX",
+		initialTTL.Milliseconds()).Err())
 
 	got, err := rdb.Get(ctx, "expiry:short").Result()
 	require.NoError(t, err)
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

New PR to fix the recurring `TestRedis_ExpiredKey_BecomesInvisible` flake that has been blocking #813-#816 CI all day.

Root cause: `PX 200` (200ms TTL) races against 3-node Raft SET round-trip latency (100-250ms under -race). Fix: `PX 2000` + Eventually deadline raised 1s→5s.

Tested locally with -race -count=3 — passes 3/3. Test intent unchanged; only timing constants.

bootjp:
@claude review

Round 1 addresses both findings (1 Claude bot, 2 Gemini medium):

- **CLAUDE.md violation** — removed PR-number references and 'today' phrasing from inline comments. Cross-PR diagnostic context lives in the commit message / PR description where it belongs.
- **Eventually deadline derives from initialTTL** — `initialTTL + 3*time.Second` keeps the assertion window in sync if the TTL is adjusted in future.

No behavior change. Same 2s TTL, same effective 5s assertion deadline; tested locally with -race -count=2.

PR #823, opened 2026-05-25
docs/design: propose Phase 0b snapshot logical encoder

bootjp:
## Summary

Doc-first PR for **Phase 0b** of the snapshot logical backup work (CLAUDE.md requires a `*_proposed_*` design doc to land before any wire-format implementation). Two commits:

1. **Promote** `snapshot_logical_decoder` `proposed`→`partial` (Phase 0a fully shipped: PRs #790/#791/#792/#806/#810) in a dedicated commit per the `docs/design/README` lifecycle convention.
2. **Propose** `2026_05_25_proposed_snapshot_logical_encoder.md` — `cmd/elastickv-snapshot-encode`, the inverse of the Phase 0a decoder.

## Why a separate design doc

The encoder is not a mechanical mirror of the decoder. Three decisions arise only on the encode side, each a wire-format decision the parent doc left at sketch level:

- **Internal-index reconstruction** (the load-bearing decision): the decoder *drops* every re-derivable internal index (Redis TTL scan index, DynamoDB GSI rows, SQS vis/dedup/group/by-age side records, per-scope generation counters). A *loadable* `.fsm` must contain them or the restored node serves wrong results (TTL'd keys never expire, GSI queries return nothing). The encoder must rebuild the full internal keyspace, mirroring the live adapter index builders (duplicated into `internal/backup` behind the same offline-tool boundary + staleness-review discipline already used for the snapshot-reader constants).
- **MVCC re-encoding**: the directory tree carries no per-key `commit_ts` (decode discards it), so the encoder stamps every key with `invTS = ^last_commit_ts` from `MANIFEST.json`. Keeps every restored row at-or-below the HLC ceiling seeded from the snapshot header.
- **No CRC32C footer**: the parent doc's format sketch showed a trailing CRC; that framing is the *MVCC streaming-restore* path, not the native EKVPBBL1 snapshot the decoder reads / the encoder must emit. Authoritative target format pinned down against `store/snapshot_pebble.go` `WriteTo` + `internal/backup/snapshot_reader.go` `ReadSnapshot`.

## Contents

- Authoritative `.fsm` target format (sorted entries, size caps, cleartext-only).
- Per-adapter reverse-encoder breakdown (Redis / DynamoDB / S3 / SQS), route-for-route against `internal/backup/decode.go`.
- Directory-level round-trip self-test (`dir -> encode -> .fsm -> decode -> dir'`, exact; reverse `.fsm`-byte-identical is explicitly a non-goal).
[...]

PR #825, opened 2026-05-25
backup: Phase 0b M1 — snapshot encoder core (EKVPBBL1 writer + MVCC re-encoding)

bootjp:
## Summary

**Phase 0b M1 — snapshot encoder core.** The format-level machinery every per-adapter reverse-encoder (M2–M5) will share, the inverse of `decode.go`'s dispatch. No adapter directory-walking yet — that lands per-adapter in the following milestones.

Design: `docs/design/2026_05_25_proposed_snapshot_logical_encoder.md` (merged in #823).

`internal/backup/encode.go`:
- **`ResolveCommitTS`** — `--last-commit-ts` override resolution with the fail-closed contract: an override is accepted only when `>= manifest.last_commit_ts` (raising the restored HLC ceiling is safe; lowering it risks a post-restart timestamp colliding with a restored row → `ErrLastCommitTSRegression`).
- **`encodeMVCCKey` / `encodeMVCCValue`** — exact inverse of `snapshot_reader.go::decodeSnapshotEntry`: `<userKey><^commitTS BE>` key, `<flags=0><expireAt LE><body>` value. Reuses the same MVCC constants the decode direction defines so the two cannot drift.
- **`snapshotBuilder`** — accumulates MVCC-framed entries; fails closed on duplicate encoded keys (`ErrEncodeDuplicateKey`) and per-entry size-cap violations (`ErrEncodeKeyTooLarge` / `ErrEncodeValueTooLarge`); `WriteTo` emits the sorted EKVPBBL1 stream (magic + `lastCommitTS` LE + length-prefixed KV pairs, **no checksum footer** — matches `store/snapshot_pebble.go::WriteTo`).
- **`DecodeLiveEntries`** — the round-trip self-test primitive: decodes a just-written buffer back through `ReadSnapshot` so a caller never emits an unloadable `.fsm`.

## Risk

New code only; no existing behavior changed. The package's offline-tool boundary is preserved (no live-cluster imports). `ResolveCommitTS` introduces fail-closed semantics but has no production caller yet — the only caller is the test; the production caller lands with the M6 CLI.

## Self-review (5 lenses)

- **Data loss**: encoder writes live records only, never tombstones; `DecodeLiveEntries` verifies decodability before finalize (the design's round-trip gate).
- **Concurrency/distributed**: `snapshotBuilder` is single-goroutine by contract (adapters feed sequentially); no shared state.
- **Performance**: in-memory sort; worst-case bound documented in the design (external-sort follow-up milestone).
- **Consistency**: uniform `commitTS` stamping keeps every key at/below the restored HLC ceiling; fail-closed override prevents a ceiling regression.
- **Test coverage**: MVCC key/value layout, decode round-trip via the **real** `decodeSnapshotEntry`, single-Redis-string `build→write→DecodeLiveEntries` round-trip, sorted output regardless of Add order, duplicate-key rejection, oversize-key rejection, `ResolveCommitTS` override table (incl. fail-closed regression).

## Test plan

- [x] `go test ./internal/backup/` (full package, no regression)

### DIFF
--- a/internal/backup/encode.go
+++ b/internal/backup/encode.go
@@ -150,6 +150,16 @@ func newSnapshotBuilder(commitTS uint64) *snapshotBuilder {
 // order-dependent `.fsm`. userKey/userValue are copied — callers may
 // reuse their buffers after Add returns.
 func (b *snapshotBuilder) Add(userKey, userValue []byte, expireAt uint64) error {
+	// A builder is single-use. Once WriteTo has flushed, any further
+	// Add would stage entries that can never be written (the second
+	// WriteTo fails closed before re-sorting), silently dropping
+	// records. Reject with the same sentinel WriteTo uses so a caller
+	// reusing an exhausted builder gets one consistent signal
+	// regardless of which method it calls (claude review on PR #825 —
+	// a silent-data-loss footgun for the M2-M5 adapter feed loops).
+	if b.written {
+		return errors.WithStack(ErrSnapshotBuilderReused)
+	}
 	// Size-check from the user-buffer lengths before allocating the
 	// framed buffers — encKey = userKey + snapshotTSSize, encVal =
 	// snapshotValueHeaderSize + userValue — so an oversize record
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

PR #831, opened 2026-05-25
backup: Phase 0b M2b — Redis wide-column reverse encoders (hash/set/list/zset/stream)

bootjp:
## Summary

**Phase 0b M2b — Redis wide-column collection reverse encoders.** Completes the Redis reverse encoder (M2) on top of M2a (strings + HLL, #828): the inverse of the hash/set/list/zset/stream decoders. Builds on M1's `snapshotBuilder` (#825). Design: `docs/design/2026_05_25_proposed_snapshot_logical_encoder.md`.

Per-type reverse encoders in `internal/backup/encode_redis_coll.go`, each pinned against the **live store** key/value layout so the emitted `.fsm` loads into a running cluster:

- **hash** — `!hs|meta|` (8-byte BE field count) + `!hs|fld|` per field. Emits only the consolidated base meta (no `!hs|meta|d|` deltas — the post-compaction steady state).
- **set** — `!st|meta|` count + `!st|mem|<member>` (empty value, matching the live SADD write).
- **list** — `!lst|meta|` (24-byte `[Head][Tail][Len]`) + `!lst|itm|<sortableInt64(seq)>`. The dump records order but not seqs, so items get the canonical contiguous form (Head=0, seq=i, Tail=Len); **order is preserved**.
- **zset** — both indexes: `!zs|mem|<member>` → 8-byte BE float score, and `!zs|scr|<sortableFloat64><member>` → empty range index. `+inf`/`-inf` handled; NaN rejected.
- **stream** — `!stream|meta|` (24-byte `[Length][LastMs][LastSeq]`) + `!stream|entry|` whose value is the magic-prefixed `pb.RedisStreamEntry{Id, Fields}` the live store writes (Id set so XRANGE/XREAD report the right ID). JSONL parsed line-by-line; `_meta` line drives meta + TTL.

Cross-cutting:
- Non-string collection TTL is emitted as an `!redis|ttl|` scan-index row (matching `buildTTLElems`), unlike strings' inline TTL.
- Binary field/member names round-trip through `unmarshalRedisBinaryValue` (inverse of the decoder's `{"base64":...}` envelope). Stream field values must be valid UTF-8 — the proto3 string field rejects non-UTF-8 on both the live write and this encoder.
- `walkJSONDir` reads through `os.Root` + `lstatDumpDir` symlink-refusal + `IsRegular` + `refuseHardLink`, inheriting M2a's full read-path hardening.

## Risk

New code only; no existing behavior changed. Offline-tool boundary preserved.

[...]

PR #834, opened 2026-05-25
backup: post-open IsRegular re-check in readRootFile (claude review #831)

bootjp:
## Summary

**Post-open `IsRegular` re-check in `readRootFile`** — a security hardening claude flagged during the PR #831 review that merged without the fix.

`readRootFile` (the Redis blob `*.bin` reader) is reached via `walkBlobDir` → `handleBlobEntry`, which pre-filters entries by the **ReadDir** entry type (`ent.Type().IsRegular()`). That type is **stale**: a FIFO/device swapped in for a `*.bin` between `ReadDir` and `root.Open` passes the stale check and reaches `io.ReadAll` — blocking on a writer-attached FIFO, or ingesting attacker-controlled bytes. This adds the authoritative **post-open fstat** `IsRegular` check (the read-side analogue of the post-open guard the JSON-collection walk paths already use), ordered before `refuseHardLink` so a non-regular target is classified as not-regular rather than hard-linked.

## Risk

Hardening only; tightens an untrusted-dump-input boundary in the offline encode tool. No change to the happy path (regular `.bin` files read identically).

## Self-review (5 lenses)

- **Data loss**: none — read-only offline tool; rejects bad input rather than dropping good data.
- **Concurrency/distributed**: closes a TOCTOU window (stale ReadDir type vs. authoritative post-open fstat) in a single-goroutine offline encoder.
- **Performance**: one extra `Mode()` comparison on an already-`Stat`'d fd; no new syscall.
- **Consistency**: matches the post-open `IsRegular` pattern already used by the JSON-collection walk and the DDB schema reader.
- **Test coverage**: new co-located regression test calls `readRootFile` on a directory (cross-platform non-regular stand-in) and asserts `ErrRedisEncodeNotRegular`; confirmed failing without the guard (returned the hard-link error) and passing with it.

## Test plan

- [x] Regression test fails without the guard, passes with it (verified by temporarily reverting the guard)

### DIFF
--- a/internal/backup/encode_redis.go
+++ b/internal/backup/encode_redis.go
@@ -367,7 +367,7 @@ func readRootFile(root *os.Root, name string) ([]byte, error) {
 	// stale, so a FIFO/device swapped in between ReadDir and Open
 	// would otherwise reach io.ReadAll (blocking, or reading
 	// attacker-controlled bytes from a writer-attached FIFO). The
-	// post-open fstat is authoritative (claude review on PR #831).
+	// post-open fstat is authoritative; the ReadDir type is not.
 	if !info.Mode().IsRegular() {
 		return nil, errors.Wrapf(ErrRedisEncodeNotRegular, "%s (mode=%s)", name, info.Mode())
 	}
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
@claude review

PR #829, opened 2026-05-25
docs(keyviz): propose sub-range hot-key sampling (Phase 2-A+)

bootjp:
## What

Design proposal (docs-only) for **sub-range / hot-key sampling** in the Key Visualizer.

Today the sampler keys every counter on `RouteID`, so the heatmap's Y axis is route-granular. A broad route (common before any hotspot split) collapses all its traffic onto **one** row, hiding hot keys *inside* it — the opposite of what a BigTable-style Key Visualizer is meant to show. The hot path already has the key in scope at every `Observe` call site (`mut.Key` for writes; `key` for reads) but discards it.

This doc proposes subdividing each **individual** route's `[Start, End)` into `K` **order-preserving** sub-range buckets (even division over a fixed byte window past the `Start`/`End` common prefix), so a hot key/sub-range renders as a bright cell. Sub-bucket metadata is precomputed and immutable after `RegisterRoute`, keeping the hot path **lockless and allocation-free**.

## Scope

- **Sampler + JSON only** — no proto / gRPC / Raft / FSM change (the finer rows are ordinary `MatrixRow`s; the SPA already keys rows by `Start`/`End`, so no structural SPA change).
- Gated behind `--keyvizKeyBucketsPerRoute` (**default 1 == today's exact route-level behaviour**), so the change is inert until opted into.
- Non-goals: exact Top-K heavy-hitter ranking (deferred as a future drill-down), sub-dividing virtual aggregate buckets, adapter-direct read sampling, persistence.

## Why design-doc-first

Per CLAUDE.md, this touches the sampler hot path and the wire shape, so the proposal lands before implementation. The doc is reviewable on its own. Implementation (signature change to `Observe`, per-route sub-bucket counters, `Flush` emitting sub-rows, flag, tests/bench) follows once accepted.

## Open questions (please weigh in)

1. **Unbounded-tail bucketing** — high-order-bits-over-a-window fallback for the last route's open `End`, vs keep it a single row until a split gives it a finite `End`.
2. **Mixed-`K` in fan-out** — should tolerate peers with different `K` (rows keyed by `BucketID` now embedding `#subIdx`); add an explicit test.
3. **Eager vs lazy** sub-bucket allocation — eager keeps the hot path simple at a bounded ~5 MB worst case; lazy saves idle-route memory at the cost of hot-path branches.
4. **Window width `W = 8`** — fixed constant vs configurable.

## Self-review

[...]

PR #836, opened 2026-05-26
feat(keyviz): per-route sub-range (hot-key) sampling (Phase 2-A+)

bootjp:
## What

Implements the merged design `docs/design/2026_05_25_proposed_keyviz_subrange_sampling.md` (#829): **per-route sub-range (hot-key) sampling**. Each individual route's `[Start, End)` is divided into `K` **order-preserving** sub-buckets, so the Key Visualizer surfaces hot keys / hot sub-ranges (BigTable-style) instead of collapsing a whole route to a single heatmap row.

Opt-in via `--keyvizKeyBucketsPerRoute` (**default 1 = today's exact route-level behaviour**, cap 256). Inert until raised.

## How

- **Sampler** (`keyviz/sampler.go`): `Observe(routeID, key, op, valueLen)` now takes the key and buckets it via `subBucketIndex`. Each `routeSlot` carries an **immutable** sub-layout (`subPrefixLen/subStart/subEnd/subSpan` + `subLo/subHi` bound snapshots + `subBuckets []subCounter`), set once at register and never recomputed — so the hot path reads it lock-free. Bucketing = even division over an 8-byte window past the Start/End common prefix, exact 128-bit math (`bits.Mul64/Div64`), with a **full-key edge clamp** so it is globally order-preserving even for out-of-range keys. `Flush` emits one row per non-empty sub-bucket, bounds tiling `[Start,End)` exactly (bucket 0 → route Start, last → route End). New `MatrixRow.SubBucket/SubBucketCount`.
- **Pivot** (`internal/admin/keyviz_handler.go` + `adapter/admin_grpc.go`): re-keyed from bare `RouteID` to `BucketID` (embeds `#subIdx` when `SubBucketCount>1`) so a route's sub-rows don't collide. K=1 keeps the exact legacy `route:<id>` id. **No proto change** — `BucketId` already distinguishes the sub-rows.
- **Coordinator**: `observeMutation`/`observeRead` forward the full key (both call sites already had it in scope).

## Performance

`BenchmarkObserveSubRange` (hot path):

| K | ns/op | allocs/op |
|---|---|---|
| 1 (default) | 8.7 | 0 |
| 64 | 18.4 | 0 |

Allocation-free at both; the K=64 cost is one `windowUint64` + `Mul64`/`Div64` + two `bytes.Compare` clamps. Memory bounded `K×32 B/route`; cap 256 → ~80 MB worst case at the default `MaxTrackedRoutes`.

## Caller audit (semantic-changing signature)

`Observe` dropped `keyLen` for `key []byte`. Non-test callers grep-audited: `observeMutation` (`mut.Key`) and `observeRead` — which itself changed signature; both its call sites (`LinearizableReadForKey`, `LeaseReadForKey`) already pass the full key. `recordingSampler` / benchmark / `sampler_test` updated. No caller treats the argument as anything but the observed key.

1. **Data loss** — read-only sampling path; no Raft/FSM/Pebble/TTL. Sub-bucketing only re-attributes already-counted traffic. Grace-window reclaim **preserves** counters (never zeroed/resized) — pinned by `TestSamplerGraceWindowReRegistrationKeepsStaleLayout`. Existing retired-slot/grace tests green.
2. **Concurrency / distributed** — hot path lock-free: sub-layout (incl. `subLo/subHi`) immutable for the slot's lifetime, read without `metaMu`; counters per-sub-bucket atomics; `Flush` drains with `atomic.Swap`. Reclaim does not recompute the layout, so there is no writer to race the reader. Aggregates stay single-bucket (no new `metaMu`). `go test -race ./keyviz/...` green.
3. **Performance** — see table; allocation-free, no extra Raft round-trips, HLC untouched. Flush cost rises with non-empty sub-buckets (off the request path).
4. **Data consistency** — sub-rows carry the parent `(RaftGroupID, LeaderTerm)` unchanged, so fan-out dedupe + per-cell conflict operate identically. Bucketing deterministic + monotone (rapid property). Pivot re-key prevents sub-row collisions (admin + adapter regression tests). Mixed-K peers coexist (not merged). K=1 BucketID byte-identical to today.
5. **Test coverage** — `computeSubLayout` table; `subBucketIndex` edges + single-bucket + **rapid order-preservation property** (which caught a real out-of-range monotonicity bug, fixed via the full-key clamp); hot-key-on-distinct-row e2e; exact tiling; grace-window stale-layout + post-grace fresh; K=1 regression; pivot-collision (admin + adapter); mixed-K fan-out coexist; K=1-vs-K=64 benchmark.

## Test evidence

`go build ./...`, `go vet ./...`, `golangci-lint` (0 issues), `go test ./keyviz/... ./internal/admin/... ./kv/...` and scoped `./adapter/` keyviz tests all green; `go test -race ./keyviz/...` green. The full `./adapter/...` suite (S3/SQS/Dynamo/Redis integration) exceeds the local 10-min sandbox budget — left to CI.

## Implementation notes vs the proposed doc (for the doc follow-up)

Two refinements surfaced during implementation; the proposed→implemented doc rename will fold these in plus the round-5 clarity edits that landed after #829 merged:

1. **Full-key edge clamp.** The window only sees bytes past the common prefix, so window-only bucketing isn't order-preserving for keys *outside* `[Start,End)` (the rapid property found this). Added immutable `subLo/subHi` bound snapshots and clamp on the full key first. In-range keys (the only ones Observe sees in production) are unaffected.
2. **Degenerate case is empty-range-only.** Because `commonPrefixLen` lands exactly on the first differing byte, the window always captures a real difference for `Start < End`; the "window too narrow" degenerate case in §3.2 is therefore unreachable in practice — only `Start == End` collapses to a single bucket. The guard remains as correct defensive code.

### DIFF
--- a/adapter/admin_grpc.go
+++ b/adapter/admin_grpc.go
@@ -684,10 +684,18 @@ func rowActivityTotal(r *pb.KeyVizRow) uint64 {
 // to trust route_count for drill-down decisions.
 func newKeyVizRowFrom(mr keyviz.MatrixRow, numCols int) *pb.KeyVizRow {
 	total := mr.MemberRoutesTotal
-	if !mr.Aggregate && total == 0 {
+	switch {
+	case !mr.Aggregate && total == 0:
 		// Individual slots fall through to RouteCount=1 when the
 		// sampler predates MemberRoutesTotal or never set it.
 		total = 1
+	case mr.Aggregate && total == 0:
+		// Defensive fallback mirroring the JSON pivot
+		// (keyviz_handler.go): a just-coalesced virtual bucket
+		// serialized before its count is set would otherwise render
+		// "0 routes", which is nonsense for an aggregate. Fall back to
+		// the visible member count so route_count stays meaningful.
+		total = uint64(len(mr.MemberRoutes))
 	}
 	row := &pb.KeyVizRow{
 		BucketId:          bucketIDFor(mr),
--- a/adapter/admin_grpc.go
+++ b/adapter/admin_grpc.go
@@ -600,16 +600,23 @@ func matrixToProto(cols []keyviz.MatrixColumn, pick func(keyviz.MatrixRow) uint6
 	resp := &pb.GetKeyVizMatrixResponse{
 		ColumnUnixMs: make([]int64, len(cols)),
 	}
-	rowsByID := make(map[uint64]*pb.KeyVizRow)
-	order := make([]uint64, 0)
+	// Keyed by BucketId, not bare RouteID: sub-range bucketing gives a
+	// route's sub-rows the same RouteID, so keying on RouteID would
+	// collide them. BucketId embeds the sub-bucket index when
+	// SubBucketCount > 1, so it is unique per (route, sub-bucket). The
+	// proto KeyVizRow needs no new field — the BucketId distinguishes
+	// sub-rows. See design §5.1.
+	rowsByID := make(map[string]*pb.KeyVizRow)
+	order := make([]string, 0)
 	for j, col := range cols {
 		resp.ColumnUnixMs[j] = col.At.UnixMilli()
 		for _, mr := range col.Rows {
-			pr, ok := rowsByID[mr.RouteID]
+			id := bucketIDFor(mr)
+			pr, ok := rowsByID[id]
 			if !ok {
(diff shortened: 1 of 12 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

PR #833, opened 2026-05-25
backup: Phase 0b M3a — DynamoDB schema reverse encoder

bootjp:
## Summary

**Phase 0b M3a — DynamoDB schema reverse encoder.** First slice of the DynamoDB reverse encoder (M3): table **schemas**. Reconstructs the `!ddb|meta|table|` schema record + `!ddb|meta|gen|` generation counter from each table's `_schema.json`. Item records (`!ddb|item|`, needing the ordered primary-key encoding) and derived GSI rows land in follow-up PRs on this milestone, matching how Redis (M2) shipped incrementally.

Builds on M1's `snapshotBuilder`. Design: `docs/design/2026_05_25_proposed_snapshot_logical_encoder.md`.

`internal/backup/encode_dynamodb.go`:
- **`publicToSchema`** — inverse of `dynamodb.go`'s `schemaToPublic`: rebuilds `pb.DynamoTableSchema` (TableName, PrimaryKey, AttributeDefinitions map, GSIs) from `_schema.json`. Stamps `KeyEncodingVersion=V2` (ordered) so the restored table reads items with the encoding the item encoder will produce.
- **Schema value** = `storedDDBSchemaMagic` + deterministic proto marshal; key = `!ddb|meta|table|<base64url(name)>`. Generation counter = `!ddb|meta|gen|<base64url(name)>` = decimal.
- Reads `_schema.json` through the same `os.Root` + post-open `IsRegular` + `refuseHardLink` + `decodeOneJSON` (trailing-data) + `format_version` gate the Redis encoders use.

## Decision gate: generation (design §"DynamoDB")

`_schema.json` carries no original generation, so the encoder stamps a uniform `ddbRestoreGeneration` into **both** the schema proto's `Generation` field **and** the `!ddb|meta|gen|` counter — and the item encoder (next PR) will embed the **same** generation in every `!ddb|item|` key, so the counter and item-key generation always agree. This is the **Option-B** path: internally consistent for a single-cluster restore; the original generation number is unrecoverable from the public dump. (Option A — adding a `generation` field to the decoder's `_schema.json` — remains available if exact-fidelity is later required.)

## Risk

New code only; no existing behavior changed. Offline-tool boundary preserved (proto/key formats reproduced, not imported from the live adapter).

## Self-review (5 lenses)

- **Data loss**: gold-standard directory round-trip via the **real** `DecodeSnapshot` confirms schema fidelity; fail-closed on bad `format_version` / non-regular file / hard-link.
[...]

PR #837, opened 2026-05-26
backup: Phase 0b M3b-1 — DynamoDB base-item reverse encoder (no GSI)

bootjp:
## Summary

**Phase 0b M3b-1 — DynamoDB base-item reverse encoder.** Second slice of the DynamoDB reverse encoder (M3): table **items**. Walks each table's `items/<pk>[/<sk>].json` and stages the `!ddb|item|<base64url(table)>|<gen>|<orderedHash><orderedRange>` records (value = `storedDDBItemMagic{0x00,'D','I',0x01}` + deterministic `DynamoItem` proto).

> **Stacked on #833 (M3a, schema encoder).** Base branch is `backup/snapshot-encode-dynamodb`; the diff is M3b-only. Will retarget to `main` once #833 merges.

Builds on M3a's `DynamoDBEncoder`. Design: `docs/design/2026_05_25_proposed_snapshot_logical_encoder.md`.

### Key fidelity
Pinned against the **live adapter** write path (`dynamoItemKey` / `encodeDynamoKeySegment`), **not** the legacy base64url-segment `EncodeDDBItemKey` test helper. The restored schema stamps `KeyEncodingVersion=V2`, so item keys use the **ordered** encoding (escape `0x00`→`0x00 0xFF`, `0x00 0x01` terminator per segment). A hand-computed key-layout test guards this — necessary because the decoder round-trip reconstructs items from the proto **value** and never validates the key's ordered payload.

- `publicToItem` / `publicToAttributeValue`: inverse of `itemToPublic`, covering all 10 attribute types (S, N, B, BOOL, NULL, SS, NS, BS, L, M). Dispatch split into scalar/collection switches (no var init-cycle, complexity well under the limit).
- Generation: every item key embeds `ddbRestoreGeneration` — the same uniform value M3a's schema proto and `!ddb|meta|gen|` counter carry — so the restored counter and item keys always agree.
- Item files read through the same `os.Root` + `IsRegular` + `refuseHardLink` + `decodeOneJSON` guards as the schema reader.

## Scope / deferred (fail-closed, not silent)
- **Numeric (N) keys** fail closed with `ErrDDBEncodeNumericKeyUnsupported` — the ordered numeric encoding (`encodeNumericKeyBytes`) is a follow-up slice (**M3b-2**). S and B keys work now.
- **Derived `!ddb|gsi|` rows** are **not emitted yet** (**M3b-3**). A table restored from this slice has its base items but no secondary-index rows; GSI queries would need a post-restore rebuild until M3b-3. (The decoder already drops GSI rows as a no-op, so the encode→decode round-trip is unaffected.)

## Risk
New code + one wiring line in `encodeTable`. No existing behavior changed. Offline-tool boundary preserved (key/proto formats reproduced, not imported from the live adapter).

## Self-review (5 lenses)
- **Data loss**: every item file read; round-trip via the **real** `DecodeSnapshot` confirms attribute fidelity; fail-closed on bad JSON / numeric key / missing key / non-regular file. Decoder's items layout is a fixed 2 levels and the walk matches it.
- **Concurrency/distributed**: single-goroutine, one builder.
- **Performance**: one streamed read + preallocated key buffers per item; no consensus (offline).
- **Consistency**: ordered key + magic + proto reproduce the live adapter exactly (key-layout test); generation coherent with M3a schema/counter; `KeyEncodingVersion=V2` matches the ordered key.
- **Test coverage**: composite (all 10 attr types), hash-only, binary key with `0x00` escape in a subdir (descent), full key-bytes assertion, numeric fail-closed, missing-hash-key fail-closed, missing-items-dir no-op, ordered-segment unit.

## Test plan
- [ ] Follow-up: M3b-2 numeric-key ordered encoding; M3b-3 GSI derivation.

### DIFF
--- a/internal/backup/encode_dynamodb.go
+++ b/internal/backup/encode_dynamodb.go
@@ -138,7 +138,10 @@ func (e *DynamoDBEncoder) encodeTable(b *snapshotBuilder, root *os.Root, tableDi
 
 	genKey := append([]byte(DDBTableGenPrefix), encTable...)
 	genVal := []byte(strconv.FormatUint(ddbRestoreGeneration, 10))
-	return b.Add(genKey, genVal, 0)
+	if err := b.Add(genKey, genVal, 0); err != nil {
+		return err
+	}
+	return e.encodeItems(b, root, tableDir, tableName, schema)
 }
 
 // readSchema opens <table>/_schema.json within root (symlink-escape /
--- a/internal/backup/encode_dynamodb_items.go
+++ b/internal/backup/encode_dynamodb_items.go
@@ -395,7 +395,13 @@ func decodeAVNull(v any) (*pb.DynamoAttributeValue, error) {
 	if !ok {
 		return nil, errors.Wrap(ErrDDBEncodeInvalidItem, "NULL value is not a boolean")
 	}
-	return &pb.DynamoAttributeValue{Value: &pb.DynamoAttributeValue_NullValue{NullValue: b}}, nil
+	// AWS DynamoDB's NULL is definitionally {"NULL": true}; {"NULL": false}
+	// is invalid and unreachable from real data, so fail closed rather
+	// than emit a record the live API would reject on restore.
+	if !b {
+		return nil, errors.Wrap(ErrDDBEncodeInvalidItem, "NULL value must be true")
+	}
+	return &pb.DynamoAttributeValue{Value: &pb.DynamoAttributeValue_NullValue{NullValue: true}}, nil
 }
 
 func decodeAVStringSet(v any) (*pb.DynamoAttributeValue, error) {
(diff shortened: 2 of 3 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
**Conflict resolution (gemini HIGH ↔ codex P1/P2).** gemini flagged that empty `SS`/`NS`/`BS` and `{"NULL": false}` should be rejected (AWS-API validity). codex (P1/P2) flagged that rejecting them breaks **invertibility**. codex is correct for this path:

- The decoder (`internal/backup/dynamodb.go`) *intentionally* serializes nil/empty sets as `[]` and the NULL boolean as-is, to preserve data rather than drop attributes.
- The reverse encoder must round-trip those structural cases, or recovery of a snapshot with a legacy/drifted record fails.
- This path reconstructs **internal** records; AWS-API payload validity is enforced at the adapter's `PutItem` boundary, not here.

So I reverted the strict validation and added a round-trip regression test (`TestDDBEncodeItemPreservesStructuralEdgeCases`: empty SS/NS/BS + NULL=false survive encode → real decode). Net for #837: gemini's copy-paste/format items remain fixed; the empty-set/NULL guards are intentionally **not** applied, by the round-trip contract above.

@claude review

bootjp:
@claude review

PR #838, opened 2026-05-26
backup: Phase 0b M3b-2 — DynamoDB numeric (N) primary-key ordered encoding

bootjp:
## Summary

**Phase 0b M3b-2 — DynamoDB numeric (N) primary-key ordered encoding.** Third slice of the DynamoDB reverse encoder (M3). `encode_dynamodb_numeric.go` reproduces the live adapter's order-preserving numeric key encoding (`adapter/dynamodb.go` `encodeNumericKeyBytes` + parse/normalize/ordered-int helpers): sign marker (`0x00` neg / `0x01` zero / `0x02` pos) + order-encoded decimal exponent + `0x00` separator + significant digits, with the body bit-inverted for negatives so larger magnitudes sort earlier.

> **Stacked on #833** (which now contains M3a schema + M3b-1 base items after #837 merged into it). Base branch is `backup/snapshot-encode-dynamodb`; the diff is M3b-2-only. Will retarget to `main` once #833 merges.

`ddbPrimaryKeyBytes` now returns these bytes for an `N` key; they flow through the **same** `encodeDDBOrderedKeySegment` escape/terminator as S/B — exactly as the live `attributeValueAsKeySegment` wraps `attributeValueAsKeyBytes`. This removes the M3b-1 fail-closed stub (`ErrDDBEncodeNumericKeyUnsupported` deleted); malformed number literals fail closed via `ErrDDBEncodeInvalidItem`.

## Why the order-preserving property test matters
The decoder round-trip reconstructs items from the proto **value** and never validates the ordered key payload, so a value round-trip can't catch a mis-encoded numeric key. `TestDDBNumericKeyOrderPreserving` asserts the encoded segments sort byte-lexicographically (the comparison Pebble uses) in numeric order across negative / zero / positive / fractional / exponent values — including `1000 == 1e3` encoding **equal** — which is the actual correctness guarantee the encoding exists to provide.

## Risk
New file + one changed branch in `ddbPrimaryKeyBytes`. No change to S/B/value handling. Offline-tool boundary preserved (numeric encoding reproduced, not imported).

## Self-review (5 lenses)
- **Data loss**: numeric keys now reconstructed; round-trip verified; malformed numbers fail closed.
- **Concurrency/distributed**: single-goroutine.
- **Performance**: O(digits) per key; no hot path (offline tool).
- **Consistency**: faithful reproduction verified by the order-preserving property + canonical zero layout; N bytes wrapped identically to S/B; generation/key format unchanged.
- **Test coverage**: numeric hash round-trip, composite S/N round-trip (negative + fractional), order-preserving property, zero layout (`0x01`), malformed-literal rejection.

## Caller audit (semantic change)
`ddbPrimaryKeyBytes`'s N case went from fail-closed error → success. Its sole caller `ddbItemKeyBytes` applies `encodeDDBOrderedKeySegment` uniformly to the returned bytes for both hash and range positions, so the new success path is handled identically to S/B. `ErrDDBEncodeNumericKeyUnsupported` had no other callers (grep).

## Test plan
- [ ] Follow-up: M3b-3 GSI row derivation.

### DIFF
--- a/internal/backup/encode_dynamodb_items.go
+++ b/internal/backup/encode_dynamodb_items.go
@@ -13,20 +13,15 @@ import (
 )
 
 // encode_dynamodb_items.go is the Phase 0b DynamoDB ITEM reverse encoder
-// (milestone M3b-1) — the inverse of the per-item JSON emitter in
-// dynamodb.go (writeDDBItem / itemToPublic). It reconstructs the
+// — the inverse of the per-item JSON emitter in dynamodb.go (writeDDBItem
+// / itemToPublic). It reconstructs the
 // !ddb|item|<base64url(table)>|<gen>|<orderedHash><orderedRange> records
 // from each table's items/<pk>[/<sk>].json files.
 //
-// Scope of this slice (M3b-1):
-//   - S (string) and B (binary) primary/range keys only. Numeric (N)
-//     keys take a separate ordered encoding (encodeNumericKeyBytes in the
-//     live adapter) that is reproduced in a follow-up slice; until then a
-//     numeric key fails closed (ErrDDBEncodeNumericKeyUnsupported) rather
-//     than emitting a wrongly-ordered key the live store would mis-sort.
-//   - Base items only. The derived !ddb|gsi| index rows (which the
-//     decoder drops as a no-op because they are derivable from the base
-//     item set + schema) land in a later slice.
+// Primary keys: S (string), B (binary), and N (number) are all supported;
+// the N ordered encoding lives in encode_dynamodb_numeric.go. The derived
+// !ddb|gsi| index rows (which the decoder drops as a no-op because they
+// are derivable from the base item set + schema) land in a later slice.
 //
 // Format fidelity is pinned against the live adapter write path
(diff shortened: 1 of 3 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
Thanks — all four prior observations are addressed.

On the remaining non-blocking suggestion (`ddbSignedMagnitude`: `math/big` → bitwise abs): **keeping `big.Int` intentionally.**

- This is an offline, one-time-restore cold path — the per-key allocation has no measurable cost, so the optimization buys nothing here.
- The whole value of `encode_dynamodb_numeric.go` is being a faithful, line-by-line mirror of the live adapter's `signedMagnitude` / `encodeNumericKeyBytes`, so audits are a trivial diff (as this review did). Diverging this one function to a hand-rolled two's-complement negation would break that parity and is marginally less obviously correct for `math.MinInt64` (which the review confirmed `big.Int` handles correctly).

So leaving it as-is for parity + provable `MinInt64` safety; the `math/big` import stays used. No code change.

PR #840, opened 2026-05-26
backup: Phase 0b M3b-3 — DynamoDB GSI row derivation (completes M3)

bootjp:
## Summary

**Phase 0b M3b-3 — DynamoDB GSI row derivation.** Final slice of the DynamoDB reverse encoder (M3). `encode_dynamodb_gsi.go` is the inverse of the decoder's **no-op** `HandleGSIRow` (which drops GSI rows precisely because they are derivable from the base item set + schema). It reproduces the live `gsiEntryKeysForItem` (V2 ordered path):

```
key   = !ddb|gsi|<b64(table)>|<gen>|<b64(index)>|<orderedIndexHash><orderedIndexRange><orderedPKHash><orderedPKRange>
value = <the base item's !ddb|item| key>   (pointer back to the item)
```

> #833 (M3a+M3b-1+M3b-2) has merged to main; this PR now targets `main` directly and completes M3.

- **Sparse indexing**: an item missing the index hash attribute (or a defined index range attribute) contributes no row for that index — matching the live `gsiKeyValues` `include=false` path. Index names emitted in sorted order.
- **Refactor**: primary-key segment extraction is now `ddbOrderedPrimaryKeySegments`, shared by the item key and the GSI rows' embedded PK segments, so they are guaranteed identical. `encodeOneItem` stages the item record then its GSI rows.

## Why tests use `DecodeLiveEntries`
The decoder drops GSI rows, so an encode→decode directory round-trip can't observe them. Tests instead encode → write → `DecodeLiveEntries` and inspect the staged `!ddb|gsi|` records directly.

## Risk
New file + a behavior-preserving refactor of `ddbItemKeyBytes` + GSI-row staging in `encodeOneItem`. Offline-tool boundary preserved (GSI key/value format reproduced, not imported).

## Self-review (5 lenses)
- **Data loss**: GSI rows derivable; items fully restored; round-trip unaffected; sparse skip matches live.
- **Concurrency/distributed**: single-goroutine.
- **Performance**: O(GSIs) per item; offline tool.
- **Consistency**: key/value reproduce `gsiEntryKeysForItem` exactly (hand-computed layout test + value == item key + sparse + sorted index names); PK segments shared with the item key; generation = `ddbRestoreGeneration` throughout.
- **Test coverage**: GSI row points to item (+ prefix), sparse skip, full key-bytes layout, no-GSI no-rows; existing item/numeric round-trips still pass.

## Caller audit
`ddbItemKeyBytes` refactored to call `ddbOrderedPrimaryKeySegments` — behavior-preserving (existing key-layout + round-trip tests pass, byte-identical keys). `encodeOneItem` gained GSI-row staging after the item record; no existing return semantics changed. `ddbGSIRows` is new. GSI keys embed the full PK so they cannot collide with item keys or each other (`b.Add` dup-detection would otherwise fail closed).

## Test plan

## Summary by CodeRabbit

* **Chores**
  * Enhanced DynamoDB backup handling to properly encode and capture Global Secondary Index (GSI) rows, including support for sparse indexing (when index keys are missing) and multiple indexes per table, ensuring secondary index entries are accurately preserved and referenced with primary items in backups.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/840?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/backup/encode_dynamodb_gsi.go
+++ b/internal/backup/encode_dynamodb_gsi.go
@@ -1,0 +1,115 @@
+package backup
+
+import (
+	"encoding/base64"
+	"sort"
+	"strconv"
+
+	pb "github.com/bootjp/elastickv/proto"
+)
+
+// encode_dynamodb_gsi.go derives the global-secondary-index rows for each
+// base item — the inverse of the decoder's no-op HandleGSIRow, which drops
+// GSI rows precisely because they are derivable from the base item set +
+// schema. It reproduces the live adapter's gsiEntryKeysForItem (V2 ordered
+// path): for every GSI whose key attributes the item carries, emit
+//
+//	key   = !ddb|gsi|<b64(table)>|<gen>|<b64(index)>|
+//	        <orderedIndexHash><orderedIndexRange><orderedPKHash><orderedPKRange>
+//	value = <the base item's !ddb|item| key>   (a pointer back to the item)
+//
+// Sparse indexing: an item missing the index hash attribute (or a defined
+// index range attribute) contributes no row for that index — matching the
+// live gsiKeyValues include=false path.
+
+// ddbGSIRow is one derived index row: the index key and the base item key
+// it points to.
+type ddbGSIRow struct {
+	key   []byte
+	value []byte
+}
+
+// ddbGSIRows derives the GSI rows for one item. itemKey is the item's
+// primary !ddb|item| key, stored as each row's value.
+func ddbGSIRows(tableName string, generation uint64, schema *pb.DynamoTableSchema, item *pb.DynamoItem, itemKey []byte) ([]ddbGSIRow, error) {
+	gsis := schema.GetGlobalSecondaryIndexes()
+	if len(gsis) == 0 {
+		return nil, nil
+	}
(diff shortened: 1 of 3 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
Re: the remaining optional nil-guard on `ddbOrderedPrimaryKeySegments` — considered, and leaving it out: it's **redundant**.

`encodeTable` already fails closed on an empty primary hash key **before** it ever calls `encodeItems`:

```go
if schema.GetPrimaryKey().GetHashKey() == "" {
    return errors.Wrapf(ErrDDBEncodeInvalidSchema, "%s/_schema.json: empty primary hash key", tableDir)
}
...
return e.encodeItems(b, root, tableDir, tableName, schema)
```

So by the time `ddbOrderedPrimaryKeySegments` (and `ddbGSIRows`, which also calls it) runs, the schema's hash-key **name** is guaranteed non-empty — the `item missing hash-key attribute ""` confusing-message case is unreachable from any real flow. Adding a second guard here would be dead defensive code.

The two test gaps are closed; everything else flagged is cosmetic/offline. No further changes — ready to merge.

PR #848, opened 2026-05-28
docs(encryption): Stage 7b proposed - runtime writer re-registration

bootjp:
## Summary
Stage 7b — runtime writer re-registration design doc. Addresses the two cache-driven runtime cases the codex **P1**s on PR #847 explicitly identified as the deferred 7a follow-on:

1. **Phase-0 boot → runtime `EnableStorageEnvelope`.** The envelope flips active on a node that skipped process-start registration; the storage gate now fail-closes direct encrypted writes until restart.
2. **Non-proposer node after runtime `RotateDEK`.** `applyRotateDEK` only inserts the *proposer's* registration row; every other node has `activeStorageDEKID` re-pointed to the new DEK and `Registered()` returns false for it.

Both produce the same trigger condition (gate would fail-closed). 7b adds a single runtime watcher goroutine that observes the shared `StateCache`, proposes a registration via the existing 7a `runWriterRegistration` helper for `(activeDEK, fullNodeID, w.epoch)`, and seeds `MarkRegistered` through the existing `releaseBarrier` on commit. **No new wire-format event.** **No fail-OPEN fallback** — design §2.3 of 7a-2 still applies; the gate stays fail-closed until the node actually registers.

## Key design decisions
- **Polling, not channel.** `StateCache` has no subscription API (atomic predicates only); polling at 1 s is O(2 atomic loads/tick), bounded by `runCtx`, no risk of dropped events.
- **`local_epoch` policy.** The registration uses `w.epoch` (the nonce factory's pinned value), NOT `sidecar.Keys[newDEK].LocalEpoch`. The nonce factory reuses the same `(node, epoch)` across DEK rotations and §4.1 nonce uniqueness is per-DEK, so this is safe and consistent with the on-disk nonces.
- **Idempotency vs. process-start.** If process-start and 7b race a propose, §4.1 case 2-idempotent makes the duplicate apply a no-op.
- **No new coordinator barrier.** 7a's per-load barrier closes once at startup; runtime cases only need the storage-layer `Registered()` to flip true.

## Out of scope
- 7c (ConfChange-time registration for a freshly-joining node) — distinct membership-driven mechanism.
- Replacing polling with a notification channel — possible follow-on.
- Rebuilding the nonce factory on rotation — deliberately not done; reusing the pinned epoch across DEKs is §4.1-safe.

## Test plan
- [x] Doc-only; lint clean
- [ ] Implementation + tests in the follow-up PR (fake-cache transitions for cutover and rotation; shutdown cancels cleanly; concurrent process-start + watcher propose is no-op via case 2-idempotent)

## Summary by CodeRabbit

* **Documentation**
  * Added design documentation for a new runtime writer re-registration mechanism, including implementation requirements, test specifications, and operational guidelines for specific system scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/848?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/design/2026_05_28_proposed_7b_runtime_reregistration.md
+++ b/docs/design/2026_05_28_proposed_7b_runtime_reregistration.md
@@ -152,9 +152,14 @@ gating on the deferred rotation branch. It runs for the lifetime of
    (§2.2).
 
 4. **Marks** on success via the existing `releaseBarrier(barrier,
-   cache, activeDEK)` helper (with a fresh per-attempt barrier whose
-   only consumer is the watcher). Once `MarkRegistered(activeDEK)`
-   stores, `cache.Registered()` returns true and the gate clears.
+   cache, activeDEK)` helper. The barrier is a fresh per-attempt
+   `chan struct{}` passed in solely to satisfy
+   `runWriterRegistration`'s call signature; **the watcher does NOT
+   `select` on it** — the coordinator's per-load barrier was closed by
+   process-start and is not re-opened by the watcher. The meaningful
+   side effect inside `releaseBarrier` is the `MarkRegistered(activeDEK)`
+   call. Once that stores, `cache.Registered()` returns true and the
+   gate clears.
 
 5. **Backs off** between attempts using the existing
    `registrationRetryInitial / registrationRetryMax /
@@ -365,10 +370,23 @@ already confirmed).
    crashed before `MarkRegistered` will be picked up by `verifyRegistered`
    on the next propose, short-circuiting before re-proposing.
 
-6. Add a unit test for the pre-bootstrap branch
-   (`bootDEKID == 0` → runtime `BootstrapEncryption` + cutover): the
-   watcher proposes for the freshly-installed DEK X at `w.epoch == 0`.
-   This pins the codex P1 round-2 fix.
+6. Add unit tests for the pre-bootstrap branch
+   (`bootDEKID == 0`), split per claude P2 round-3 to cover both
+   structurally distinct sub-paths:
+   - **6a — bootstrap-then-cutover.** `bootDEKID == 0` → runtime
+     `BootstrapEncryption` installs DEK X → `EnableStorageEnvelope`
+     fires while X is the active DEK → watcher proposes
+     `(X, node, 0)`. (`sidecar.Keys[X].LocalEpoch == 0 == w.epoch`.)
+     This pins the codex P1 round-2 fix.
+   - **6b — bootstrap-then-rotate-then-cutover.** `bootDEKID == 0` →
+     `BootstrapEncryption` installs DEK X → `RotateDEK` X→Y (before
+     first cutover) → `EnableStorageEnvelope` → `activeDEK == Y`,
+     `bootDEKID == 0` keeps the skip-condition false, watcher proposes
+     `(Y, node, 0)`. The invariant `sidecar.Keys[Y].LocalEpoch == 0 ==
+     w.epoch` still holds because `writeRotationSidecar` initialises
+     `LocalEpoch` to 0 for the newly-installed DEK and the live nonce
+     factory is still pinned at `w.epoch == 0` (no `BumpLocalEpoch`
+     ever ran on a pre-bootstrap load).
 
 ## 6. Deferred — rotation case (7b')
 
--- a/docs/design/2026_05_28_proposed_7b_runtime_reregistration.md
+++ b/docs/design/2026_05_28_proposed_7b_runtime_reregistration.md
@@ -1,4 +1,4 @@
-# Stage 7b — runtime writer re-registration (cutover + rotation cases)
+# Stage 7b — runtime writer re-registration (cutover case only; rotation deferred)
 
 | Field | Value |
 |---|---|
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

Stage 7b design doc — runtime writer re-registration. Addresses the two cache-driven runtime cases the codex P1s on PR #847 explicitly identified as the deferred 7a follow-on (Phase-0 boot → runtime cutover; non-proposer after RotateDEK).

Key design choices to review:
- Polling (1s) over a notification channel — keeps `StateCache` lock-free.
- `local_epoch` for the runtime propose = `w.epoch` (nonce-factory-pinned), not the freshly-rotated DEK's sidecar epoch.
- Reuses 7a's `runWriterRegistration` + `releaseBarrier` — no new propose pathway, no fail-OPEN fallback.
- Out of scope: 7c (ConfChange-time registration), notification-channel evolution, rebuilding the nonce factory on rotation.

bootjp:
@claude review

Round-1 (593e9cae) — addressed both reviewers' findings:

**codex P1 + gemini medium #2** (both line 125, same finding): the original `local_epoch = w.epoch` policy works for cutover (active DEK unchanged; `Keys[activeDEK].LocalEpoch == w.epoch` post-BumpLocalEpoch) but **breaks for rotation** (Keys[newDEK]=0; registering at w.epoch advances registry past sidecar → brick on next restart, or nonce reuse if §9.1 guard were bypassed).

**Fix:** narrow 7b scope explicitly to **cutover-only**; defer rotation to **7b'** with §6 listing three architecture options (per-DEK nonce factory, per-node sidecar coordination, or proposer-pinned rotation epoch). The watcher takes a deferred-skip branch (log once, no propose) when `activeStorageDEKID` has changed from the boot DEK.

**gemini medium #1** (line 90, concurrency): §2.1 now spells out the synchronous single-goroutine model — propose runs *in* the watcher goroutine (at most one in-flight propose; no accumulation, no duplicate Raft proposals from the watcher); `registrationAttemptTimeout` (5s) bounds sub-attempts. A mid-propose rotation is benign because the next loop body takes the deferred-skip branch.

Doc-only; no code change. (codex inline reply 404d because it's a file-level comment.)

bootjp:
@claude review

Round-2 (869da9a7) — addressed codex P1 round-2 + claude P2s and minors:

**codex P1 round-2** (line 108): the `activeDEK == bootDEK` scope check broke the pre-bootstrap → runtime `BootstrapEncryption` + `EnableStorageEnvelope` path. A node booting before any bootstrap has `bootDEK=0`, `w.epoch=0`; the runtime DEK X arrives with `Keys[X].LocalEpoch=0` matching the live nonce factory — NOT a rotation. **Fix:** refined the in-scope condition to `activeDEK == bootDEKID || bootDEKID == 0`. §0 case 1 now explicitly lists both sub-branches (Phase-0 + pre-bootstrap); §2.2 walks the cross-restart invariant for each.

**claude P2** ("log once" mechanism): specified as a `lastLoggedSkipDEK uint32` field on the watcher; deferred-skip WARN fires only when `activeDEK != lastLoggedSkipDEK`.
**claude P2** (`connCache` lifetime): decided in §5 — goroutine-scoped per `runWriterRegistration` call, matching 7as pattern.
**claude P2** (`verifyRegistered` callback): specified registry-read matching 7a, for crash-between-propose-and-MarkRegistered safety.
**claude minors:** `bootDEKID` is captured at watcher construction time (made explicit in §2.1); §6 option (b) wording fixed ("called from the apply path", not "exposed to the watcher" — 7bs watcher is cutover-only).

Doc-only; lint clean. (codex inline reply 404d — file-level comment.)

bootjp:
@claude review

Round-3 (1621f7d0): addressed the two non-blocking items from your round-3 verdict ("design is correct and ready for the implementation PR"):

- **P2 (test coverage split):** §5 item 6 now split into 6a (bootstrap-then-cutover, watcher proposes `(X, node, 0)`) and 6b (bootstrap-then-rotate-then-cutover, watcher proposes `(Y, node, 0)`; `bootDEKID==0` keeps the skip-condition false, and `sidecar.Keys[Y].LocalEpoch == 0 == w.epoch` because `writeRotationSidecar` initialises `LocalEpoch` to 0 and no `BumpLocalEpoch` runs on a pre-bootstrap load).
- **Minor (barrier artifact):** §2.1 step 4 now explicitly states the watcher does NOT `select` on the barrier; the barrier exists only to satisfy `runWriterRegistration`'s signature; only the `MarkRegistered` side effect matters.

Doc-only; lint clean.

bootjp:
@claude review

Round-4 (a40f5996) — addressed codex **P1** (line 367) + your matching round-4 minor:

The original §5 item 5 spec ("returns true when an existing row's LastSeenLocalEpoch >= w.epoch") was ambiguous enough that an implementer following 7a's `registryLastSeen(uint16)` helper pattern — which swallows `ok=false` and returns 0 — would have produced a fail-OPEN watcher in the pre-bootstrap branch (`w.epoch==0` matches the helper's zero return, `MarkRegistered` fires without proposing, gate opens with no durable row, next restart hits §9.1's missing-row-post-cutover guard).

§5 item 5 now requires the callback to gate explicitly on `ok=true` from `GetRegistryRow` BEFORE comparing the epoch, with an embedded Go snippet showing the exact shape. Also notes that the existing `registryLastSeen` helper MUST NOT be reused directly; an implementation may inline the body or introduce a new `registryLastSeenWithExistence` helper used by both 7a and 7b.

7a's process-start was incidentally safe (the not-bootstrapped guard short-circuits before reaching the propose path with `w.epoch==0`); 7b's watcher reaches that state via the pre-bootstrap branch, so the explicit gating is load-bearing.

Doc-only; lint clean.

PR #842, opened 2026-05-26
backup: Phase 0b M4-1 — S3 bucket-meta reverse encoder

bootjp:
## Summary

**Phase 0b M4-1 — S3 bucket-meta reverse encoder.** First slice of the S3 reverse encoder (M4): bucket **metadata**. Reconstructs the `!s3|bucket|meta|` record (JSON `s3LiveBucketMeta`) and the `!s3|bucket|gen|` generation counter (8-byte big-endian, reproducing `encodeS3Generation`) from each bucket's `_bucket.json`. Object bodies (re-chunked into `!s3|blob|` + `!s3|obj|head|` manifests) and multipart uploads land in follow-up slices (M4-2, M4-3).

Design: `docs/design/2026_05_25_proposed_snapshot_logical_encoder.md` §"S3". Targets `main` (independent of the DynamoDB M3 work).

**Two commits** (design-doc-first):
1. **Doc reconciliation** — the design said "Option A chosen" for generation counters, but M3 (DynamoDB) shipped **Option B** (uniform `ddbRestoreGeneration = 1`). Reconciled: Option B is the implemented cross-adapter decision (M4 uses `s3RestoreGeneration = 1`); Option A retained as the exact-fidelity upgrade path. **Decided with the maintainer.**
2. **Implementation** — `encode_s3.go`.

### Details
- `S3RecordEncoder` walks `s3/<bucket>/` via `os.Root` and reads each `_bucket.json` through the same `Lstat` + `IsRegular` + `refuseHardLink` + `decodeOneJSON` guards the DynamoDB schema reader uses. Bucket name comes from the JSON's authoritative `Name`; keys built with the reused `s3keys.BucketMetaKey` / `BucketGenerationKey`.
- Generation: uniform `s3RestoreGeneration` (Option B). `created_at_hlc` is not represented in the dump and restores as `0` (documented lossy field).
- Named `S3RecordEncoder` to avoid colliding with the decoder's `S3Encoder` in `s3.go`.

## Risk
New code only + a design-doc edit. No existing behavior changed. Offline-tool boundary preserved (key/value formats reproduced; only the in-package `s3keys` constructors are reused).

## Self-review (5 lenses)
- **Data loss**: round-trip via the **real** `DecodeSnapshot` confirms bucket fidelity; fail-closed on non-regular file / empty name / bad JSON.
- **Concurrency/distributed**: single-goroutine, one builder.
- **Performance**: streams each `_bucket.json`.
[...]

PR #849, opened 2026-05-28
backup: fail-closed on user S3 key collision with reserved dump dirs (#842 P1 follow-up)

bootjp:
## Summary
- Codex P1 (#842 inline) flagged that `walkObjectEntry` silently dropped any user S3 object whose key prefix collided with the reserved `_incomplete_uploads/` or `_orphans/` dump directories.
- The decoder's `resolveObjectFilename` does not rename such keys — they land at their natural path. The encoder was skipping the entire subtree unconditionally → data loss on restore.
- Disambiguate by sidecar presence (reserved-dir payload has none; real objects always do) and fail closed via `ErrS3EncodeReservedPrefixCollision` when a sidecar is found inside.

## Self-review (5 passes)
1. **Data loss** — the regression itself is data loss; the fix turns it into a fail-closed error.
2. **Concurrency / distributed failures** — pure local-walk change, no Raft / lock / leader interaction.
3. **Performance** — adds at most one recursive walk per top-level reserved dir per encode; bounded by the dump tree.
4. **Data consistency** — caller chain (`encodeBucket → encodeBucketObjects → walkObjects → walkObjectEntry → walkObjectSubdir`) propagates errors uniformly with the existing `ErrS3EncodeUnsupportedCollision` / `ErrS3EncodeInvalidManifest` failures; semantics match end-to-end (audited).
5. **Test coverage** — new `TestS3EncodeRejectsUserObjectUnderReservedDir` (table-driven over both reserved names, would have failed against pre-fix code) and `TestS3EncodeSkipsLegitimateReservedDumpDirs` (pins the benign-skip path so future tweaks don't over-trigger).

## Test plan
- [x] `go test -race ./internal/backup/...` — passes
- [x] `golangci-lint run ./internal/backup/...` — 0 issues
- [x] New tests would have failed against the pre-fix code (the failing scenario hits `walkObjectEntry`'s old unconditional `return nil`)

## Summary by CodeRabbit

* **Bug Fixes**
  * S3 backup processing now skips reserved dump-control directories only when they contain no metadata sidecar files; if a user object is found under a reserved prefix with sidecars, the backup fails explicitly and instructs renaming the colliding key to avoid silent drops.

* **Tests**
  * Added tests covering collision detection with reserved directories and correct handling of legitimate system payloads.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/849?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
Applied the uniform `errors.WithStack(err)` wrap on `reservedDirHoldsSidecar` top-level error path (PR comment [4564303804](https://github.com/bootjp/elastickv/pull/849#issuecomment-4564303804) micro-observation). Marked non-blocking by claude bot but applied for code uniformity with the recursive path (already wrapped in 93263ce) and `rootEntryExists`.

Semantics audit: `reservedDirHoldsSidecar` returns `(bool, error)`; only caller is `walkObjectSubdir` which propagates the error directly. No semantic change — only stack-trace metadata added.

@claude review

PR #850, opened 2026-05-28
docs(keyviz): propose per-cell hot-key drill-down (Top-K, Phase 2-A++)

bootjp:
## What

Proposes the **per-cell hot-key drill-down** (Top-K), the deferred follow-up explicitly flagged in the sub-range design (§2.2 / §8 of `2026_05_25_proposed_keyviz_subrange_sampling.md`). On a live single-route deploy (`route:1 = [nil,nil)`), even K=64 sub-ranges are leading-byte-coarse — a hot heatmap cell still doesn't name the hot key. This adds an on-click endpoint that returns the actual hottest keys for the clicked cell.

## How (high level)

- **Algorithm**: per-route Space-Saving sketch (`m = 64`); aggregates are not tracked (their RouteIDs are synthetic and span many real routes).
- **Hot path**: PRNG-gated 1-in-R sampling (default `R = 16`), pooled key clone, **non-blocking single bounded channel** send (`select` with `default` → drop on full). **No mutex, no map mutation, no allocation on the disabled or unsampled paths.** Disabled-case overhead target: ≤ 1 ns over baseline.
- **Background aggregator**: one goroutine drains the channel and updates per-route sketches single-threaded (no lock contention on update). It publishes immutable snapshots via `atomic.Pointer.Store` per route; drill-down reads lock-free.
- **API**: `GET /admin/api/v1/keyviz/hotkeys?bucket_id=route:1#3&series=writes&top=20` — for `route:<id>#<subIdx>` the server resolves the sub-range and filters the route's top-`m` by that range. **Per-route sketches + range filter approximate per-cell top-K at a fraction of per-sub-bucket memory** (`K_buckets × m × routes` would otherwise blow up).
- **Fan-out**: existing keyviz fan-out infra. Merge = sum counts per key across nodes, take global top-K. Mixed-feature peers reuse the existing degraded-node status.

## Why this design

- The drill-down's UX target is "the operator clicked a hot cell, what keys made it hot?" Those keys almost always appear in the route's top-`m` (they're what made the cell hot in the first place), so per-route + range filter is the right granularity.
- Naive per-key map updates would serialise the hot path on the very keys we're trying to identify. Background aggregation keeps `Observe` near-lock-free.
- Approximate (Space-Saving) is the only bounded-memory option; the `error_bound = N/m` is returned in the response so callers can't confuse it with exact counts.

## Privacy & auth

This is the **first** keyviz feature that retains actual user key bytes. **Off by default** (`--keyvizHotKeysEnabled=false`); behind the existing admin SessionAuth (full-access role); audit log per query (counts logged, key bytes only in the response). In-memory only; never persists to Pebble or Raft.

## Open questions

§10 of the doc: sample rate / accuracy trade-off (`R = 16` vs `R = 4`), snapshot cadence, reads-series in v1, mixed-feature fan-out test.

## Five-lens self-review

In the doc (§11). Highlights: hot path lock-free (single-writer sketches, atomic.Pointer snapshot publish); bounded drop-on-overload preferring approximate top-K over latency; documented `error_bound` so callers can't mistake the result for exact; race-test gate.

## Next

Doc-only. Implementation + deploy wiring follow in a separate PR once this lands. CLAUDE.md design-doc-first.

## Summary by CodeRabbit

* **Documentation**
  * Added design proposal for KeyViz heatmap drill-down feature, outlining infrastructure and approach for detailed key inspection within selected cells.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/850?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/design/2026_05_28_proposed_keyviz_hot_key_topk.md
+++ b/docs/design/2026_05_28_proposed_keyviz_hot_key_topk.md
@@ -110,16 +110,16 @@ flow is:
    Observe (hot path)                       background goroutine
    ─────────────────                        ────────────────────
    if !hotKeysEnabled return                  drain ch
-   if rand/v2.IntN(R) != 0 return             for (rid, kc) := range ch:
-   if len(key) > maxKeyLen {                     per-route SS update (no
-     skippedLong.Add(1); return                   lock — single-writer);
-   }                                              SS COPIES key bytes,
+   if len(key) > maxKeyLen {                  for (rid, kc) := range ch:
+     skippedLong.Add(1); return                   per-route SS update (no
+   }                                              lock — single-writer);
+   if rand/v2.IntN(R) != 0 return                 SS COPIES key bytes,
    kc := pool.Clone(key)                          pool buffer released
-   select {                                    periodic atomic.Pointer.Store
-     case ch <- {rid, kc}:                       of a deep-copied snapshot
-     default: dropped.Add(1)                      (also reads dropped /
-       pool.Release(kc); drop                     skippedLong counters)
-   }
+   select {                                    periodic deep-copy snapshot
+     case ch <- {rid, kc}:                       publish via atomic.Pointer
+     default: dropped.Add(1)                     (reads dropped /
+       pool.Release(kc); drop                    skippedLong counters);
+   }                                            then reset live sketch
 ```
 
 Concretely:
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

Rebuttal: MAJOR-1 (write-count inflation via FSM apply) — premise is incorrect. `sampler.Observe` is called only from `kv/sharded_coordinator.go` (the coordinator dispatch path), never from `kv/fsm.go`. Followers don't observe writes that they replicate via Raft. Verified with `grep -rn 'sampler.Observe' --include='*.go'`. Added an explicit §6 note to preempt the misreading. Other findings addressed in `e926b06f → <latest>`.

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

PR #852, opened 2026-05-28
fix: prevent stale reads on follower nodes in gRPC handlers

bootjp:
## Root Cause

When a write (Put/Delete) arrives at a follower node it is forwarded to the leader via the internal `Forward` RPC. That RPC returns as soon as Raft quorum is reached, but the **originating follower may not have applied the log entry locally yet**. The subsequent `Get` call on the same follower computes `readTS = snapshotTS() = follower.LastCommitTS()`, which still reflects the pre-write watermark. The proxy read to the leader carries this stale timestamp, so `latestVisible(versions, stale_ts)` misses the just-committed write — producing either a nil result after a Put or a non-nil result after a Delete.

This caused `Test_grpc_transaction` to fail under `go test -race ./...` starting around iteration 9755, where cumulative race-detector slowdown first triggered leader churn and the follower redirect path.

## Changes

- **`adapter/ts.go`** — added `globalSnapshotTS(ctx, clk, st)`: uses `store.GlobalLastCommitTS` when the store supports it (`LeaderRoutedStore`). On the leader this is a zero-cost local `LastCommitTS()` read; on a follower it issues a single lightweight `RawLatestCommitTS` RPC to the leader, guaranteeing the snapshot timestamp is ≥ any write that has already returned to the client.

- **`adapter/grpc.go`** — updated the four gRPC read handlers (`Get`, `Scan`, `RawGet` with `ts=0`, `RawScanAt` with `ts=0`) to call `globalSnapshotTS` instead of `snapshotTS`.

## Testing

- `Test_grpc_transaction` passes all 9999 iterations locally (≈25 s).
- Full `adapter` package test suite passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Read operations now compute effective read timestamps using a global snapshot when available, with a reliable fallback to prior behavior—improves consistency for reads.

* **Documentation**
  * Added a formal TLA+ safety specification describing verification approach, safety invariants, liveness checks, modeling choices, and a multi-phase verification plan.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/852?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/adapter/grpc.go
+++ b/adapter/grpc.go
@@ -84,7 +84,7 @@ func (r *GRPCServer) clock() *kv.HLC {
 func (r *GRPCServer) RawGet(ctx context.Context, req *pb.RawGetRequest) (*pb.RawGetResponse, error) {
 	readTS := req.GetTs()
 	if readTS == 0 {
-		readTS = snapshotTS(r.clock(), r.store)
+		readTS = globalSnapshotTS(ctx, r.clock(), r.store)
 	}
 
 	v, err := r.store.GetAt(ctx, req.Key, readTS)
@@ -137,7 +137,7 @@ func (r *GRPCServer) RawScanAt(ctx context.Context, req *pb.RawScanAtRequest) (*
 
 	readTS := req.GetTs()
 	if readTS == 0 {
-		readTS = snapshotTS(r.clock(), r.store)
+		readTS = globalSnapshotTS(ctx, r.clock(), r.store)
 	}
 
 	var res []*store.KVPair
@@ -276,7 +276,7 @@ func (r *GRPCServer) Get(ctx context.Context, req *pb.GetRequest) (*pb.GetRespon
 		return nil, errors.WithStack(err)
 	}
 
-	readTS := snapshotTS(r.clock(), r.store)
+	readTS := globalSnapshotTS(ctx, r.clock(), r.store)
 	v, err := r.store.GetAt(ctx, req.Key, readTS)
 	if err != nil {
 		switch {
@@ -330,7 +330,7 @@ func (r *GRPCServer) Scan(ctx context.Context, req *pb.ScanRequest) (*pb.ScanRes
 			Kv: nil,
 		}, err
 	}
-	readTS := snapshotTS(r.clock(), r.store)
+	readTS := globalSnapshotTS(ctx, r.clock(), r.store)
 	res, err := r.store.ScanAt(ctx, req.StartKey, req.EndKey, limit, readTS)
 	if err != nil {
 		return &pb.ScanResponse{
--- a/adapter/ts.go
+++ b/adapter/ts.go
@@ -1,6 +1,8 @@
 package adapter
 
 import (
+	"context"
+
 	"github.com/bootjp/elastickv/kv"
 	"github.com/bootjp/elastickv/store"
 )
@@ -21,4 +23,22 @@ func snapshotTS(_ *kv.HLC, st store.MVCCStore) uint64 {
 		ts = ^uint64(0)
 	}
 	return ts
+}
+
+// globalSnapshotTS is like snapshotTS but uses GlobalLastCommitTS when the
+// store supports it (i.e. LeaderRoutedStore). On the leader this is
+// equivalent to LastCommitTS(); on a follower it queries the leader via RPC
+// so the snapshot timestamp is always >= the latest committed entry.
+// This prevents stale reads when a write was forwarded from the follower to
+// the leader and the follower has not yet applied the log entry locally.
+func globalSnapshotTS(ctx context.Context, clk *kv.HLC, st store.MVCCStore) uint64 {
+	type globalTS interface {
+		GlobalLastCommitTS(ctx context.Context) uint64
+	}
+	if g, ok := st.(globalTS); ok {
+		if ts := g.GlobalLastCommitTS(ctx); ts > 0 {
+			return ts
+		}
+	}
+	return snapshotTS(clk, st)
 }
--- a/adapter/ts.go
+++ b/adapter/ts.go
@@ -27,10 +27,16 @@ func snapshotTS(_ *kv.HLC, st store.MVCCStore) uint64 {
 
 // globalSnapshotTS is like snapshotTS but uses GlobalLastCommitTS when the
 // store supports it (i.e. LeaderRoutedStore). On the leader this is
-// equivalent to LastCommitTS(); on a follower it queries the leader via RPC
-// so the snapshot timestamp is always >= the latest committed entry.
-// This prevents stale reads when a write was forwarded from the follower to
-// the leader and the follower has not yet applied the log entry locally.
+// equivalent to LastCommitTS(); on a follower it prefers querying the
+// leader via RPC for fresher alignment. If leader lookup or the RPC fails
+// (nil store, empty leader address, connection error, RPC error, or
+// resp.Ts == 0), it falls back to local LastCommitTS() — see
+// kv/leader_routed_store.go GlobalLastCommitTS. On the leader path the
+// fresher-than-local guarantee is exact; on the follower it holds only
+// when the RPC succeeds and the leader has applied the relevant log
+// entry, so this best-effort path narrows but does not eliminate stale
+// reads triggered by writes that were forwarded from this follower to the
+// leader and have not yet been applied locally.
 func globalSnapshotTS(ctx context.Context, clk *kv.HLC, st store.MVCCStore) uint64 {
 	type globalTS interface {
 		GlobalLastCommitTS(ctx context.Context) uint64
--- a/docs/design/2026_05_28_proposed_tla_safety_spec.md
+++ b/docs/design/2026_05_28_proposed_tla_safety_spec.md
@@ -202,6 +202,25 @@ cannot check them as state invariants.
     current `kv/hlc.go` does not implement any of them, the TLC run for
     HLC-4 is expected to surface a counterexample — surfacing that
     counterexample is one of the motivating reasons for this spec.
+  - **(iii) Commit-time ceiling fencing.** Bounded skew (i) and logical
+    handoff (ii) are not enough on their own: a leader that misses
+    `RunHLCLeaseRenewal` for longer than `hlcPhysicalWindowMs` (e.g.
+    partitioned from the quorum that owns the ceiling group) but
+    remains a Raft leader will continue advancing its wall clock past
+    the last committed ceiling and can still issue persistence
+    timestamps. A subsequent leader that has only restored the same
+    stale ceiling and has a lagging wall clock can then issue a
+    strictly smaller timestamp, violating HLC-4. The spec therefore
+    requires every `Next()` that backs a persistence ts to be gated on
+    `wall_now < lastAppliedCeiling`; a leader whose ceiling has
+    expired must either re-apply a fresh ceiling via Raft before
+    committing, or refuse to commit (fail-closed). The current
+    implementation in `kv/sharded_coordinator.go` `RunHLCLeaseRenewal`
+    only logs renewal failures and dispatch paths call `HLC.Next()`
+    without checking ceiling freshness — this is the second design
+    gap the spec is meant to surface. As with (ii), if the gap is
+    real, TLC will produce a counterexample for HLC-4 under a model
+    that lets renewals fail.
 
 ### 5.2 OCC
 
@@ -420,7 +439,12 @@ The downloaded jar is checksum-pinned. No system-wide install required.
 
 `tla-check` is intended to run in CI on PRs that touch:
 
-- `kv/hlc*.go`, `kv/transaction.go`, `kv/fsm.go`, `kv/lock_resolver.go`
+- `kv/hlc*.go`, `kv/coordinator.go`, `kv/sharded_coordinator.go`,
+  `kv/transaction.go`, `kv/fsm.go`, `kv/lock_resolver.go` — the HLC
+  renewal scheduler, the timestamp-issuance dispatch path, and the
+  OCC commit-ts assignment all live in these files (per Section 3
+  anchors); a change here can violate HLC/OCC invariants and must
+  re-run `tla-check`.
 - `store/mvcc_store.go`
 - `distribution/`
 - the spec files themselves
--- a/docs/design/2026_05_28_proposed_tla_safety_spec.md
+++ b/docs/design/2026_05_28_proposed_tla_safety_spec.md
@@ -89,8 +89,8 @@ design abstraction.
 | Subsystem | Implementation anchors | Modelled state |
 |---|---|---|
 | Raft (abstracted) | `internal/raftengine/etcd`, `kv/fsm.go` | per-group log of committed entries + current leader + term; snapshot install marker |
-| HLC | `kv/hlc.go` (`Next`, `SetPhysicalCeiling`, `Observe`) | per-node `last`, per-node `physicalCeiling`; the ceiling is itself a Raft-applied entry |
-| OCC | `kv/transaction.go`, `kv/lock_resolver.go`, `kv/fsm_occ_test.go` | per-txn `read_ts`, `write_set`, `read_set`, lock map keyed by `(key, lock_ts)` |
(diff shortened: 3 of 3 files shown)


bootjp:
Addressed review round 1 in 4e7268f5:

- **gemini medium (line 190)**: added OCC-5 (start_ts consistency across all ops in a transaction; RMW re-uses read_ts).
- **codex P2 (line 190, OCC-4 is liveness)**: reclassified OCC-4 as bounded safety ("no stranded lock at quiescence"); moved the eventual-release form to a new §5.6 Liveness section as OCC-L1, with explicit fairness assumptions and `PROPERTY` (not `INVARIANT`) framing.
- **codex P2 (line 213, Composed-1 / Routes-3 too strong)**: weakened Routes-3 to catalog-atomicity + handling-node sync only; called out that propagation to other nodes' `RouteEngine`s is asynchronous via `CatalogWatcher` and is covered by Routes-4. Cross-referenced `adapter/distribution_server.go` (`saveSplitResultViaCoordinator` + `applyEngineSnapshot`) and `distribution/watcher.go` so the abstraction matches the implementation.
- Also added Routes-L1 as the liveness counterpart of Routes-4 (eventual catalog convergence under fairness on `CatalogWatcher.SyncOnce`).
- Updated §4.1 / §5 intro / §8 milestone table to keep the safety vs liveness distinction consistent.

No code semantic changes (docs-only PR), so no caller audit was needed.

@claude review

bootjp:
Addressed review round 2 in `a74f5838`:

**codex P2 (commit `4e7268f5`, line 181) — HLC-4 logical handoff**
Strengthened HLC-4 with explicit preconditions:
1. `MaxClockSkewMs < hlcPhysicalWindowMs` ASSUME so the renewed ceiling actually floors the new leader.
2. **Logical-counter handoff requirement.** The 16-bit logical half is in-memory only, so a fresh leader can issue `(ceiling, 0)` while the previous leader committed `(ceiling, k≥1)`. Spec now requires one of three normative strategies (persist max-logical alongside ceiling / +1 ms on election / FSM scan + `Observe` on election). If the current `kv/hlc.go` does not implement any of them, TLC is expected to surface a counterexample — surfacing it is one of the motivating reasons for this spec.

**codex P2 (commit `4e7268f5`, line 318) — SYMMETRY scope**
§6.3 replaced the blanket "SYMMETRY over nodes and keys" with a per-module table. Keys are symmetric in HLC + OCC only; not in MVCC, Routes, or Composed (they participate in ordered ranges via `SplitRange`). Full key symmetry would collapse the stale-route / data-handoff schedules the spec is meant to analyse.

**claude bot review (6 findings)**
1. §3 anchor table — added `kv/coordinator.go` + `kv/sharded_coordinator.go` to HLC row (where `hlcRenewalInterval` / `hlcPhysicalWindowMs` are defined).
2. Routes-L1 — renamed `engineVersion[n]` → `catalogVersion[n]` to match `distribution/engine.go` `Engine.catalogVersion`.
3. §3 OCC state — added `start_ts` as first-class state now that OCC-5 needs it.
4. Composed-2 — noted SplitRange is same-group only today (per CLAUDE.md); the cross-group clause is a forward-looking guard rail.
5. OCC-3 — added lock-encoding clarification: OCC-3 holds only when no conflicting lock at `lock_ts ≤ read_ts`; conflicting locks are resolved via a `LockResolver` action.
6. HLC-4 skew guard — covered by the new precondition (i) above (same finding as codex P2 from a different angle).
- Nit (§7.1): enumerated `tla/README.md` required sections so M1 has a self-contained contract.
- Nit (§8.2): tightened `partial` → `implemented` promotion criterion to "M5's TLC run passes in CI on main".

Docs-only change. No semantic changes to Go code; no caller audit was required. The new HLC-4 preconditions are normative for M1 and may surface as a follow-up code change against `kv/hlc.go`.

@claude review

bootjp:
Addressed review round 3 in `5b66e629`. Both codex P2 findings on `a74f5838`:

**codex P2 (line 187) — Fence writes when HLC ceiling expires**

Added HLC-4 precondition **(iii) Commit-time ceiling fencing**. The earlier preconditions (i) bounded skew and (ii) logical handoff are not enough on their own: a leader that misses `RunHLCLeaseRenewal` for longer than `hlcPhysicalWindowMs` (e.g. partitioned from the ceiling-owning quorum) but remains Raft leader continues advancing its wall clock past the last committed ceiling, and can still issue persistence timestamps. A subsequent leader that has only restored the stale ceiling and has a lagging clock can then issue a strictly smaller ts.

Spec now requires every `Next()` backing a persistence ts to be gated on `wall_now < lastAppliedCeiling` — the leader must either re-apply a fresh ceiling via Raft or refuse to commit (fail-closed). The current `kv/sharded_coordinator.go` `RunHLCLeaseRenewal` only logs renewal failures and dispatch paths call `HLC.Next()` without checking freshness; this is the second design gap the spec is meant to surface via a TLC counterexample for M1.

**codex P2 (line 425) — Run TLA checks for timestamp coordinator changes**

Added `kv/coordinator.go` and `kv/sharded_coordinator.go` to the §7.3 CI path filter. `RunHLCLeaseRenewal`, `nextStartTS`, and the commit-ts dispatch live there (per §3 anchors); a change in those files can violate HLC/OCC invariants without retriggering `tla-check`.

Docs-only change. Precondition (iii) is normative for the M1 spec implementation and may surface as a follow-up code change against the coordinator + `HLC.Next()` gate; the Go code itself was not modified in this commit, so no caller audit was required.

@claude review

bootjp:
Addressed claude bot's round 3 review in `e5ca47f8`. All six findings:

1. **HLC-4 precondition (i) prose direction.** Rewrote the justification — the hazard is logical-counter overflow pushing the issued wall part to `ceiling+1`, not "wall ahead of ceiling" (which is the safe path). The `ASSUME MaxClockSkewMs < hlcPhysicalWindowMs` itself was already correct.

2. **HLC-4 (ii) logical-handoff strategies.** Option **(b)** now explicitly requires a serving fence (the new leader must refuse to serve persistence `Next()` until the election-triggered ceiling proposal is applied — otherwise a request landing between `BecomeLeader` and the first ceiling apply still issues from the stale ceiling). Strategy **(c)** (`hlc.Observe(fsm.MaxAppliedHLC())` on election) is designated as the **M1 default** — simplest to model and implement, no extra fence required.

3. **Routes-L1 `LiveNodes` qualifier.** `<>(∀ n: catalogVersion[n] ≥ v)` would be falsified by any permanently-crashed node since it never resumes `CatalogWatcher.SyncOnce`. Replaced with `<>(∀ n ∈ LiveNodes: catalogVersion[n] ≥ v)` and noted the same pattern applies to OCC-L1 and future `∀`-shaped liveness properties.

4. **Composed-3 real-time encoding.** TLA+ has no inherent real-time. M5 adopts the **Raft-log-index proxy** (`(group_id, index)` strictly smaller ⇒ happens-before) as the primary encoding — conservative but cheap. Auxiliary `realtime_order` history variable in `lib/Env.tla` documented as the fallback if the proxy is too lossy; M5 PR makes the final choice.

5. **§6.3 HLC keys symmetry rephrased.** "Keys are symmetric in HLC.tla (keys do not appear)" → "HLC.tla references no key set; `MCHLC.cfg` omits the `Keys` constant entirely." Removes the spurious question of declaring a `Keys` set in `MCHLC.cfg`.

6. **§8.1 M1 milestone note.** M1 row now explicitly recommends strategy (c) as the default logical-handoff and requires deviations to be justified in the M1 PR.

Docs-only change. Normative spec changes (strategy (c) on election, serving fence for option (b)) are guidance for the M1 implementation; no Go code modified in this commit, so no caller audit was required. Caller audits will be due in the M1 implementation PR against `ShardedCoordinator` and `HLC`.

@claude review

bootjp:
Addressed claude bot's round 4 review (on `5b66e629`, posted 13:37:22 — I missed reading it before pushing `e5ca47f8`). Round 5 in `cd03835a`. All five findings:

1. **Variable naming alignment.** HLC-4 (iii) referred to `lastAppliedCeiling` while §3 uses `physicalCeiling` (matching `kv/hlc.go`). Renamed (iii) to `physicalCeiling[n]` so M1 spec authors do not define two variables for one concept.

2. **HLC-4 (iii) ceiling-group anchor.** `ShardedCoordinator`'s ceiling group is always `c.defaultGroup` (`kv/sharded_coordinator.go`), so the failure scenario is precisely "a node that leads a non-default shard but is partitioned from the default group's quorum". Added this anchor parenthetically in (iii).

3. **§8.1 "TLC passes" vs "TLC will produce a counterexample" contradiction resolved.** The intended semantics is: the M1 spec encodes preconditions (i)/(ii)/(iii) as TLA+ `ASSUME`s (spec of the correct design), so TLC passes. A separate `MCHLC_gap.cfg` configuration that removes the `ASSUME`s is expected to surface counterexamples demonstrating the preconditions are necessary. Both configs land in M1; the gap config is the motivating evidence for the follow-up Go code fixes.

4. **§8.1 M1 scope mentions precondition (iii).** Previously only (ii)'s strategy choice was called out; M1 must encode all three preconditions, including the ceiling-fencing gate.

5. **§9 risk 7: fail-closed availability under partition.** Precondition (iii)'s fail-closed semantics is a CP-not-AP trade-off — a leader partitioned from the default group's quorum for longer than `hlcPhysicalWindowMs` (3s) rejects all client commits until renewal. Stricter than the current implementation. Cross-linked from precondition (iii)'s new "Availability note" sentence so operators planning deployments see it.

Docs-only change. The normative requirements in (iii) — fencing gate + fail-closed — will translate into Go code in the M1 implementation PR, where caller audits of `HLC.Next()` and `ShardedCoordinator.RunHLCLeaseRenewal` will be required (return-value semantic change: `Next()` may need an error return path).

@claude review

bootjp:
Addressed claude bot's two round-5 reviews (on `e5ca47f8` and `cd03835a`). Round 6 in `1c4a2369`.

The reviews confirmed all round-4/round-5 prior findings were correctly addressed. Four **new** findings remained:

1. **`fsm.MaxAppliedHLC()` does not exist — strategy (c) needs new API.** Clarified that `kv/fsm.go` currently tracks only the physical ceiling, so strategy (c) requires adding `MaxAppliedHLC()` to the FSM (preferred: running max across applied writes; fallback: scan on election). "Simplest" was misleading without this — re-framed as "smallest API surface, not trivial one-liner".

2. **OCC-L1 needs `LiveTransactions` qualifier applied directly.** The trailing Routes-L1 note anticipated this for OCC-L1 but the property itself was unqualified. A crashed-mid-flight transaction never releases its lock through the owner path; the lock-resolver action then becomes the only path and its progress depends on `LiveNodes` fairness. Property is now `lockHeld(k, ts) ∧ owner ∈ LiveTransactions ⇒ <> ¬lockHeld(k, ts)`.

3. **Composed-3 Raft-log-index proxy scoped to same-group only.** Lexicographic `(group_id, index)` would make all group-1 writes happen-before group-2 writes regardless of timing. Re-scoped: same-group uses log index (total order); cross-group uses HLC `commit_ts` (the only global clock both groups share). `realtime_order` history variable remains the fallback.

4. **§8.1 TLA+ terminology precision.** Only precondition (i) fits `ASSUME` (constant predicate `MaxClockSkewMs < HlcPhysicalWindowMs`). Preconditions (ii) and (iii) are behavioural and must be **action guards** — strategy (c) is a state update in `BecomeLeader`, the ceiling fence is an enabling condition on `IssueTimestamp`. The companion `HLC_gap.tla` + `MCHLC_gap.cfg` is now described as "removes the `ASSUME` *and* relaxes the action guards" so M1 authors taking the earlier wording literally don't produce invalid TLA+.

Docs-only change. The new normative requirement against `kv/fsm.go` (adding `MaxAppliedHLC`) is spec-level guidance for the M1 implementation PR; no Go code modified here.

@claude review

bootjp:
Addressed claude bot's round 6 review (on `1c4a2369`) in `8ed93276` (round 7). Three substantive findings + one nit:

1. **§9 risk 7 internal contradiction with §8.1 — `ASSUME` vs action guard.** §8.1 was updated in round 6 to correctly describe precondition (iii) as an action guard, but §9 risk 7 still said "encodes the fence as `ASSUME`". `ASSUME` is only valid for constant predicates. Updated risk 7 to say "action guard on `IssueTimestamp` (ENABLED ⇔ wallNow[n] < physicalCeiling[n])" matching §8.1.

2. **§3 OCC lock map value type unspecified.** OCC-L1 (added in round 5) uses `owner(k, ts) ∈ LiveTransactions`. For `owner` to be derivable, the lock map must record the owning txn. Updated §3 from "lock map keyed by (key, lock_ts)" to "lock map: (key, lock_ts) → txn_id" with an explicit note about deriving `owner`.

3. **HLC `commit_ts` cross-group collision risk for Composed-3.** Per CLAUDE.md and `kv/hlc.go`, the physical ceiling is Raft-replicated but the 16-bit logical half is per-node in-memory — two leaders of different groups on different nodes can independently produce identical 64-bit timestamps. Resolution: M5 commits to a **`DISTINCT_TS` `StateConstraint`** as the default (sound under bounded-skew), with a **`(node_id, group_id)` tiebreaker** documented as the fallback if removing `DISTINCT_TS` surfaces interesting counterexamples. `realtime_order` history variable remains the deeper fallback.

4. **(nit)** §8.1 `BecomeLeader` pseudocode wrapped in `LET maxAppliedHLC == MaxAppliedHLC(fsm[n]) IN …` so `maxAppliedHLC` is no longer a free identifier; LET cross-references the new FSM operator from §5.1.

Rebased over the new `38fac251 fix: use GlobalLastCommitTS in gRPC read handlers...` commit that was pushed to this branch — that commit is code-level (orthogonal to the spec) and doesn't change the design-level invariants documented here.

Docs-only change to my commit. No Go code modified by me; no caller audit required for the spec PR.

@claude review

bootjp:
Coderabbit's round on `38fac251` / `8ed93276` produced 3 findings. Disposition in `14b00114`:

1. **`docs/...spec.md` line 606 (Major) — risk 7 ASSUME contradicts §8.1.** **Stale** — this was reviewed at the `38fac251` snapshot, which had my round 6 doc (line 606 ≈ risk 7 ASSUME). My round 7 commit `8ed93276` already replaces that with "action guard on `IssueTimestamp`". Current line 606 in `8ed93276` / `14b00114` is the TSO-model risk, unrelated. No action needed.

2. **`docs/...spec.md` line 109 (Minor) — MD040 language tag.** Fixed in `14b00114`: §4.1 directory-layout fenced code block now tagged ` ```text `.

3. **`adapter/ts.go` line 33 (Minor) — overstated freshness guarantee.** This is on `38fac251`, **not my commit**. The change was authored by @bootjp directly to this branch and the comment wording is their call. Deferring to @bootjp on whether to soften the wording; I have not modified that file.

No further docs findings outstanding on the spec from this coderabbit pass.

@claude review

bootjp:
Round 8 in `e0f887bc`. Addressed claude bot's round 7 review on `8ed93276` (3 findings + 1 nit):

1. **§3 OCC lock map: `txn_id` → `start_ts`.** Round 7 introduced `txn_id` as the lock map value type, but transactions are identified by `start_ts` throughout the spec (OCC-5) and the implementation field is `txnLock.StartTS` (`kv/txn_codec.go`). Renamed to `start_ts` and cross-referenced the implementation field + OCC-5 so M2 authors can derive `owner(k, ts)` without re-reading code.

2. **`adapter/ts.go` overstated freshness comment.** Confirmed against `kv/leader_routed_store.go` `GlobalLastCommitTS`: five fallback paths return `local.LastCommitTS()` (nil store, empty leader addr, conn error, RPC error, `resp.Ts == 0`). Rewrote the comment to enumerate the fallback paths and state that the follower path "narrows but does not eliminate" stale reads. **Comment-only change** — function body and callers unchanged, so no caller audit required.

3. **§5.5 `DISTINCT_TS` soundness (nit).** Added a parenthetical clarifying that a `StateConstraint` prunes reachable states rather than proving the invariant in pruned ones — Composed-3 is proved against the realistic regime, with pathological collisions documented by tiebreaker (ii) so the pruning is intentional.

**Disposition of remaining coderabbit findings:**
- Line 109 MD040 was already addressed in `14b00114`; claude bot flagged it again because it reviewed the pre-fix snapshot `8ed93276`. Verified current state in `e0f887bc`: ` ```text ` tag is present. No further action.

@claude review

PR #856, opened 2026-05-28
tla: M1 — HLC.tla + lib + make tla-check (HLC-1..4 invariants + gap counterexample)

bootjp:
## Summary

M1 of the [TLA+ safety spec proposal](https://github.com/bootjp/elastickv/blob/feat/tla-m1-hlc-spec/docs/design/2026_05_28_partial_tla_safety_spec.md) lands:

- `tla/lib/Env.tla` + `tla/lib/Raft.tla` (shared scaffolding).
- `tla/hlc/HLC.tla` with invariants **HLC-1..HLC-4** + `TypeOK`.
  - **Precondition (i)** bounded clock skew encoded as a TLA+ `ASSUME` in `Env.tla`.
  - **Precondition (ii)** logical-counter handoff — strategy (c) — encoded as a state update inside `BecomeLeader_HLC` (`hlc.Observe(maxAppliedHLC)` analogue).
  - **Precondition (iii)** ceiling fence encoded as an enabling action guard on `IssueTimestamp` (`wallNow[n] < physicalCeiling[n]`).
- `tla/hlc/MCHLC.tla` + `MCHLC.cfg` (correct design — TLC passes) and `MCHLC_gap.cfg` (preconditions disabled — TLC produces the HLC-4 counterexample the design doc anticipated).
- `make tla-check` Makefile target: downloads checksum-pinned `tla2tools.jar` v1.8.0 (`SHA-256 237332bd…`) to `.cache/tla/` and runs both configurations, inverting the exit code on the gap config so the expected-FAIL counts as PASS for CI.
- `tla/README.md` covering install / run / what each module proves / reading list / how to map a TLC counterexample back to the implementation, per §7.1 of the design doc.
- Design doc renamed `proposed_` → `partial_` (per CLAUDE.md lifecycle), header `Status:` updated, in a dedicated commit.

## Verification (local)

```
$ make tla-check
================================================================
  TLC: tla/hlc/MCHLC.cfg  (correct design, expected PASS)
================================================================
... Model checking completed. No error has been found.
... 61 states generated, 28 distinct states found, 0 states left on queue.

================================================================
  TLC: tla/hlc/MCHLC_gap.cfg  (no preconditions, expected FAIL)
================================================================
... Error: Invariant HLC4_NoRegressionAcrossTerms is violated.
... 44 states generated, 28 distinct states found, 14 states left on queue.
... depth 5

OK: MCHLC_gap.cfg failed as designed (HLC-4 counterexample).

tla-check: all model-check outcomes match the design contract.
```

The gap-config counterexample (depth 5):

1. `BecomeLeader(n1)` — n1 elected term 1
2. `IssueTimestamp(n1)` — commits `(wall=0, logical=0)` in term 1
3. `IssueTimestamp(n1)` — commits `(0, 1)` in term 1 (logical increments)
4. `BecomeLeader_HLC(n2)` — n2 elected term 2; **EnableSafety = FALSE** so strategy (c) does NOT raise n2's `hlcLast` to `maxAppliedHLC = (0, 1)`
5. `IssueTimestamp(n2)` — commits `(0, 1)` in term 2 (same as term 1's last commit; HLC-4 requires strict `>`)

This is exactly the bug the design doc §5.1 HLC-4 (ii) said the spec would surface.

## Self-review (per CLAUDE.md five-lens checklist)

1. **Data loss** — n/a: spec-only PR, no persistence paths touched.
2. **Concurrency / distributed failures** — the spec itself models the concurrency bugs (HLC-4 cross-leader regression); both `make tla-check` configurations explore the full bounded state space.
3. **Performance** — `make tla-check` runs in <1s on a developer laptop at default bounds.
4. **Data consistency** — `MCHLC.cfg` proves HLC-1..HLC-4 hold under preconditions (i)/(ii)/(iii); `MCHLC_gap.cfg` proves the preconditions are necessary by surfacing the documented counterexample.
5. **Test coverage** — Go code untouched, no Go tests added. `make tla-check` is the new coverage layer; running it is the test.

## Out of scope

This PR is **spec-only**. The Go code changes that strategy (c) and the ceiling fence require (per §5.1 HLC-4 (ii) and (iii)) are not in this PR — they need:

- Add `MaxAppliedHLC()` to `kv/fsm.go` (running max across applied write entries).
- Call `hlc.Observe(fsm.MaxAppliedHLC())` on new-term detection in `kv/sharded_coordinator.go` `RunHLCLeaseRenewal`, before any persistence `Next()` is served.
- Add `wallNow < physicalCeiling` fence in `kv/hlc.go` `Next()` with a fail-closed return path and operator-visible metrics — caller audits of every site that calls `Next()` will be required (per the loop-feedback discipline) since `Next()` changes from infallible to error-returning.

Tracked as a follow-up issue; this PR lands the design contract that the follow-up must satisfy.

## Test plan

- [ ] CI runs `make tla-check` (CI integration is deferred to a future PR per §7.3 of the design doc — local invocation is the contract for now)
- [ ] Reviewer runs `make tla-check` locally and confirms the safe pass + gap-config counterexample
- [ ] Reviewer cross-checks each invariant in `HLC.tla` against the corresponding entry in §5.1 of the design doc
- [ ] Reviewer cross-checks the implementation anchors in `tla/README.md` "How to interpret a TLC failure" against the current Go source

### DIFF
--- a/.github/workflows/tla-check.yml
+++ b/.github/workflows/tla-check.yml
@@ -66,13 +66,18 @@ jobs:
       # GitHub release artefacts on every run.  Key includes the Makefile
       # hash so a version-pin change invalidates the cache automatically
       # (TLA_VERSION + TLA_SHA256 both live in the Makefile).
+      #
+      # NO restore-keys: a prefix-match restore would bring back an older
+      # jar after a TLA_VERSION/TLA_SHA256 bump.  Because the Makefile's
+      # $(TLA_JAR) recipe is gated on file existence, the restored stale
+      # jar would skip checksum verification entirely — defeating the
+      # whole point of the pin (codex P2 on PR #856 round 4).  A full
+      # miss-and-redownload is correct here.
       - name: Cache tla2tools.jar
         uses: actions/cache@v5
         with:
           path: .cache/tla
           key: ${{ runner.os }}-tla-tools-${{ hashFiles('Makefile') }}
-          restore-keys: |
-            ${{ runner.os }}-tla-tools-
 
       - name: Pre-fetch tla2tools.jar (logs separately from tla-check)
         run: make tla-tools
--- a/.github/workflows/tla-check.yml
+++ b/.github/workflows/tla-check.yml
@@ -1,0 +1,81 @@
+name: TLA+ model check
+
+# Runs `make tla-check` on PRs (and pushes to main) that touch the
+# subsystems modelled by the spec or the spec / tooling themselves.
+# Per docs/design/2026_05_28_partial_tla_safety_spec.md §7.3.
+#
+# The job runs both MCHLC.cfg (correct design — expected PASS) and
+# MCHLC_gap.cfg (preconditions disabled — expected FAIL with the HLC-4
+# counterexample); the Makefile target greps for the specific HLC-4
+# violation string so the gap evidence is validated, not just the exit
+# code.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'tla/**'
+      - 'Makefile'
+      - 'kv/hlc*.go'
+      - 'kv/coordinator.go'
+      - 'kv/sharded_coordinator.go'
+      - 'kv/transaction.go'
+      - 'kv/fsm.go'
+      - 'kv/lock_resolver.go'
+      - 'store/mvcc_store.go'
+      - 'distribution/**'
+      - '.github/workflows/tla-check.yml'
+  push:
+    branches: [main]
+    paths:
+      - 'tla/**'
+      - 'Makefile'
+      - 'kv/hlc*.go'
+      - 'kv/coordinator.go'
+      - 'kv/sharded_coordinator.go'
+      - 'kv/transaction.go'
+      - 'kv/fsm.go'
+      - 'kv/lock_resolver.go'
+      - 'store/mvcc_store.go'
+      - 'distribution/**'
+      - '.github/workflows/tla-check.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-tla-check
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  tla-check:
+    name: tla-check
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Check out
+        uses: actions/checkout@v6
+
+      - name: Set up JDK 21 (Temurin)
+        uses: actions/setup-java@v5
+        with:
+          distribution: temurin
+          java-version: '21'
+
+      # Cache the checksum-pinned tla2tools.jar so we don't re-download
+      # GitHub release artefacts on every run.  Key includes the Makefile
+      # hash so a version-pin change invalidates the cache automatically
+      # (TLA_VERSION + TLA_SHA256 both live in the Makefile).
+      - name: Cache tla2tools.jar
+        uses: actions/cache@v5
+        with:
+          path: .cache/tla
+          key: ${{ runner.os }}-tla-tools-${{ hashFiles('Makefile') }}
+          restore-keys: |
+            ${{ runner.os }}-tla-tools-
+
+      - name: Pre-fetch tla2tools.jar (logs separately from tla-check)
+        run: make tla-tools
+
+      - name: Run make tla-check
+        run: make tla-check
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,11 @@ jepsen/.ssh/
 .golangci-cache/
 server
 
+# TLA+ TLC artifacts (counterexample traces, state directories)
+tla/**/MCHLC_TTrace_*.tla
+tla/**/MCHLC_TTrace_*.bin
+tla/**/states/
+
 # Admin SPA build outputs. The placeholder internal/admin/dist/index.html
 # is committed so `go build` succeeds in a fresh clone (the //go:embed
 # directive needs at least one matching file). Everything else under
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,18 @@ $(TLA_JAR):
 	@mkdir -p $(dir $(TLA_JAR))
 	@echo "Downloading tla2tools.jar $(TLA_VERSION)..."
 	@curl -fsSL -o $(TLA_JAR).tmp $(TLA_URL)
-	@actual=$$(shasum -a 256 $(TLA_JAR).tmp | awk '{print $$1}'); \
+	@# Prefer sha256sum (GNU coreutils, universal on Linux); fall back to
+	@# shasum -a 256 (default on macOS).  Either yields the same hex
+	@# digest in the first whitespace-delimited field.
+	@if command -v sha256sum >/dev/null 2>&1; then \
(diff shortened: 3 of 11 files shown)


bootjp:
Addressed PR #856 round 1 review in `791c65b3`. Four findings:

1. **gemini HIGH (HLC.tla line 97) — Nat subtraction underflow.** Rewrote the `TickWall` skew check using addition instead of subtraction. The old form `(wallNow[n] + 1) - wallNow[m] <= MaxClockSkewMs` would underflow in TLA+ Naturals the moment `MaxClockSkewMs > 0` admitted an out-of-order tick, halting TLC. The equivalent additive form stays inside Nat.

2. **gemini MEDIUM (Makefile) — sha256sum portability.** Prefer GNU `sha256sum` (universal on Linux), fall back to macOS `shasum -a 256`, error out if neither is available. Without this `make tla-check` failed on minimal Linux CI images.

3. **codex P2 (HLC.tla line 180) — HLC-1 was vacuously true.** Old form `HLCLE(r1, r2) \/ HLCLE(r2, r1)` is just totality on a totally ordered set, so a regression `(5, 0)` → `(4, 0)` would have passed. Added a per-commit `seq` ordinal (= global `opCount` at issuance time) and changed the invariant to `r1.node = r2.node /\ r1.seq < r2.seq => HLCLT(r1.ts, r2.ts)`. Also added the transition-level `HLC1_Action` property to pin "hlcLast weakly increases on every step" so future actions cannot silently regress.

4. **codex P2 (HLC.tla line 186) — HLC-2 was vacuously implied by TypeOK.** Old form `physicalCeiling[n] >= 0` was just type non-negativity. Replaced with `HLC2_NonNeg` (the residual TypeOK content; harmless) plus the real `HLC2_Action == [][\A n: physicalCeiling[n] <= physicalCeiling'[n]]_vars` transition property registered under PROPERTIES.

The TLA+ "same set of ceiling entries → equal across nodes" half of HLC-2 is structurally satisfied in M1 because `ApplyCeiling` propagates atomically (per the action comment); per-node async propagation is an M5 concern when route-catalog/HLC interactions cross multiple groups.

**Verification.** `make tla-check` still PASSES on `MCHLC.cfg` (now also exercising the two action properties) and FAILS on `MCHLC_gap.cfg` as designed, with the HLC-4 counterexample at depth 5 unchanged. No Go code semantics changed — spec-only PR, no caller audit required.

@claude review

bootjp:
PR #856 round 2 review (codex P2 ×2 on `791c65b3`) addressed in `1123a0e1`:

1. **codex P2 (HLC.tla line 101) — TickWall self-skew block.** The `\A m \in Nodes:` quantifier included `m = n`, so under `MaxClockSkewMs = 0` the constraint became `wallNow[n] + 1 <= wallNow[n] + 0` (unsatisfiable). TickWall was never enabled and the model **never explored wall-clock advancement, ceiling expiry, or skew schedules** — the safe config was passing on a degenerate 28-state subspace. Excluded self from the universal: `\A m \in Nodes \ {n}: …`.

  Also raised `MaxClockSkewMs` from 0 to 1 and `HlcPhysicalWindowMs` from 1 to 2 (so the ASSUME `MaxClockSkewMs < HlcPhysicalWindowMs` still holds) in both `.cfg` files so the skew dimension is actually part of the explored space.

2. **codex P2 (Makefile line 81) — Validate gap failure reason.** `tla-check` previously accepted any non-zero exit from the gap TLC run as the expected HLC-4 counterexample. Capture stdout/stderr, print it, then grep for `Invariant HLC4_NoRegressionAcrossTerms is violated` before declaring the gap evidence intact. A parse error, deadlock, JVM crash, or different invariant violation now correctly fails the target instead of masking a harness regression.

**Verification.** Safe config explores **3,594 distinct states** (vs 28 previously) and completes without invariant violation. Gap config still surfaces the HLC-4 counterexample at depth 5; the Makefile correctly identifies it by string match.

Spec/Makefile-only change, no Go code touched. No caller audit required.

@claude review

bootjp:
Addressed claude bot's round 2 review (on `791c65b3`) in `9901e6ca`.

1. **MEDIUM (stale `proposed_` path):** Updated 4 broken references to the design doc path that became invalid after the `proposed_` → `partial_` rename in `66b51539`:
   - `tla/hlc/HLC.tla` line 4
   - `tla/lib/Env.tla` line 4
   - `tla/lib/Raft.tla` line 4
   - `Makefile` line 17

2. **LOW (README invariant table):** `tla/README.md` listed `HLC2_CeilingMonotonic`, which no longer exists after round 1's rework. Replaced with the actual operators: `HLC1_PerNodeMonotonic`, `HLC1_Action` (PROPERTY), `HLC2_NonNeg`, `HLC2_Action` (PROPERTY); also clarified HLC-1's state form uses the strict `seq` ordering, not the earlier vacuous total-order form.

3. **INFO (TickWall `MaxClockSkewMs = 0`):** claude bot was reviewing round-1 commit `791c65b3`; this finding is already addressed in round-2 commit `1123a0e1` (TickWall now excludes self, `MaxClockSkewMs` raised to 1, safe config explores 3,594 distinct states).

Docs-only changes. `make tla-check` still passes on MCHLC.cfg (3,594 distinct states) and fails on MCHLC_gap.cfg as designed.

@claude review

bootjp:
Wired `make tla-check` into GitHub Actions in `ca0ecfc0`.

New file: `.github/workflows/tla-check.yml`
- Triggers on PR (opened / synchronize / reopened) and push to `main`, with a path filter matching §7.3 of the design doc (`tla/**`, `Makefile`, the HLC/coordinator/transaction/fsm/lock_resolver Go files, `store/mvcc_store.go`, `distribution/**`, and the workflow file).
- Temurin JDK 21 (consistent with `jepsen-test.yml`).
- Caches `.cache/tla/tla2tools.jar` keyed on Makefile hash so any TLA_VERSION / TLA_SHA256 change invalidates the cache automatically.
- Pre-fetches the jar via `make tla-tools` so the download is logged separately from the model check.
- Runs `make tla-check`, which keeps the round-2 gap-config validation: a non-zero TLC exit on `MCHLC_gap.cfg` only counts as success if the output contains the exact `Invariant HLC4_NoRegressionAcrossTerms is violated` string.

Updated §7.3 of the design doc: removed the `(deferred)` qualifier and named the workflow file as the now-active CI binding.

`actionlint` clean. The workflow is already running on this PR (`ca0ecfc0`): https://github.com/bootjp/elastickv/actions/runs/26591495196

bootjp:
Addressed codex P2 on `ca0ecfc0` in `2a2238af`.

**Finding**: `.github/workflows/tla-check.yml` cache `restore-keys` could bring back an older `tla2tools.jar` after a `TLA_VERSION` / `TLA_SHA256` bump in the Makefile. Because the Makefile's `$(TLA_JAR)` target is gated on file existence, the restored stale jar would skip the checksum verification entirely — defeating the pin.

**Fix**: Dropped the `restore-keys` block. A full cache miss is the correct behaviour on a version-pin change: the recipe runs, downloads from GitHub, and verifies the SHA-256 before serving the jar to TLC. Documented inline so the next reviewer doesn't re-add it.

This is a workflow-only change. The Makefile's checksum verification path is unchanged — and is now load-bearing for every version-pin bump.

@claude review

PR #854, opened 2026-05-28
feat(keyviz): implement per-cell hot-key Top-K drill-down (Phase 2-A++)

bootjp:
## What

Implements the merged design **PR #850** (`docs/design/2026_05_28_proposed_keyviz_hot_key_topk.md`): per-cell hot-key drill-down. Each individual route owns a Space-Saving sketch fed off the hot path via a single bounded channel + one background aggregator goroutine. A new admin endpoint returns the actual hottest keys for a clicked heatmap cell.

**Off by default** — opt in with `--keyvizHotKeysEnabled`. Disabled-case overhead on the sampler hot path is one nil-check branch.

## Scope

Single-node v1: sampler core + admin endpoint + 5 flags + tests + admin auth/audit. Three pieces deliberately deferred (recorded in the commit message and below): **fan-out merge**, **deploy wiring**, **SPA cell-click handler**. The handler design supports fan-out cleanly (matches the matrix fan-out shape).

## How (mapping to design §)

| § | Decision | Implementation |
|---|---|---|
| §3 | Per-route Space-Saving (m, default 64, cap 256) | `keyviz/space_saving.go` — Misra-Gries-equivalent, SS-owned key copies, deep-copied snapshots. Heavy-hitter guarantee pinned by a rapid property. |
| §4 hot path | length-check **FIRST** → splitmix64 sample gate → pool clone → non-blocking channel send | `keyviz/hot_keys.go::observe`. Length skipped count is deterministic (not subject to R) per Codex round-7 L116. |
| §4 PRNG | lock-free, race-free, no shared Source ownership | atomic.Uint64 + splitmix64 mixer. Closes Codex P1 round-4 (PCG sync) + L135 (phase-lock) + gosec G404 in one design — no `math/rand/v2` import. |
| §4 aggregator | single goroutine, single writer to every per-route sketch | `keyviz/hot_keys.go::run` — `select { case <-ch: ...; case <-tick: drain+publish+reset }`. Reset on every publish gives one-keyvizStep-window snapshots (Codex P2 round-6 L186). |
| §4 ownership | pool buffer transient; SS copies key bytes; snapshot deep-copies again | implemented in `consume` + `spaceSaving.observe` + `spaceSaving.snapshot`. Gemini medium round-1 use-after-free closed. |
[...]

PR #863, opened 2026-05-29
feat(deploy): wire KeyViz hot-key Top-K flags through rolling-update

bootjp:
## What

Closes the **deploy-wiring** deferred follow-up from PR #854 (KeyViz per-cell hot-key Top-K drill-down, Phase 2-A++). The five new `--keyvizHotKeys*` CLI flags are now reachable from operator env via `rolling-update.sh`, mirroring the existing `--keyvizEnabled` + `--keyvizFanoutNodes` wiring pattern.

## Scope

Pure deploy-script change (2 files). No Go changes, no schema change, no SPA change, no new design doc — the design is in `docs/design/2026_05_28_proposed_keyviz_hot_key_topk.md` §8.

### `scripts/rolling-update.env.example`
Documented block for the five hot-keys env vars (master switch + four tuning knobs). Defaults are left **empty** so the daemon falls back to the sampler-side constants in `keyviz/sampler.go`; bumping a Go-side default in a future release then rolls out without an env-file edit.

### `scripts/rolling-update.sh` (five touchpoints, mirrors `KEYVIZ_FANOUT_NODES`)

1. **Top-level defaults** for all five env vars.
2. **Bool validation** — `KEYVIZ_HOT_KEYS_ENABLED` added to the bool-validated set so a typo'd value (`True`, `yes`, `1`, …) hard-errors at the script level rather than producing an inscrutable failure inside the SSH heredoc.
3. **SSH env forward + `printf '%q'` pre-quoting** for the four integer tuning knobs (matches the `SQS_PORT_Q` / `KEYVIZ_FANOUT_NODES_Q` precedent — anything not bool-validated goes through `%q` before crossing the SSH boundary).
4. **`build_keyviz_flags`** extended to emit `--keyvizHotKeysEnabled` + the four optional knobs only when `KEYVIZ_HOT_KEYS_ENABLED=true` AND each tuning knob's env var is non-empty. Master `KEYVIZ_ENABLED` still gates the whole emission so hot-keys cannot ride without the sampler being on.
5. **`config_fp()`** includes all five env vars in the deploy fingerprint so a hot-keys-only config change correctly triggers a container recreate.

## Test evidence

- `bash -n scripts/rolling-update.sh` clean.
- Extracted `build_keyviz_flags` into an isolated harness and exercised six matrices:
  - master off → no flags
  - master on / hot-keys off → only `--keyvizEnabled`
  - hot-keys on with sampler defaults (empty knobs) → no tuning flags emitted
  - hot-keys on with all knobs set → all flags emitted with the configured values
  - fan-out + hot-keys → both blocks compose correctly
  - master off overriding hot-keys=true → no flags (master-switch wins)

  All six match expected output.
- Bool validation rejects `KEYVIZ_HOT_KEYS_ENABLED='yes'` with the documented error message.

1. **Data loss** — N/A. Deploy-script change.
2. **Concurrency** — N/A.
3. **Performance** — N/A. Disabled-case is one early-return in the helper; no behaviour change for existing deploys where `KEYVIZ_HOT_KEYS_ENABLED` stays unset.
4. **Consistency** — `config_fp` coverage means a hot-keys-only config change triggers a recreate; the bool validation rejects ambiguous values before they reach the daemon; the master-switch dependency (`KEYVIZ_ENABLED` gates hot-keys) matches the Go-side dependency.
5. **Test coverage** — exercised `build_keyviz_flags` directly via an extracted-function harness. The script itself has no Go test surface; integration is covered by the unit-test side that landed in PR #854.

## Deferred (last follow-up)

One PR #854 item remains: **Fan-out merge** (design §6) — handler is single-node v1. A separate PR will replicate the matrix fan-out infrastructure for the hot-keys endpoint.

### DIFF
--- a/scripts/rolling-update.env.example
+++ b/scripts/rolling-update.env.example
@@ -125,3 +125,36 @@ ADMIN_ENABLED="false"
 # full design.
 KEYVIZ_ENABLED="false"
 # KEYVIZ_FANOUT_NODES="10.0.0.1:8080,10.0.0.2:8080,10.0.0.3:8080"
+
+# KeyViz hot-key Top-K drill-down (Phase 2-A++). When
+# KEYVIZ_HOT_KEYS_ENABLED=true (and KEYVIZ_ENABLED is also true), every
+# tracked route grows a Space-Saving sketch that powers the
+# /admin/api/v1/keyviz/hotkeys endpoint and the heatmap cell-click
+# drill-down. Disabled-case overhead on the sampler hot path is one
+# nil-check branch.
+#
+# Tuning (all four are sampler-side clamped to the limits documented
+# in keyviz/sampler.go; an out-of-range value rounds into the band):
+#   KEYVIZ_HOT_KEYS_PER_ROUTE — Space-Saving capacity m per route.
+#       Default 64, max 256. Each route holds at most m keys; with
+#       m=64 a key whose true frequency f > N/64 is guaranteed to
+#       survive eviction. Memory: m × (maxKeyLen + 16) per tracked
+#       route.
+#   KEYVIZ_HOT_KEYS_SAMPLE_RATE — R: the hot path enqueues 1 in R
+#       observes. Default 16, max 1024. Higher R = less CPU on the
+#       hot path + a coarser sketch (the response scales counts back
+#       up by R and surfaces `error_bound`).
+#   KEYVIZ_HOT_KEYS_QUEUE_SIZE — bounded channel between the hot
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

Deploy wiring for KeyViz hot-keys. Mirrors the existing `KEYVIZ_FANOUT_NODES` pattern at all five touchpoints (defaults, bool validation, SSH forward + pre-quote, `build_keyviz_flags`, `config_fp`).

Self-test results in the PR description; six positive matrices verified.

PR #860, opened 2026-05-28
feat(admin): KeyViz heatmap cell click opens hot-key drill-down panel

bootjp:
## What

Closes the last deferred follow-up from **PR #854** (per-cell hot-key Top-K drill-down, Phase 2-A++): SPA cell on-click handler that fetches `/admin/api/v1/keyviz/hotkeys` and renders a ranked Top-K panel below the heatmap.

**Depends on PR #854** — that PR ships the handler this UI calls. Merge order: #854 → this PR.

## Scope

Pure SPA change: two files, no Go changes, no schema change, no new design doc (the design is already in `docs/design/2026_05_28_proposed_keyviz_hot_key_topk.md` §5).

### `web/admin/src/api/client.ts`
- `KeyVizHotKey` / `KeyVizHotKeysResponse` / `KeyVizHotKeysParams` types mirror `internal/admin/keyviz_hotkeys_handler.go::hotKeyResponse`.
- `api.keyVizHotKeys(params, signal)` — passes `route_id`, `sub_bucket`, `series`, `top` through. Deliberately omits `from_unix_ms`/`to_unix_ms` so the panel always reflects "now" and never triggers the `out_of_snapshot_window` 400.

### `web/admin/src/pages/KeyViz.tsx`
- `selectedCell` state on Heatmap, toggled by `onClick`: clicking the same cell twice clears it (no separate close button on the canvas).
- `SelectionOverlay` — 2px outline SVG layered inside the same scroll container as `ConflictOverlay`, so the highlight tracks the canvas horizontally on overflow.
- Matrix-identity effect clears the selection on refresh / series switch / rows resize.
- `parseBucketID` inverts `bucketIDFor` (server-side keyviz_handler.go):
  - `virtual:<routeID>` → aggregate (no drill-down available)
  - `route:<routeID>` → K=1 route, no sub-bucket axis
  - `route:<routeID>#<sub>` → sub-bucket row of a K>1 route
- **HotKeysPanel** surfaces every unsupported branch explicitly so the user never sees a blank box:
  - `series !== "writes"` → static notice (server only samples writes, design §3).
  - virtual aggregate → notice with the routes-coarsened count and a pointer to `--keyvizMaxTrackedRoutes`.
  - 404 `no_snapshot` → "drive some traffic and click again".
  - 503 `hot_keys_disabled` → "start the server with `--keyvizHotKeysEnabled`".
[...]

PR #866, opened 2026-05-29
feat(admin): KeyViz hot-keys cluster-wide fan-out (design §6)

bootjp:
## What

Closes the **fan-out merge** deferred follow-up from PR #854 (KeyViz per-cell hot-key Top-K drill-down, Phase 2-A++). The `/admin/api/v1/keyviz/hotkeys` endpoint now aggregates the local Space-Saving sketch with peer responses, mirroring the matrix fan-out infrastructure.

After this lands, all three deferred items from PR #854 are shipped (SPA cell-click in PR #860, deploy wiring in PR #863, fan-out merge here). The hot-key design doc `docs/design/2026_05_28_proposed_keyviz_hot_key_topk.md` is ready to rename `proposed_` → `implemented_` (proposed for a tiny follow-up housekeeping PR).

## Scope

Pure Go/server-side. Six files: one new fan-out implementation, one new test file, four existing files extended in-place. No schema/wire-format change beyond a new optional `fanout` field on the hot-keys response (parallel to `KeyVizMatrix.fanout`). No SPA change — PR #860 introduces the `KeyVizHotKeysResponse` TS type and a follow-up will add the `fanout?` field there.

### `internal/admin/keyviz_hotkeys_fanout.go` (new)

`KeyVizHotKeysFanout` mirrors `KeyVizFanout`'s structure: peer list + HTTP client + per-peer timeout + body cap; same `With*` options. `Run` merges local with parallel peer fetches and never returns an error (peer failures surface in `FanoutResult.Nodes[].OK=false`).

`mergeHotKeyResponses` applies the design §6 rules:

| Field | Rule | Rationale |
|---|---|---|
| `count` per key | SUM | Responses already carry scaled-to-true estimates from `buildHotKeysResponse` |
| `error_bound` | **SUM**, not MAX | Per-peer bounds compose additively for the sum-of-independent-estimates. MAX would understate the noise floor by up to peer_count× and mislead any client treating it as a confidence interval |
| `sample_rate` | MAX | Coarsest peer dominates the merged accuracy |
| `sampled_n` | SUM | Informational |
| `dropped_samples` | SUM | |
| `skipped_long_keys` | SUM | |
| `degraded` | **OR** | drops>0 OR skips>0 from any peer; a peer that cleanly drained its queue but skipped a too-long hot key still flips degraded so the cluster can't report 'clean' while omitting the cell's hottest entry |
| `snapshot_at` | MAX | User sees the freshest publish across the cluster |
| `keys` | re-sort desc by summed count, tie-break KeyB64 asc, take `params.top` | Determinism for tests + stable post-merge order |

### `internal/admin/keyviz_hotkeys_handler.go`

- `hotKeyResponse.Fanout *FanoutResult` (omitempty) reuses the matrix `FanoutResult` shape so the SPA can render a `FanoutBanner`-style notice with no extra type plumbing.
- `WithFanout` wires the aggregator; `ServeHTTP` delegates to `maybeFanout` which gates on the `X-Admin-Fanout-Peer` anti-recursion header (same rule the matrix fan-out uses). The helper extraction keeps `ServeHTTP` under cyclop's CC limit.

### `internal/admin/keyviz_fanout.go` (extracted helper)

Pulled the HTTP / decode / body-cap boilerplate (req build → cookie attach → recursion header → `client.Do` → body close → status check → JSON decode → over-cap warn) into shared `fetchPeerJSON`. Killed the dupl complaint that the second copy would otherwise trigger, and a future security fix to the peer-HTTP path (tightened body cap, additional header policy, etc.) now applies to both surfaces in one edit.

### `internal/admin/server.go` + `main_admin.go`

`ServerDeps.KeyVizHotKeysFanout` passes through the same wire as `ServerDeps.KeyVizFanout`. `buildKeyVizHotKeysFanout` mirrors `buildKeyVizFanout` (same `keyVizFanoutConfig`, same self-filter rule), so operators get the cluster-wide drill-down for free by leaving `--keyvizFanoutNodes` set — no separate flag, matching design §6.

**Mixed-K** (some nodes hot-keys-enabled, others not): peers returning 503 `hotkeys_disabled` appear as `ok=false` in per-node status and their contribution is simply omitted, matching design §6's *"peers that don't sample omit their contribution"*.

## Tests (`keyviz_hotkeys_fanout_test.go`)

- `TestHotKeysFanout_NoPeers_EchoesLocalWithSelfStatus` — single-node fast path.
- `TestMergeHotKeyResponses_DesignSection6` (7 cases) — every merge rule in the table above, plus the top-K truncation and tie-break determinism.
- `TestBuildKeyVizHotKeysPeerURL` (3 cases) — host:port shorthand → http://, full URL pass-through, sub_bucket + from/to forwarding.
- `TestHotKeysFanout_AntiRecursionHeader` — outgoing peer fetch sets `X-Admin-Fanout-Peer: 1`.
- `TestHotKeysFanout_PeerErrorRecordedNotFatal` — peer 503 records ok=false but local still contributes.
- `TestHotKeysHandler_FanoutIntegration` — end-to-end: handler with httptest peer; merged response contains both contributions with correct SUM.
- `TestHotKeysHandler_FanoutSkippedOnPeerHeader` — inbound peer header suppresses recursive fan-out.

## Verification

- `go build ./...` clean.
- `go test -race -count=3 ./internal/admin/ ./keyviz/` green.
- `golangci-lint ./internal/admin/... ./keyviz/...` → **0 issues**.

1. **Data loss** — N/A. Read-only fan-out of an in-memory sketch.
2. **Concurrency** — per-peer goroutine pool with a `wg` + timeout-bounded `callCtx`; cancellation propagates via the parent ctx; no shared mutable state crosses goroutine boundaries (each writes to its own `results[i]` slot).
3. **Performance** — single fan-out round per inbound request; peer responses bounded at 64 MiB (shared cap with matrix); merge is O(total_keys × log total_keys) for the final sort. Disabled-case (no fanout, or peer-header present) is a single nil-check + header read.
4. **Consistency** — `Approximate` is forced `true` on merge (SS sketches are always approximate; OR semantics); `error_bound` is SUM not MAX to preserve the confidence-interval contract; `degraded` is OR-of-drops-or-skips so neither degradation source alone can silently leave the cluster reporting clean; `SnapshotAt` is MAX so the user sees the freshest publish.
5. **Tests** — seven merge-rule cases, three URL builder cases, four end-to-end fan-out cases (no-peers, anti-recursion outbound, peer 503 mixed-K, handler integration), plus the inbound anti-recursion gate. `-race -count=3` green.

## Summary by CodeRabbit

* **New Features**
  * Cluster-wide hot-keys aggregation for the admin hot-keys drill-down, merging results from peers to show comprehensive top-K across the cluster with deterministic ordering and correct top-K promotion.

* **Bug Fixes / Resilience**
  * Fan-out tolerates individual peer failures (records per-node status but returns merged data when available) and fills missing local snapshots from peers when possible.

* **Behavior**
  * Fan-out is protected against recursive calls to avoid loops.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/866?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bootjp:
@claude review

Hot-keys cluster-wide fan-out (last PR #854 deferred follow-up).

Mirrors KeyVizFanout: same peer-list construction, same anti-recursion header, same cookie forwarding, same per-peer body cap. Refactored the matrix fan-out's fetchPeer to share `fetchPeerJSON` so the body-cap + over-cap warn plumbing has one home.

Design §6 merge implemented: SUM counts/bounds/sampled_n/drops/skips, MAX sample_rate, OR degraded, MAX snapshot_at. error_bound is deliberately SUM not MAX (preserves the confidence-interval contract); degraded ORs drops OR skips (a peer that drained cleanly but skipped a too-long hot key still flips degraded).

Independent of PR #860 (SPA cell-click) and PR #863 (deploy wiring) — none depend on each other. After all three land, the design doc renames `proposed_` → `implemented_` in a small housekeeping PR.

bootjp:
@claude review

PR #866 round-1 review feedback addressed in 745e7397.

- 🟠 **codex P1** (peer top truncation) → `buildKeyVizHotKeysPeerURL` now requests `keyviz.MaxHotKeysPerRoute` (256) from every peer instead of forwarding `params.top`. Each peer returns every locally tracked candidate; the merger applies the operator's `top` after summing.
- 🔴 **gemini HIGH** (early 404 prevents fan-out) → ServeHTTP no longer 404s on nil local snapshot when fan-out is configured + not a peer call. Empty local placeholder feeds the merger; post-fan-out check downgrades back to 404 only when even peers had no data. Three new regression tests pin both directions.
- 🚫 **reviewdog gci** (`main_admin.go`) → struct-field alignment auto-fixed.
- Claude bot non-blocking minors (all addressed): atomic.Int32 for peerHits, ErrorBound integration assertion (locks down SUM-not-MAX end-to-end), seeding-pattern comment in mergeHotKeyResponses.

Caller audit: `buildLocalResponse` is new; its single caller (ServeHTTP) handles every (errStatus, errCode, errMsg) tuple uniformly. Public API unchanged.

Test evidence: `go build ./...` clean / `go test -race -count=3 ./internal/admin/ ./keyviz/` green / `golangci-lint` 0 issues.

bootjp:
@claude review

Round-2 codex P1 (defer local top truncation until after fan-out merge) addressed in 61cefea7.

The round-1 fix sent every PEER the SS sketch ceiling, but the LOCAL node's collectEntries was still truncating to `params.top` before the merger. Same correctness bug, peer side closed but local side still leaking. `buildLocalResponse` now overrides `localParams.top = keyviz.MaxHotKeysPerRoute` when fan-out is configured + not a peer call. Single-node and peer-call paths keep the operator's top so behaviour there is unchanged.

New regression test `TestHotKeysHandler_LocalTopDeferredUntilAfterFanoutMerge` exercises the textbook failure scenario (local x=100/z=99 + peer z=99, top=1) and asserts the merged top-1 is z(1683), not x(1600).

Caller audit: collectEntries has one caller (buildLocalResponse). Semantic change is local-only; no external API change.

bootjp:
@claude review

Codex P2 round-3 (mixed-K: allow fan-out before declaring local hotkeys disabled) addressed in 6387a2cf.

The fix is symmetric with the gemini HIGH round-1 fix for nil snapshot: keep the 503 hotkeys_disabled only when fan-out cannot reach peers (single-node OR this request is itself a peer call). When fan-out CAN reach peers and the local node doesn't sample, build an empty local placeholder and let the merger overlay peer contributions. The existing post-fan-out empty-cluster check downgrades to 404 no_snapshot if no peer had data either.

Extracted `prepareRequest` to keep ServeHTTP under cyclop's CC limit. Three regression tests pin:
- mixed-K !enabled-local + peer-with-data → 200 with peer data + SampledN + ErrorBound surfaced.
- !enabled-local + no fan-out → 503 still fires (single-node operator still sees the config error).
- !enabled-local + inbound peer-header → 503 without fan-out (recursion guard preserved).

Caller audit: 503 codes only produced by this handler; SPA's HotKeysErrorNotice handles both 503 codes for flag hints but does not depend on 503 firing for any specific topology. The behavioural change (mixed-K → 200) is the intended fix.

bootjp:
@claude review

Codex P2 round-4 (preserve empty peer snapshot instead of 404) addressed in e8d8cd4b.

Fix: use `out.SnapshotAt.IsZero()` as the no-source-published-anything signal. The merger sets SnapshotAt to MAX across responses, so non-zero means at least one source contributed a valid snapshot — even one with `keys=[]` and `sampled_n=0`. Matches the single-node path's 200-with-empty-body semantics for tracked-but-quiet routes.

- All-peers-failed (404/timeout) case: SnapshotAt stays zero → 404 (unchanged).
- Peer returns valid empty snapshot: SnapshotAt non-zero → 200 with peer's SnapshotAt preserved.

New regression test `TestHotKeysHandler_FanoutNilLocalPeerHasEmptySnapshot` pins the new behaviour; `TestHotKeysHandler_FanoutNilLocalAllPeersEmpty` still passes (peer 404 → ok=false → no merge contribution → SnapshotAt zero → 404).

Caller audit: change is local to ServeHTTP's post-fan-out decision.

PR #867, opened 2026-05-29
kv(hlc): add NextFenced() for HLC-4 (iii) ceiling fence; migrate kv coordinator paths

bootjp:
## Summary

- Splits HLC timestamp allocation into two APIs so the HLC-4 (iii) fail-closed contract from the TLA+ proof can be enforced at exactly the sites that persist a ts, without the multi-file blast radius of changing `Next()`'s signature project-wide.
  - `Next() uint64` — unchanged signature. Bypasses the physical-ceiling fence. Used by diagnostics / identifiers / retry IDs / demo wiring / adapter sites whose fail-closed migration is deferred to Phase 2.
  - `NextFenced() (uint64, error)` — new. Enforces `wall_now < physicalCeiling` (HLC-4 (iii)) and returns `ErrCeilingExpired` instead of issuing a ts inside a stale leader window after RunHLCLeaseRenewal stops. Pre-bootstrap (ceiling == 0) is intentionally soft.
- Migrates the kv-layer persistence-allocation paths to `NextFenced`, propagating the error with `errors.Wrap`:
  - `kv/coordinator.go` — `nextStartTS`, `dispatchTxn` commit-ts (with `Observe`-and-retry path extracted into `resolveDispatchCommitTS` for nestif), `dispatchRaw`, `toRawRequest`.
  - `kv/sharded_coordinator.go` — `dispatchDelPrefixBroadcast`, `nextTxnTSAfter`, `nextStartTS`, `rawLogs`.
  - `adapter/internal.go` — `stampRawTimestamps`, `stampTxnTimestamps`, `fillForwardedTxnCommitTS` (allocation extracted into `forwardedTxnCommitTS` for cyclop).
- `adapter/s3.go`, `adapter/redis.go`, `adapter/redis_compat_commands.go` direct callers intentionally stay on `Next()` and are tagged for **Phase 2** follow-up. The kv-layer migration alone makes every `ShardedCoordinator`-routed and `Coordinator`-routed commit fail-closed when the ceiling expires; the remaining adapter sites can be migrated per-adapter without coordinated churn.

## Spec linkage

- `docs/design/2026_05_28_partial_tla_safety_spec.md` §5.1 — HLC-4 precondition (iii).
- `tla/hlc/MCHLC_gap.cfg` — TLA+ counterexample at depth 5 demonstrating the safety property this enforces.

## Self-review (5 lenses)

[...]

PR #869, opened 2026-05-29
adapter(s3): migrate direct HLC.Next() callers to NextFenced — PR #867 Phase 2a

bootjp:
## Summary

PR #867 Phase 2a — extends the HLC-4 (iii) fail-closed contract from the kv layer (#867 Phase 1) into the **S3 adapter's direct timestamp-allocation sites**.

The s3 adapter has two private helpers that draw from `HLC` without going through the kv coordinator:

- `txnStartTS(readTS)` — when `readTS == ^uint64(0)` (sentinel "server, please allocate"), called by every SigV4 mutation entry point and the admin RPCs.
- `nextTxnCommitTS(startTS)` — allocates the commit-ts strictly above `startTS` for multipart upload and PUT.

Both now use `clock.NextFenced()` and propagate `ErrCeilingExpired` instead of silently issuing a stale-leader timestamp.

### What changed

- `txnStartTS` signature: `func(readTS uint64) uint64` → `func(readTS uint64) (uint64, error)`. Sentinel-path uses `NextFenced`; non-sentinel path (caller-supplied `readTS`) is unchanged.
- `nextTxnCommitTS`: swapped `clock.Next()` for `clock.NextFenced()` (already returned `(uint64, error)`).
- 14 caller sites plumbed for the new error:
  - 9 in `s3.go` (createBucket, deleteBucket, putBucketAcl, putObject, deleteObject, createMultipartUpload, uploadPart, completeMultipartUpload, abortMultipartUpload).
  - 5 in `s3_admin.go` / `s3_admin_objects.go` (adminCreateBucketTxn, adminUpdateBucketAcl, AdminDeleteBucket, adminDeleteObjectTxn, adminPutObjectStream).
- Extracted `AdminDeleteBucket`'s retry closure into `adminDeleteBucketTxnBody` to keep the function under the cyclop ceiling after the new error branch.

### What is intentionally NOT migrated

- `newS3UploadID` continues to use plain `Next()`. The function returns an opaque multipart-upload identifier that is only used as a lookup key; it is not a persistence timestamp and bypasses MVCC/OCC/lease semantics. An inline comment was added so a future audit does not mistakenly migrate it.
- `adapter/redis.go`, `adapter/redis_compat_commands.go` — **Phase 2b**, a follow-up PR after this lands.

### Regression tests (added before the fix per CLAUDE.md)

- `TestS3TxnStartTSFailsClosedOnExpiredCeiling` — sentinel path propagates `ErrCeilingExpired`.
- `TestS3TxnStartTSPassesThroughExplicitReadTS` — non-sentinel path is unaffected by an expired ceiling (pin-existing-readTS contract preserved).
- `TestS3NextTxnCommitTSFailsClosedOnExpiredCeiling` — commitTS allocation propagates `ErrCeilingExpired`.

All three verified failing on the parent commit (compile-fail before the migration), passing on this commit.

### Self-review (5 lenses)

1. **Data loss** — Fenced path returns `(0, ErrCeilingExpired)` with no silent fallback. The SigV4 layer surfaces it as HTTP 5xx via the existing `writeS3InternalError` path; the admin RPC surfaces it as an error to the operator.
2. **Concurrency / distributed failures** — Both helpers are stateless wrappers around `HLC.NextFenced`, which retains the same CAS loop with the fence checked inside the loop (concurrent `SetPhysicalCeiling` between Load and CAS retries correctly).
3. **Performance** — One extra atomic load + branch per allocation; identical to the kv-layer cost from #867. No additional Raft round-trip.
4. **Data consistency** — HLC-4 (iii) ceiling fence now applies on the s3 adapter's direct-allocation path, aligning with the kv-layer migration. Combined with #867, every coordinator-routed AND every adapter-direct S3 commit fails closed when the leader's lease renewal stops.
5. **Test coverage** — 3 new unit tests added co-located in `adapter/s3_hlc_fence_test.go`. Existing s3 lifecycle tests (`TestS3Server*` suite) unchanged.

## Test plan

- [x] `go build ./...`
- [x] `go test -race -count=1 -run 'TestS3' ./adapter`
- [x] `make lint` (golangci-lint) — 0 issues
- [ ] Full `go test -race ./adapter` in CI (local suite is slow; CI is the source of truth)
- [ ] Phase 2b: redis adapter migration in a separate PR after this lands

### DIFF
--- a/adapter/s3.go
+++ b/adapter/s3.go
@@ -563,7 +563,10 @@ func (s *S3Server) createBucket(w http.ResponseWriter, r *http.Request, bucket s
 	}
 	err := s.retryS3Mutation(r.Context(), func() error {
 		readTS := s.readTS()
-		startTS := s.txnStartTS(readTS)
+		startTS, err := s.txnStartTS(readTS)
+		if err != nil {
+			return errors.Wrap(err, "s3: allocate startTS for mutation")
+		}
 		readPin := s.pinReadTS(readTS)
 		defer readPin.Release()
 
@@ -639,7 +642,10 @@ func (s *S3Server) deleteBucket(w http.ResponseWriter, r *http.Request, bucket s
 	var deletedGeneration uint64
 	err := s.retryS3Mutation(r.Context(), func() error {
 		readTS := s.readTS()
-		startTS := s.txnStartTS(readTS)
+		startTS, err := s.txnStartTS(readTS)
+		if err != nil {
+			return errors.Wrap(err, "s3: allocate startTS for mutation")
+		}
 		readPin := s.pinReadTS(readTS)
 		defer readPin.Release()
 
@@ -784,7 +790,10 @@ func (s *S3Server) putBucketAcl(w http.ResponseWriter, r *http.Request, bucket s
 
 	err := s.retryS3Mutation(r.Context(), func() error {
 		readTS := s.readTS()
-		startTS := s.txnStartTS(readTS)
+		startTS, err := s.txnStartTS(readTS)
+		if err != nil {
+			return errors.Wrap(err, "s3: allocate startTS for mutation")
+		}
 		readPin := s.pinReadTS(readTS)
 		defer readPin.Release()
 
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

PR #870, opened 2026-05-29
docs(design): propose Composed-1 cross-group commit-time ownership guard

bootjp:
## Summary

Forward-looking design proposal for the **Composed-1 cross-group commit-time ownership guard**, the implementation that would be required *before* introducing any cross-group route mutation (cross-group SplitRange, MoveRange, group decommission).

Today the implementation is vacuously safe with respect to Composed-1 (proved in `tla/composed/Composed.tla` / `MCComposed.cfg`) because `SplitRange` is same-group only and no other route-mutating RPC exists. This doc designs the apply-time check that any cross-group RPC would need to land alongside — review-and-ready instead of patched in under time pressure.

## What the doc covers

- The Composed-1 invariant from `Composed.tla` and the canonical 4-step gap counterexample at depth 4.
- Three structural facts that make today's implementation vacuously safe (same-group-only SplitRange; no MoveRange/Rebalance; leader-side-cached routes).
- The hazard if cross-group route mutation is introduced — G1c-style lost-write across a group boundary, invisible to current Jepsen workloads because the RPC surface doesn't expose the trigger.
- Proposed plumbing:
  - `ObservedRouteVersion` field on `OperationGroup` propagated through `dispatchTxn` into `pb.Request`.
  - FSM `verifyComposed1` at apply time + new `ErrComposed1Violation` sentinel.
  - Coordinator retry on the sentinel (re-read routes, re-route txn, retry once).
  - Catalog version retention ring in `distribution/Engine` (default depth 32; soft-failure on miss).
- TLA+-to-implementation correspondence — the apply check directly refines the spec's `Commit` precondition.
- Five milestones M1..M5 (plumbing → ring → check → retry → Jepsen workload), each independently testable.
- Five-lens self-review per CLAUDE.md.
- Three open questions (deprecation of unpinned txns; routeHistoryDepth default; hard-vs-soft missing-version behaviour).

## Why now, why this PR

CLAUDE.md mandates "design-doc-first" for changes touching routing, OCC, or FSM apply semantics. The implementation does not land with this PR — only the proposal. Landing the doc first lets the eventual implementation PR cite a reviewed contract instead of inventing one inline.

## Test plan

- [ ] Reviewer cross-checks §4.2 (FSM check) against the spec's `Commit` predicate in `tla/composed/Composed.tla`.
- [ ] Reviewer confirms the milestone ordering allows M1+M2 to land independently without a behaviour change.
- [ ] Reviewer agrees with the three open questions or proposes alternatives.
- [ ] No code changes expected from this PR.

### DIFF
--- a/docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md
+++ b/docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md
@@ -1,0 +1,298 @@
+# Composed-1 — cross-group commit-time ownership guard
+
+Status: Proposed
+Author: bootjp
+Date: 2026-05-29
+
+> **Forward-looking proposal.** Today's implementation is vacuously
+> safe with respect to Composed-1 because `SplitRange` is the only
+> route-mutating control-plane RPC and it is same-group only (per
+> CLAUDE.md). This doc designs the commit-time ownership guard that
+> would be required *before* introducing any cross-group route
+> mutation (cross-group `SplitRange`, `MoveRange`, key-range
+> rebalancing). The implementation is not justified by current
+> deployment state; the proposal exists so that when a follow-up
+> introduces cross-group movement, the guard is reviewed-and-ready
+> rather than patched in under time pressure.
+
+---
+
+## 1. Background — Composed-1 from the TLA+ spec
+
+The TLA+ Composed module models the cross-module seam between OCC and
+the route catalog. Composed-1 (proved in
+`tla/composed/Composed.tla` / `MCComposed.cfg`) states:
+
+> Every committed write key was owned by the committing group at the
+> transaction's observed catalog version.
+
+Formally:
+
+```
+Composed1_CommitToOwningGroup ==
+  \A t \in {tx \in TxnIds : txnState[tx] = "Committed"} :
+    \A k \in txnWriteSet[t] :
+      routes[txnObservedVer[t]][k] = txnCommitGroup[t]
+```
+
+The gap config (`MCComposed_gap.cfg`) drops the
+`routes[v][k] = g` precondition from `Commit` and surfaces the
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

bootjp:
All four review comments addressed in 124a36e0:

**codex P1 — late commit to old owner**. Single biggest finding — the first revision's observed-version-only check refines the spec's literal `Commit` predicate but does NOT prevent the §3 MoveRange G1c trace (a commit lands on `g1` because `routes[v_obs][k1] = g1` passes observed-check, while readers at `v_obs+1` route to `g2` and miss it). Added §4.4 specifying a **current-version cross-version-read fence** the FSM also runs alongside the observed-version check at apply time. Updated §3 to call out the gap explicitly and §5 to flag the apply-time fence as strictly stronger than the spec predicate (with extending Composed.tla to model `currentVersion` noted as a follow-up).

**gemini medium #2 + codex P2 — fail closed on GCd version**. §4.2 / §4.3 now return a new `ErrComposed1VersionGCd` sentinel on `SnapshotAt(observedVer) == nil`. Coordinator's retry path handles it identically to `ErrComposed1Violation`. Retention depth is now a liveness knob, not a safety knob.

**gemini medium #1 — kvFSM fields don't exist**. Added explicit prerequisite comment block in §4.2 covering the struct extension (routes kv.RouteHistory + shardGroupID uint64). Merged the engine ring and kvFSM extension into a single M2 milestone since they're tightly coupled through constructor wiring.

**Self-review / open questions**. §8 lens 1 updated to describe the three-way guard. §9 Q2 clarified as liveness-only. §9 Q3 replaced — soft-failure question is resolved (hard error); new Q3 is apply-time fence vs. MoveRange drain protocol, pending real operational data.

@claude review

PR #881, opened 2026-05-29
kv(composed1): M1 plumbing — ObservedRouteVersion in OperationGroup → pb.Request

bootjp:
## Summary

First milestone (**M1**) of the Composed-1 cross-group commit-time guard per [`docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md) §M1.

Plumbs a new `ObservedRouteVersion` field end-to-end so the M3 FSM apply-time gate (separate PR) can read it without further plumbing churn. **Behaviour-neutral**: every existing caller leaves the field at its zero value, and the FSM at this milestone ignores it.

## What changes

- **`proto/internal.proto`** — adds `uint64 observed_route_version = 6` on `pb.Request` with an inline comment pointing at the design doc. Regenerated `proto/internal.pb.go` via `protoc 29.6` + `protoc-gen-go v1.36.11`.
- **`kv/transcoder.go`** — adds `ObservedRouteVersion uint64` to `OperationGroup` with a doc comment that explicitly calls out the zero default and the M3 follow-up.
- **`kv/coordinator.go`** —
  - `onePhaseTxnRequest` gains the `observedRouteVersion uint64` param and sets it on the returned `pb.Request`.
  - `Coordinate.dispatchTxn` gains the param and forwards it from the calling `OperationGroup`.
  - `Coordinate.Dispatch` already routes through `dispatchTxn`, so the OperationGroup field flows in.
  - `Coordinate.buildRedirectRequests` (follower→leader forward path) reads `reqs.ObservedRouteVersion` so the field survives a redirected commit too.
- **`kv/sharded_coordinator.go`** — `ShardedCoordinator.dispatchTxn` and `dispatchSingleShardTxn` gain the same param, propagated from the calling `OperationGroup`.

## What does NOT change

- The FSM (`kv/fsm.go`) does **not** read `pb.Request.ObservedRouteVersion`. That's the M3 milestone.
- No new errors, no new sentinels, no new control-flow branches on the hot path.
- No adapter-side change: every adapter calls `Dispatch` with an OperationGroup whose `ObservedRouteVersion` is zero by default, which the M3 gate will treat as "unpinned" (legacy) and skip.

## Tests

- `TestCoordinateDispatchTxn_PassesObservedRouteVersionToRaftEntry` — the round-trip witness: `dispatchTxn(..., observedRouteVersion=42)` produces `pb.Request.ObservedRouteVersion == 42`. This is the M1 "done when" criterion from §M1.
- `TestCoordinateDispatchTxn_UnpinnedObservedRouteVersionStaysZero` — legacy-default witness: `dispatchTxn(..., observedRouteVersion=0)` produces `pb.Request.ObservedRouteVersion == 0`.
- Existing dispatch tests updated for the new signature (one extra trailing `0` per call site).

## Verification

- `go build ./...` — clean
- `go test -race -count=1 ./kv` — 10.0 s, pass
- `make lint` — 0 issues

## Self-review (5 lenses)

1. **Data loss** — no new write paths; field is plumbed but not yet read by the FSM. Cannot lose a write because no new branch could refuse one.
2. **Concurrency** — additive; no new locks, no new goroutines.
3. **Performance** — one extra uint64 field on every `pb.Request`; one extra uint64 param on the dispatch path. No new atomics, no new allocations.
4. **Data consistency** — the field flows OperationGroup → dispatchTxn → onePhaseTxnRequest → pb.Request; M3 will close the loop at FSM apply.
5. **Test coverage** — round-trip + legacy-default tests cover the M1 contract.

## Test plan

- [x] `go test -race ./kv`
- [x] `make lint`
- [ ] Reviewer confirms the proto regen produces only the expected diff (one field added; one Getter; one descriptor entry) — `git diff proto/internal.pb.go` should be reviewable as a normal proto field addition
- [ ] (Follow-up PRs) **M2** — catalog version retention ring in `distribution/Engine` + `kvFSM` struct extension (`routes`, `shardGroupID`). **M3** — FSM apply-time `verifyComposed1` + coordinator retry on `ErrComposed1Violation`. **M4/M5** — Jepsen workload.

## Resolves

The M1 row in the Composed-1 design doc.

## Release Notes

* **Chores**
  * Updated internal transaction coordination and dispatch mechanisms to track and propagate route version information consistently across all transaction lifecycle phases, from preparation through final commit operations.

* **Tests**
  * Added regression tests validating that route version information is properly propagated and preserved through transaction dispatch operations in both single-shard and multi-shard transaction scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/881?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


bootjp:
@claude review

bootjp:
Gemini critical addressed in the latest commit:

The single-shard fast path was forwarding ObservedRouteVersion correctly via `onePhaseTxnRequest`, but the multi-shard 2PC path (`prewriteTxn` / `commitPrimaryTxn` / `commitSecondaryTxns`) and the `buildTxnLogs` helper silently dropped the field — every cross-shard PREPARE and COMMIT envelope arrived at the FSM with ObservedRouteVersion = 0, silently bypassing the M3 apply-time gate exactly for the workload most at risk of a cross-group route shift.

Fixed all 5 sites per gemini's enumeration:
- `prewriteTxn` — PREPARE pb.Request gains ObservedRouteVersion
- `commitPrimaryTxn` — primary COMMIT gains it
- `commitSecondaryTxns` — secondary COMMIT gains it
- `buildTxnLogs` — both PREPARE and COMMIT it builds gain it
- `dispatchTxn` (3 call sites) + `txnLogs` forward observedRouteVersion from the OperationGroup

Regression test added before the fix per CLAUDE.md convention:
`TestShardedCoordinatorDispatchTxn_CrossShardPropagatesObservedRouteVersion` — routes 2 PUTs across 2 shards, sets ObservedRouteVersion=42 on the OperationGroup, asserts every PREPARE+COMMIT envelope across both shards carries the pinned value. Verified failing on the parent commit (0x0 vs 0x2a, error message identified the leaking PREPARE phase) and passing now.

@claude review

PR #884, opened 2026-05-30
feat(txn-dedup): M3 EXEC reuse — option-2 dedup for MULTI/EXEC

bootjp:
## Summary

Follow-up to #796 (M1 + M2 + M3 RPUSH/LPUSH). Extends option-2 write-set reuse + exact-ts dedup probe to **MULTI/EXEC** dispatched through `runTransaction`. Mirrors `listPushCoreWithDedup` at the EXEC granularity. Gated on the existing `onePhaseTxnDedup` flag — **default off** — so no mixed-version divergence window.

This PR is **stacked on #796**: the base branch is `docs/txn-idempotency-design`. Once #796 merges this rebases onto `main`.

## Mechanism

1. **`txnContext.commit()` refactor** — split build-and-dispatch into `prepareDispatch()` returning a `preparedTxnDispatch{elems, commitTS, readKeys, ctx, cancel}`. `commit()` now calls `prepareDispatch()` then dispatches; `runTransactionWithDedup` calls them separately so it can intercept between prepare and dispatch. External behavior of `commit()` is unchanged.
2. **`reusableExecTxn`** — EXEC analogue of `reusableListPush`. The cached `results` array is the M3 R1 result reconstruction: computed once from attempt 1's `startTS` snapshot, returned as-is on any reuse. Same invariance argument as RPUSH/LPUSH `length`.
3. **`dispatchExecReuse`** — one reuse iteration. Dispatches with `PrevCommitTS=pending.commitTS`. On `WriteConflict`, the self-inflicted-conflict guard probes `CommittedVersionAt(probeKey, freshCommitTS)`; if hit, return cached results (codex P1 round-10 class). Otherwise drop pending.
4. **`runTransactionWithDedup` + `firstExecAttempt`** — the option-2 retry loop. First iteration builds the txn and dispatches; retryable failure stashes pending; subsequent iterations call `dispatchExecReuse` with `PrevCommitTS`.
5. **`runTransaction` gate** — when `onePhaseTxnDedup` is on, route to `runTransactionWithDedup`; otherwise keep the legacy retry loop byte-identical.

- `prepareDispatch` (new): callers are `commit()` and `firstExecAttempt`; both honor the `defer prepared.cancel()` contract.
- `commit()`: internal structure changed; external behavior preserved (no test directly invokes it).
- `runTransactionWithDedup` / `firstExecAttempt` / `dispatchExecReuse` / `reusableExecTxn`: all new symbols, exercised only from the gated `runTransaction` path.

## Tests (`adapter/redis_exec_dedup_test.go`)

| Test | Scenario | Pins |
|---|---|---|
| `TestExecDedup_LandedPriorAttempt_ReturnsCachedResults` | Attempt 1 lands then errors | Reuse FSM probe hits; cached "OK" returned; value matches |
| `TestExecDedup_PriorAttemptDidNotLand_Applies` | Attempt 1 pre-rejects | Reuse applies fresh; results match |
| `TestExecDedup_GenuineConflictRebuildsAndApplies` | Concurrent SET advances key past `pending.startTS` | Reuse OCC-conflicts; probe misses; pending dropped; fresh attempt succeeds |
| `TestExecDedup_SelfInflictedReuseConflict_ReturnsSuccess` | Reuse lands then surfaces WriteConflict | Self-conflict guard probes fresh `commitTS`; cached results returned (no double-apply) |
| `TestExecDedup_DisabledKeepsLegacyPath` | Gate off | No probe; same result as legacy `runTransaction` |

## Validation

- `go test ./adapter/ -run 'Txn|MULTI|EXEC|Dedup|TxnStartTS'` passes
- `go test ./kv/ ./store/` both pass
- `gofmt`, `go vet`, `golangci-lint run` all clean (0 issues across `adapter/kv/store`)

## Scope

- **Single-mop EXEC** is the conservative scope per the design doc's "Open questions". The mechanism (cache results array; OCC fence on `readKeys`) works the same for multi-mop EXEC under the existing proof, but the test matrix doubles, so multi-mop validation is a follow-up.
- Per-command reconstruction hooks for SET/INCR/HSET etc. when dispatched **outside** MULTI are still a follow-up; those go through individual command paths, not `runTransaction`.

## Design doc reference

`docs/design/2026_05_21_proposed_txn_secondary_idempotency.md` §M3.

### DIFF
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -2654,7 +2654,30 @@ func (t *txnContext) validateReadSet(ctx context.Context) error {
 	return nil
 }
 
-func (t *txnContext) commit() error {
+// preparedTxnDispatch is the fully-assembled write set + read set + commit
+// timestamp for a MULTI/EXEC transaction, ready to be passed to
+// coordinator.Dispatch. Split out from commit() so the option-2 dedup
+// path (runTransactionWithDedup) can intercept between prepare and
+// dispatch — it needs to capture (elems, commitTS, readKeys) for a
+// possible retry under PrevCommitTS without otherwise duplicating the
+// commit-building logic. The owned ctx is the redisDispatchTimeout-
+// bounded context the caller must run Dispatch under and Cancel after.
+type preparedTxnDispatch struct {
+	elems    []*kv.Elem[kv.OP]
+	commitTS uint64
+	readKeys [][]byte
+	ctx      context.Context
+	cancel   context.CancelFunc
+}
+
+// prepareDispatch builds everything Dispatch needs (elems, commitTS,
+// readKeys, ctx) without actually calling Dispatch. Callers must always
+// invoke `cancel()` on the returned prepared value once the dispatch
+// attempt finishes (commit() does this via defer; the dedup path does it
+// per retry iteration). When the transaction has no writes this returns
+// a prepared value with empty `elems` and a no-op cancel — callers can
+// check len(prepared.elems)==0 and skip the dispatch.
+func (t *txnContext) prepareDispatch() (preparedTxnDispatch, error) {
 	elems := t.buildKeyElems()
 
 	// Pre-allocate commitTS so Delta keys can embed it in their bytes before
@@ -2663,7 +2686,7 @@ func (t *txnContext) commit() error {
 	listElems := t.buildListElems(commitTS)
 	zsetElems, err := t.buildZSetElems(commitTS)
 	if err != nil {
-		return err
+		return preparedTxnDispatch{cancel: func() {}}, err
 	}
 	// TTL elements: string keys have TTL embedded in value (buildKeyElems handles that),
 	// non-string keys get a !redis|ttl| element written in the same transaction.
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

Initial request — PR is stacked on #796 (base branch `docs/txn-idempotency-design`). M3 EXEC reuse extends option-2 dedup to MULTI/EXEC mirroring `listPushCoreWithDedup`.

Status:
- mergeStateStatus: CLEAN
- gemini-code-assist: no findings
- codex-connector: quota exhausted (no review)
- coderabbitai: skipped (non-default base)
- go test ./adapter/ -run 'Txn|MULTI|EXEC|Dedup|TxnStartTS' passes
- go test ./kv/ ./store/ both pass
- gofmt + go vet + golangci-lint clean (0 issues)

Please verify:
1. The `commit()` → `prepareDispatch()` + `dispatch` split is externally indistinguishable (defer-cancel discipline, error mapping, single-shot ctx).
2. `reusableExecTxn.results` caching is safe when the FSM probe hits (cached results describe attempt 1's intent against attempt 1's snapshot — invariant under reuse).
3. The self-conflict guard in `dispatchExecReuse` matches PR #796 round-11's contract.
4. `runTransaction` gate (`onePhaseTxnDedup` on/off) keeps the legacy path byte-identical when off.

PR #846, opened 2026-05-26
backup: Phase 0b M5-1 — SQS queue + message reverse encoder

bootjp:
## Summary

**Phase 0b M5-1 — SQS queue + message reverse encoder.** First slice of the SQS reverse encoder. Inverse of the SQS decoder in `sqs.go`.

- `_queue.json` → `!sqs|queue|meta|` (`storedSQSMetaMagic` + live `sqsQueueMeta` JSON) + `!sqs|queue|gen|` (decimal `"1"`).
- `messages.jsonl` → one `!sqs|msg|data|` record per line (`storedSQSMsgMagic` + live message JSON). The dump body (human-readable string or `{base64}` envelope) is translated to the live value's base64-std `[]byte`; message attributes pass through as raw JSON.
- FIFO queues additionally emit `!sqs|queue|seq|` = `max(sequence_number)+1`.

> Independent of the S3 chain (#842/#845) — targets `main`, reuses only main-available helpers.

### Reconstruction (Option-B, matching DDB/S3)
Uniform `sqsRestoreGeneration=1` in the meta, gen counter, message keys, and each message's `queue_generation`. Visibility state restores **zeroed** (every message visible), matching the decoder default. Generation/incarnation, created-at timestamps, tags, throttle config are not in the dump and restore to zero.

### Scope
Classic single-partition queues. **HT-FIFO partitioned queues** (`partition_count > 1`) use a different key family and **fail closed** (`ErrSQSEncodeUnsupportedPartitioned`). Derived side records (`!sqs|msg|{vis,byage,dedup,group}`) are the follow-up slice **M5-2** (the design's side-record decision gate).

Adds `sqsMessageBody.UnmarshalJSON` (symmetric with its `MarshalJSON`, additive — the decoder never unmarshals it) so the encoder can read `messages.jsonl` bodies back. Reads go through the `sqs/` `os.Root` with the `Lstat` + `IsRegular` + `refuseHardLink` guard.

## Self-review (5 lenses)
- **Data loss**: round-trip via the **real** `DecodeSnapshot` confirms meta + message + body fidelity; fail-closed on bad JSON / non-regular / partitioned / unknown version.
- **Concurrency/distributed**: single-goroutine, one builder.
- **Performance**: streamed JSONL decode.
- **Consistency**: key/value formats reproduced against the live store (prefixes, base64url + BE-u64 keys, magics, live JSON shapes); generation uniform per Option B; FIFO seq restored to max+1.
[...]

PR #885, opened 2026-05-30
docs(backup): propose Phase 0b M5-2 — SQS side-record derivation

bootjp:
## Summary
- Resolves the SQS side-record decision gate that M5-1 (#846) explicitly deferred. M5-1 emits queue meta + messages + counters; **without M5-2's side records, restored FIFO queues redeliver acked messages and the reaper silently never expires retention.**
- Picks **full reconstruction** over lazy-rebuild — the lazy path is rejected because dedup loss is silent, byage loss breaks retention silently, and the lazy work equals the offline work anyway.
- Specifies key shapes / values for vis / byage / dedup / group across both classic AND partitioned-FIFO families, sourced from `adapter/sqs_keys.go` constructors.
- Lays out the cross-check test plan (mirrors M3b-3 GSI's pattern) plus end-to-end restore tests.

## Why doc-first
Per CLAUDE.md "Design-doc-first workflow" — any modification touching SQS / wire-format semantics needs a reviewed `*_proposed_*.md` before implementation. M5-2 touches a derived-key contract the live adapter depends on, so the spec lands first; implementation is a separate PR after this is accepted.

## Test plan
- [x] Doc reviewable on its own — references concrete code locations (`adapter/sqs_keys.go` prefixes + constructors, `adapter/sqs_fifo.go` send path) for every claim.
- [x] Decision gate explicit (recommended vs rejected), with the same "fallback is not zero-cost transparency" framing the parent design doc uses.
- [ ] Implementation lands in a follow-up PR (single slice; the families are tightly coupled so splitting them would create poison-window risks during merges).

### DIFF
--- a/docs/design/2026_05_30_proposed_sqs_side_record_derivation.md
+++ b/docs/design/2026_05_30_proposed_sqs_side_record_derivation.md
@@ -28,12 +28,15 @@ For every `!sqs|msg|data|...` record M5-1 already writes, the encoder must also 
 | `ExpiresAtMillis` | `expires_at_millis`  | `SendTimestampMs + sqsFifoDedupWindowMillis` (`= sendTs + 5 minutes`)  |
 | `OriginalSequence`| `original_sequence`  | `sequence_number` from `messages.jsonl`                                |
 
-**Source-file attribution.** The constructors come from two files, not one:
+**Source-file attribution** (M5-2 is classic-only — `*Dispatch` wrappers in `adapter/sqs_keys_dispatch.go` route between classic and partitioned constructors and are NOT needed here; M5-3 will pick them up):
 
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

addressed claude review 1 (06:08):
- vis key attribution corrected (sqs_messages.go, not sqs_keys.go)
- dedup field derivation completed (4 fields with messages.jsonl mapping)
- partitioned-FIFO scoped out of M5-2 (deferred to M5-3); test plan adjusted
- dedup 5-minute-window caveat folded into rationale
- pipe-notation footnote added under families table

addressed gemini critical (line 37) in earlier push: group rows now explicitly NOT emitted (live loadFifoGroupLock treats key presence alone as held).

bootjp:
@claude review

v4: per-file attribution as a table (vis/byage/dedup), explicit note that *Dispatch wrappers in sqs_keys_dispatch.go are NOT needed for M5-2 classic-only scope (M5-3 will pick them up). Addresses claude review 2 finding 1b which v3 missed.

bootjp:
v5: self-review pass 5 count updated to "One cross-check test (classic only; partitioned variant deferred to M5-3)" — addresses claude review 3/4 minor stale count. Claude review 4 already declared the proposal ready to accept; pushing this just keeps the doc internally consistent.

@claude review

PR #887, opened 2026-05-30
feat(txn-dedup): re-land M3 EXEC reuse on main + close M2 open + multi-mop

bootjp:
## Summary

PR #884 was merged into the stacked branch `docs/txn-idempotency-design` (at `cbbde3d7`) but never reached `main` — main has only PR #796's M1 + M2 + M3 RPUSH/LPUSH content. This PR re-lands the M3 EXEC reuse code on top of main and bundles three follow-ups that move the design doc's "still open" / "follow-up" items into "landed".

Base: `main`.

### 1. Re-land M3 EXEC reuse (PR #884's content)

- `adapter/redis.go`: `txnContext.prepareDispatch()` split out of `commit()`; `reusableExecTxn`; `dispatchExecReuse`; `runTransactionWithDedup` + `firstExecAttempt`; gate at the top of `runTransaction`.
- `adapter/redis_exec_dedup_test.go` (originally added in PR #884): 5 tests pinning all four reuse outcomes plus the gate-off legacy equivalence.
- Cherry-pick adaptation: `prepareDispatch()` uses `Clock().NextFenced() (uint64, error)` on current main; the PR #884 version targeted `Clock().Next() (uint64)`. Same downstream semantics, error wired through `preparedTxnDispatch`.

### 2. Close M2 open item — FSM other-txn exactness test

`TestOnePhaseDedup_OtherTxnVersionDoesNotMaskRetry` (`kv/fsm_onephase_dedup_test.go`) pins exactness at the FSM apply layer: a third-party version at `T_other=20` must not satisfy the FSM probe at `T1=30`, so the retry falls through and applies at the fresh `T2=40`. The store-layer pin already covers the primitive; this test covers the dispatch path that uses it.

### 3. M3 multi-mop EXEC dedup test

`TestExecDedup_MultiMopLandedPriorAttempt_ReturnsCachedResults` extends single-mop dedup to a 3-command MULTI/EXEC body (`SET a + SET b + DEL c`). Validates that cached results + OCC `readKeys` fence are mop-count-agnostic. Without dedup the DEL would re-execute to 0 on the second pass — the test rejects that.

### 4. Design doc updates

- §M2 "still open" → "LANDED" with the new FSM test reference
- §M3 "runTransaction (MULTI/EXEC) — Still open" → "LANDED via PR #884" with multi-mop test reference and acknowledgement of two intentional deviations from the M1/M2 template that claude[bot] flagged on #884 (`readKeys` assembly order, fresh per-attempt `reuseCtx`)
[...]

PR #888, opened 2026-05-30
feat(txn-dedup): standalone SET routes through option-2 dedup loop

bootjp:
## Summary

Stacked on PR-A (#887). Extends option-2 dedup to **standalone SET** (dispatched outside MULTI). When `r.onePhaseTxnDedup` is on, `r.set` wraps the single command as a 1-element `redcon.Command` queue and dispatches via `runTransactionWithDedup` — reusing the M3 EXEC machinery instead of building a per-handler `reusableSetTxn` + `dispatchSetReuse` shape. SET already has `applySet` on `txnContext`, so this is the "free" extension to any command whose txn-state-aware apply hook already exists.

The standalone fast path (`trySetFastPath`) is intentionally bypassed under the gate: dedup is opt-in, and a non-dedup'd fast path under a dedup-on cluster would split the idempotency contract.

Base: `feat/txn-dedup-docs-and-tests` (PR #887).

## Why standalone INCR / HSET are out of scope for THIS PR

INCR (in `adapter/redis_compat_commands.go`) and HSET both lack a `txnContext.applyIncr` / `applyHSet` implementation, so the "route through single-mop EXEC" pattern that worked here for SET cannot apply as-is — `runTransactionWithDedup` would reject the command at `txn.apply` ("unsupported in MULTI"). Bringing them in requires implementing `applyIncr` / `applyHSet` first (each ~30–50 LOC for the txn-state-aware read-compute-write shape), then the standalone handler routing is a one-liner via this same path. Tracked as separate follow-up PRs; until then, INCR and HSET keep today's buggy-under-churn behaviour — the design doc's stated default ("everything else keeps today's behaviour until its hook is added").

- `r.set` (handler): gate-on takes a new code path; gate-off preserved verbatim via `setLegacy`. No external callers exist (it is wired only into the redcon dispatch table).
- `setLegacy` (new): byte-identical extraction of the pre-PR `set()` body. No external callers.
- `writeRedisStandaloneResult` (new): translates a single-element `[]redisResult` from `runTransactionWithDedup` into the bare redcon reply shape (no `WriteArray` wrapper). Single caller in this PR; future SET-pattern callers will reuse it.

## Tests

- `adapter/redis_set_dedup_test.go` (new):
  - `TestStandaloneSetDedup_LandedPriorAttempt_ReturnsOK`: gate-on + ambiguous attempt-1 land + reuse → probe hits → cached "OK" returned, dispatches=2, probeNoOps=1.
  - `TestStandaloneSetDedup_DisabledKeepsLegacyPath`: gate-off path runs no probe.
- Existing `TestRedis_SET*` and dedup suites still pass.

## Validation

- `go test ./adapter/ -run 'Dedup|Txn|MULTI|EXEC|TestRedis_SET'` passes
- `gofmt`, `go vet`, `golangci-lint run` all clean (0 issues)

## Design doc

`docs/design/2026_05_21_proposed_txn_secondary_idempotency.md` §M3 — standalone SET marked LANDED, standalone INCR/HSET called out as follow-up with the precise reason they cannot land alongside SET in this PR.

### DIFF
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -1132,7 +1132,14 @@ func (r *RedisServer) set(conn redcon.Conn, cmd redcon.Command) {
 	// the standalone reply (resultString OK / resultNil for NX/XX miss /
 	// resultBulk for GET).
 	if r.onePhaseTxnDedup {
-		results, err := r.runTransaction([]redcon.Command{cmd})
+		// claude[bot] PR #888 review: call runTransactionWithDedup directly
+		// instead of going through runTransaction. runTransaction re-checks
+		// the same r.onePhaseTxnDedup gate and routes here anyway; the
+		// indirection made the call chain misleading ("dispatches via
+		// runTransactionWithDedup" was only true by indirection). Direct
+		// call makes the intent explicit and removes the double gate
+		// check.
+		results, err := r.runTransactionWithDedup([]redcon.Command{cmd})
 		if err != nil {
 			writeRedisError(conn, err)
 			return
@@ -1191,6 +1198,15 @@ func (r *RedisServer) setLegacy(conn redcon.Conn, cmd redcon.Command) {
 //
 // Empty or multi-element input is degenerate for standalone callers; we
 // default to nil so a misuse never leaks a malformed reply to the wire.
+//
+// Array-element constraint (claude[bot] PR #888 review): the resultArray
+// arm writes each element via WriteBulkString, which is correct for
+// flat arrays of strings (the shape applySet / future SET-pattern
+// callers produce). It does NOT recurse into nested arrays. A future
+// caller whose applyXxx emits resultArray with non-string elements
+// (e.g. HGETALL-like nested responses) must either pre-flatten its
+// result or extend this switch with a recursive arm; reusing it as-is
+// would silently mangle the wire reply.
 func writeRedisStandaloneResult(conn redcon.Conn, results []redisResult) {
 	if len(results) != 1 {
 		conn.WriteNull()
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -1132,13 +1132,13 @@ func (r *RedisServer) set(conn redcon.Conn, cmd redcon.Command) {
 	// the standalone reply (resultString OK / resultNil for NX/XX miss /
 	// resultBulk for GET).
 	if r.onePhaseTxnDedup {
-		// claude[bot] PR #888 review: call runTransactionWithDedup directly
-		// instead of going through runTransaction. runTransaction re-checks
-		// the same r.onePhaseTxnDedup gate and routes here anyway; the
-		// indirection made the call chain misleading ("dispatches via
-		// runTransactionWithDedup" was only true by indirection). Direct
-		// call makes the intent explicit and removes the double gate
-		// check.
+		// Call runTransactionWithDedup directly instead of going through
+		// runTransaction. runTransaction re-checks the same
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

PR-B: standalone SET routes through option-2 dedup loop.

Mechanism: when r.onePhaseTxnDedup is on, r.set wraps the command as a 1-element redcon.Command queue and dispatches via runTransactionWithDedup. setLegacy preserves the pre-PR body byte-identical for the gate-off path. writeRedisStandaloneResult translates the single-element results array to a bare redcon reply (no WriteArray wrapper).

Standalone INCR / HSET are documented as out of scope because they lack txnContext.applyIncr / applyHSet hooks (would each be a ~30-50 LOC feature add).

Please verify: (1) setLegacy is byte-identical to the pre-PR set() body; (2) the writeRedisStandaloneResult coverage matches every redisResult type emitted by applySet (resultString OK, resultNil for NX/XX miss, resultBulk for GET old-value); (3) the gate-off path is unchanged (TestStandaloneSetDedup_DisabledKeepsLegacyPath pins this).

bootjp:
@claude review

Round 1 fixes (commit 963a2d33):

- 🔶 Redundant double gate: set's gate-on branch now calls runTransactionWithDedup directly instead of runTransaction (which re-checked the same gate). Intent explicit, indirection removed.
- 🔶 Test gap: added TestStandaloneSetDedup_NXMissReturnsNil (resultNil routing) and TestStandaloneSetDedup_GETOptionReturnsOldBulk (resultBulk routing). Both fire the ambiguous-attempt-1-lands path so the result must come from the cached attempt-1 array.
- Minor: documented writeRedisStandaloneResult's shallow-array constraint in its doc comment per claude[bot]'s nested-array observation.

go test ./adapter/ -run StandaloneSetDedup passes (5 tests). gofmt/vet/golangci-lint clean.

bootjp:
@claude review

Round 2 fixes (commit fd91a6ad):

- (a) Stripped `// claude[bot] PR #888 review:` attribution prefix from the two inline comments per CLAUDE.md convention. Substantive "why" content kept (double-gate rationale; shallow-array constraint) — fix attribution lives in the commit message instead.
- (b) Added `wroteNull bool` witness field to `recordingConn` (in redis_retry_test.go). Hardened TestStandaloneSetDedup_NXMissReturnsNil to require.True(conn.wroteNull) so a wrong branch that wrote nothing at all would now be caught (not just the existing conn.bulk == nil zero-value match).

Caller audit:
- recordingConn is test-only; adding wroteNull bool default false is strictly additive. Grep'd all 8 recordingConn instantiations across redis_retry_test.go + redis_lua_linearizable_read_test.go — none read wroteNull. Only redis_set_dedup_test.go uses it.

go test ./adapter/ -run StandaloneSetDedup|ExecDedup passes; gofmt + go vet + golangci-lint clean.

bootjp:
@claude review

Round 3 cleanup (commit da6b0b41):

- Stripped the two remaining `claude[bot] PR #888` attribution refs from `redis_set_dedup_test.go` lines 56 and 78 — the same CLAUDE.md convention that was cleaned from redis.go in round 2 was left in the test file.
- Substantive rationale (NX semantics, zero-value aliasing problem) kept; attribution moves to the commit message.

grep 'claude\[bot\]\|PR #888' adapter/redis_set_dedup_test.go now returns empty. go test ./adapter/ -run StandaloneSetDedup passes. lint clean.

PR #891, opened 2026-05-30
test(sqs): structural fix — shared throttle-refill helper for drain-pattern tests

bootjp:
## Summary

Structural fix for the recurring SQS-throttle CI flake. PR #890 just merged a per-test fix for `SetQueueAttributesInvalidatesBucket`; this PR generalizes the rule across the whole file so the *next* drain-pattern test added to `adapter/sqs_throttle_integration_test.go` cannot regress to the same flake.

## Why a structural fix, not another per-test commit

The same CI-flake mechanism has been fixed three times in three different SQS-throttle tests over ~6 weeks:

| Commit | Test |
|---|---|
| [`54c6cd56`](https://github.com/bootjp/elastickv/commit/54c6cd56) | `NoOpSetQueueAttributesPreservesBucket` (PR #819) |
| [`204b7294`](https://github.com/bootjp/elastickv/commit/204b7294) | `SetQueueAttributesInvalidatesBucket` (PR #890) |
| **This PR** | structural prevention for the remaining drain-pattern tests |

Each per-test fix touched one test in isolation, leaving the same trap in **sibling tests** for the next slow CI runner to step on. Two sibling tests are still unprotected at the time of this PR:

- `TestSQSServer_Throttle_SendBucketRejectsAfterCapacity` — **HIGH risk**, drain 10 then assert 11th rejects, Refill=1
- `TestSQSServer_Throttle_RecvBucketRejectsAfterCapacity` — **HIGH risk**, receive-side mirror, same pattern

Plus `BatchChargesByEntryCount` is theoretically at risk on a very slow runner.

## Root cause (recurring across all three flake instances)

Tests that (a) set `ThrottleSend/RecvRefillPerSecond` AND (b) drain the bucket then assert a "throttled" response are racing their own wall clock. Under `-race` on slow CI runners each request elapses ~100-250 ms (full HTTP + Raft propose+apply). An N-write drain plus its assertion send takes ~N×150 ms ± noise. At Refill=1/sec the bucket accumulates ≥1 token over that window and the "throttled" assertion fires against a refilled bucket, returning 200 instead of 400 — a false positive against a real production-contract regression test.

## Structural fix

1. **Shared file-head constants + helpers** in `adapter/sqs_throttle_integration_test.go`:
   ```go
   const slowRefillRate = "0.01"  // 1 token per 100 seconds
   func newSendThrottleAttrs(capacity int) map[string]string
   func newRecvThrottleAttrs(capacity int) map[string]string
   ```
   Helpers lock the refill rate to `slowRefillRate` so a future drain-pattern test cannot pick a racing value by accident.

2. **File-head doc-comment block** explaining:
   - The recurring flake mechanism (with prior-commit refs)
   - The rule: drain-then-assert MUST use the helpers
   - The exception: config-round-trip / validator-rejection / post-raise tests where the rate is part of the test contract

3. **Migrated all drain-pattern call sites** to the helpers:
   - `SendBucketRejectsAfterCapacity` (HIGH risk — unprotected until now). Retry-After assertion updated from `"1"` to `"100"` (formula `ceil(needed/refillRate)` returns 100s for refillRate=0.01; verified by reading `computeRetryAfter` in `sqs_throttle.go:525`).
   - `RecvBucketRejectsAfterCapacity` (HIGH risk; same migration)
   - `BatchChargesByEntryCount` (LOW risk in practice, migrated for consistency)
   - `SetQueueAttributesInvalidatesBucket` (already fixed by PR #890, refactored to use the helper)
   - `DeleteQueueInvalidatesBucket` (no drain assertion, migrated for uniformity)
   - `NoOpSetQueueAttributesPreservesBucket` (already fixed by PR #819 with a local `slowRefill` const, refactored to use the shared helper)

**Not migrated (intentional):**
- `GetQueueAttributesRoundTrip` (Refill=50/5 — exact values are the test contract)
- `RejectsCapacityBelowBatchMin` (Refill=1 — value is part of the rejected input the validator must catch)
- `SetQueueAttributesInvalidatesBucket` post-raise phase (Refill=20 — raising both Capacity AND Refill is what the phase asserts fires)

Test-only file. Helpers are package-private and only referenced within this single file (verified by grep across the repo). No production code touched. No other test file uses these names.

Retry-After header assertion change in `SendBucketRejectsAfterCapacity` was the single observable behavior change: header depends on `refillRate` via the §3.4 formula. Old `"1"` matched Refill=1; new `"100"` matches Refill=0.01. The header contract is still exercised — only the asserted literal changed.

## Validation

- `go test ./adapter/ -run TestSQSServer_Throttle -race -count=3 -timeout 180s` → ok 5.1s, all three iterations green
- `gofmt` + `go vet` + `golangci-lint run` → 0 issues

## Out-of-scope follow-ups (other recurring flake categories)

Broader CI-flake history (47 fix-CI-flake commits) has other recurring patterns that deserve similar structural treatment — tracked as future work, out of scope for this PR:

- **Leader churn + Raft async apply**: many tests retry on "not leader" or wait for follower catch-up. A shared `eventuallyLeaderOK(t, ...)` helper would centralize the retry + budget knobs.
- **Redis TTL window racing**: multiple "widen TTL" commits suggest a per-test deadline-multiplier helper.
- **VM pool / cache hit-rate assertions**: dropped as flaky in `7783c279` — the structural lesson is to avoid asserting on cache hit rates.

This PR focuses on the most-repeated single flake (3 fixes for the same mechanism in 6 weeks). The remaining categories each warrant their own follow-up PR after we see how this pattern works in practice.

## Summary by CodeRabbit

* **Tests**
  * Enhanced SQS throttle test infrastructure to improve test reliability in continuous integration environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/891?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/adapter/sqs_throttle_integration_test.go
+++ b/adapter/sqs_throttle_integration_test.go
@@ -36,27 +36,35 @@ import (
 // for `mustSetQueueAttributes(... "ThrottleSendRefillPerSecond": "1" ...)`
 // at least has to think about why they didn't use `newSendThrottleAttrs`.
 
-// slowRefillRate is 1 token per 100 seconds — slow enough that no
-// realistic test-window wall-clock can accumulate to a whole token.
-// The throttle-config validator accepts fractional `float64` (see
-// `sqs_catalog.go` `SendRefillPerSecond float64`) and `0.01 != 0`
-// keeps `IsEmpty()` returning false, so throttling stays enabled.
-const slowRefillRate = "0.01"
+const (
+	// slowRefillRate is 1 token per 100 seconds — slow enough that no
+	// realistic test-window wall-clock can accumulate to a whole
+	// token. The throttle-config validator accepts fractional
+	// `float64` (see `sqs_catalog.go` `SendRefillPerSecond float64`)
+	// and `0.01 != 0` keeps `IsEmpty()` returning false, so
+	// throttling stays enabled.
+	slowRefillRate = "0.01"
+
+	// drainBucketCapacity is the canonical capacity for the
+	// drain-pattern tests in this file. 10 is the §3.2 validator's
+	// batch-floor minimum (SendCapacity ≥ 10 required so the §3.3
+	// charging table's 10-entry batch can succeed), so it doubles
+	// as the smallest value that can drain via either single or
+	// batch sends. A future test needing a different capacity
+	// should not use these helpers — the helpers exist to encode
+	// the *one* canonical shape for drain-then-assert.
+	drainBucketCapacity = "10"
+)
 
 // newSendThrottleAttrs returns an attribute map for a drain-then-assert
-// send-side throttle test. Capacity is caller-supplied; refill is
-// locked to `slowRefillRate` to immunize against the wall-clock race.
-//
-// All current callers pass `10`; the parameter is forward-looking so
-// new tests (and the §3.2 validator's batch-floor edge: Capacity ≥ 10)
-// can pick a value without going behind the helper. The unparam lint
-// is suppressed because the future-call diversity is the whole point
-// of taking a parameter.
-//
-//nolint:unparam // capacity parameter is forward-looking; see doc.
-func newSendThrottleAttrs(capacity int) map[string]string {
+// send-side throttle test. Both fields are locked: capacity to
+// drainBucketCapacity (the validator's batch-floor minimum), refill
+// to slowRefillRate (immune to the wall-clock race). Tests that
+// genuinely need different values are out of scope for this helper
+// and should construct the map literally.
+func newSendThrottleAttrs() map[string]string {
 	return map[string]string{
-		"ThrottleSendCapacity":        strconv.Itoa(capacity),
+		"ThrottleSendCapacity":        drainBucketCapacity,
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

Round 1 fixes (commit e5f31767):

- **gemini HIGH** (Retry-After exact-value assertion is itself flaky): Replaced exact `!= "100"` check with a range check [1, 100]. The formula `ceil((1 - currentTokens) / refillRate)` depends on drain-window elapsed time (refill accumulates during the loop), so asserting "100" exactly re-introduces a wall-clock flake at the assertion layer. The range pins the §3.4 contract (positive integer bounded by 1/refillRate) without racing.

- **coderabbit nitpick** (`//nolint:unparam`): Removed by making helpers zero-arg. Capacity is now a documented file-level const `drainBucketCapacity = "10"` (the §3.2 validator batch-floor minimum). All 9 call sites updated from `foo(10)` → `foo()`.

Caller audit: test-only file, helpers package-private, no external references. `go test -race -count=3` passes (5.3s); golangci-lint 0 issues.

PR #892, opened 2026-05-30
backup: Phase 0b M5-2 - SQS side-record derivation (vis/byage/dedup)

bootjp:
## Summary

Phase 0b M5-2 implementation per the merged design doc [`docs/design/2026_05_30_proposed_sqs_side_record_derivation.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_05_30_proposed_sqs_side_record_derivation.md) (#885).

Resolves the SQS side-record decision gate the M5-1 encoder (#846) explicitly deferred. Without this PR, restoring a backup silently breaks: FIFO dedup gating (the next send after restore replays an already-acked message), the by-age reaper (retention is silently ignored, queues grow unboundedly), and visibility (ReceiveMessage finds nothing).

## What lands

The encoder now emits three derived side-record families inline during M5-1's per-message walk:

| Family | Always? | Value | Why required |
|---|---|---|---|
| `!sqs\|msg\|vis\|...` | yes | `[]byte(messageID)` | ReceiveMessage scans this prefix |
| `!sqs\|msg\|byage\|...` | yes | `[]byte(messageID)` | reaper honors `MessageRetentionPeriod` |
| `!sqs\|msg\|dedup\|...` | FIFO + non-empty `MessageDedupID` | `sqsFifoDedupRecord` JSON (4 fields) | gates FIFO sends within 5-minute window |

`!sqs|msg|group|` rows are **never** emitted: the live `loadFifoGroupLock` treats key presence alone as "lock held" with no graceful empty-owner decode, so any value (even zeroed) would permanently block every group post-restore (gemini critical #885). Test `TestSQSEncodeSideRecordsNoGroupRows` pins this.

Scope is classic queues only (`partition_count = 1`). M5-1's existing `ErrSQSEncodeUnsupportedPartitioned` gate at `encodeQueue` means the side-record walk is never reached for partitioned queues. Partitioned-FIFO support requires coordinated decoder + encoder lift (deferred to M5-3 per the design doc's "Deferred (partitioned-FIFO)" section).

## Test plan

`go test -race -count=1 ./internal/backup/` — all green, including:

- `TestSQSEncodeSideRecordsCrossCheckClassic` — byte-identical key + value cross-check against the live adapter constructors (`sqsMsgVisKey`, `sqsMsgByAgeKey`, `sqsMsgDedupKey`, `sqsFifoDedupRecord`) for a 2-message FIFO fixture. The gold-standard pattern the parent encoder design (§"Encoder cross-check") mandates.
- `TestSQSEncodeSideRecordsStandardQueueOmitsFIFOFamilies` — non-FIFO queue: vis + byage emitted, dedup NOT.
- `TestSQSEncodeSideRecordsEmptyDedupOmitsDedupRow` — FIFO + empty `message_deduplication_id`: dedup row NOT emitted.
- `TestSQSEncodeSideRecordsNoGroupRows` — FIFO queue across 3 distinct `message_group_id` values: 0 `!sqs|msg|group|` rows emitted (regression pin for gemini critical #885).

The two restore round-trip tests listed in the design doc test plan (`TestSQSEncodeFifoRestoreRoundTrip`, `TestSQSEncodeReaperFindsRestoredMessage`) are deferred to the M6 CLI / Jepsen integration layer rather than this package — they require single-node cluster boot that is heavier infra than this package's existing pattern. The cross-check + three negative tests above pin the contract M5-2 is responsible for at the encoder layer; cluster-level restore validation belongs at the integration level.

`make lint` clean (verified via golangci-lint --fix on the touched files).

## Self-review (5 passes)

1. **Data loss** — Full reconstruction eliminates the silent-redeliver / invisible-message / retention-leak classes M5-1-only restores would produce. `addSideRecords` surfaces every `b.Add` error and is wired to the existing per-message walk's error path; no swallowed errors.
2. **Concurrency / distributed failures** — Pure offline derivation. No Raft, no HLC, no locks; restored cluster's first sends/receives go through the normal OCC path against the restored state.
3. **Performance** — One `b.Add` per emitted side row, no extra disk reads (records already loaded by M5-1). Snapshot grows ~2-3× message count, all small fixed-width keys. `addMessage`'s existing loop body inflates by ~5 lines, no extra iterations.
4. **Data consistency** — Group-row prohibition pinned by `TestSQSEncodeSideRecordsNoGroupRows`. Dedup gating on `(FIFO && non-empty MessageDedupID)` pinned by two tests. `visibleAt = available_at_millis` matches the parent design's "SQS: vis zeroed by default" contract. Cross-check test validates byte-identical output against the live writer for a shared fixture.
5. **Test coverage** — One cross-check (classic; partitioned variant deferred to M5-3) plus three negative tests cover the three conditional-emission rules the design enumerates. Round-trip tests deferred to M6/Jepsen as discussed above.

## Risk

Low. The encoder is offline; restoration is non-destructive (new keyspace on a fresh cluster). The change is additive to M5-1's existing per-message walk and gated on conditions verified by the negative tests; existing M5-1 round-trip tests continue to pass (verified locally). The single semantic-changing behavior is "M5-1-only restores silently break FIFO/reaper/receive" → "restore now works"; no caller of M5-1 depends on the prior broken state.

## Summary by CodeRabbit

* **New Features**
  * SQS message backups now include preservation of visibility metadata, age-based attributes, and FIFO deduplication records for complete message restoration.

* **Tests**
  * Added comprehensive test coverage validating SQS side record encoding across standard and FIFO queue scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/bootjp/elastickv/pull/892?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/backup/encode_sqs.go
+++ b/internal/backup/encode_sqs.go
@@ -211,7 +211,7 @@ func (e *SQSRecordEncoder) encodeQueueMessages(b *snapshotBuilder, root *os.Root
 		if err != nil {
 			return err
 		}
-		if err := e.addSideRecords(b, meta.Name, meta, records[i]); err != nil {
+		if err := e.addSideRecords(b, meta.Name, &meta, &records[i]); err != nil {
 			return err
 		}
 		if seq > maxSeq {
--- a/internal/backup/encode_sqs.go
+++ b/internal/backup/encode_sqs.go
@@ -211,6 +211,9 @@ func (e *SQSRecordEncoder) encodeQueueMessages(b *snapshotBuilder, root *os.Root
 		if err != nil {
 			return err
 		}
+		if err := e.addSideRecords(b, meta.Name, meta, records[i]); err != nil {
+			return err
+		}
 		if seq > maxSeq {
 			maxSeq = seq
 		}
--- a/internal/backup/encode_sqs_side.go
+++ b/internal/backup/encode_sqs_side.go
@@ -12,6 +12,12 @@ import (
 // will be expired the next time the live FIFO send path inspects them.
 const sqsFifoDedupWindowMillis int64 = 5 * 60 * 1000
 
+// sqsSideKeyAllocBytes mirrors the live adapter's sqsKeyCapLarge tuning
+// (adapter/sqs_messages.go:68): a 64-byte tail after the prefix is large
+// enough to hold the BE-u64 generation + visibleAt + base64url(messageID)
+// for typical queue / message IDs without forcing a re-allocation.
+const sqsSideKeyAllocBytes = 64
+
 // sqsFifoDedupRecord mirrors the live struct at adapter/sqs_fifo.go:25.
 // Duplicated here (rather than imported) so the encoder package can run
 // without a circular dependency on adapter, matching the pattern M3b-3
@@ -28,7 +34,8 @@ type sqsFifoDedupRecord struct {
 // base64url(messageID). Negative visibleAt clamps to zero, matching the
 // live uint64MaxZero helper.
 func sqsMsgVisKeyBytes(queueName string, gen uint64, visibleAtMillis int64, messageID string) []byte {
-	out := []byte(SQSMsgVisPrefix)
+	out := make([]byte, 0, len(SQSMsgVisPrefix)+sqsSideKeyAllocBytes)
+	out = append(out, SQSMsgVisPrefix...)
 	out = append(out, encodeSQSSegment(queueName)...)
 	out = binary.BigEndian.AppendUint64(out, gen)
 	out = binary.BigEndian.AppendUint64(out, sqsClampNonNegativeMillis(visibleAtMillis))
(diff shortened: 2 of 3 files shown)


bootjp:
@claude review

bootjp:
Addressed all 6 gemini medium-priority perf findings in de70bfca:
1. Pre-allocate the 3 key-constructor backing slices with len(prefix)+64 (matches live adapter sqsKeyCapLarge)
2. addSideRecords takes *sqsQueueMetaPublic + *sqsMessageRecord (single caller updated)
3. msgIDBytes converted once and reused for vis + byage values

Caller audit: addSideRecords has one caller (encodeQueueMessages); strictly read-only on both pointer args, no aliasing risk, error contract unchanged.

@claude review

bootjp:
v3 (b133e5cb): addresses all 4 claude review observations.

**Obs 2 (medium — REAL DATA-LOSS BUG):** CBD-FIFO queues now correctly emit dedup rows on restore. The live adapter's sqs_messages.go:680 only stores the user-supplied MessageDeduplicationId on the message record, so a CBD send round-trips through the dump as message_deduplication_id="" and the prior empty-only gate silently dropped dedup protection. New resolveDedupID(rec, meta) mirrors adapter/sqs_fifo.go:79 resolveFifoDedupID exactly (sha256 hex over rec.Body for CBD). New TestSQSEncodeSideRecordsCBDDerivesDedupID pins it.

**Obs 1 (medium):** Test comment for TestSQSEncodeSideRecordsCrossCheckClassic now accurately describes in-package consistency (the adapter constructors are unexported and not callable from this package; cross-package parity is by manual review).

**Obs 3 (nit):** Removed the unreachable empty-MessageID guard in addSideRecords (addMessage already validates first; CLAUDE.md says don't validate scenarios that can't happen).

**Obs 4 (nit):** Replaced misleading "(visible now)" comment with an accurate description acknowledging that AvailableAtMillis can be in the future for delayed sends.

Caller audit: addSideRecords has one in-package caller (encodeQueueMessages); semantic change is strictly additive (emits dedup rows in the CBD case that the prior empty-only gate skipped); error contract unchanged.

@claude review

bootjp:
v4 (3f9be739): added d2 assertDedupValue per coderabbit nit, pins per-message seq+timestamp threading (not just key presence).

@claude review

PR #886, opened 2026-05-30
feat(encryption): Stage 8a implementation — snapshot header v2 + ReadSnapshotHeader API + no-downgrade latch

bootjp:
## Summary

Implements Stage 8a — the snapshot header v2 carrier for `raft_envelope_cutover_index`. Lands on top of the merged design doc (`docs/design/2026_05_29_proposed_8a_snapshot_header_v2.md`, #877). No on-disk migration; v1 snapshots restore byte-for-byte under the new build, and only Phase-2 nodes ever emit v2.

### New API surface

- `ReadSnapshotHeader(r *bufio.Reader) (ceiling, cutover uint64, err error)` — the 6-branch §3.2 reader. Caller-owned `bufio.Reader` is passed unchanged into the inner store on every branch (the bufio-buffering-loses-bytes failure mode flagged in PR #877 review).
- `CutoverSource` interface + `WithCutoverSource(...)` FSM option — writer-side view of the sidecar's cutover index.
- `noDowngradeLatch` (process-load) — §3.3's "once non-zero observed → always v2" invariant. Preserves the last non-zero cutover across hypothetical sidecar resets and slogs an Error once on the non-zero→0 transition.
- `kvFSM.RestoredCutover()` — test seam for the §3.4 snapshot-to-applier handoff. Stage 6E will consume this via the apply hook.

### Read-path branches (all covered by tests)

| Input | Result |
|---|---|
| v1 magic `EKVTHLC1` | `(ceiling, 0, nil)` |
| v2 magic `EKVTHLC2` + valid len | `(ceiling, cutover, nil)`, trailing payload bytes skipped |
| `EKVTHLC*` w/ unknown version byte | `(0, 0, ErrSnapshotHeaderUnknownMagic)` — fail-closed |
| v2 magic + `len < 16` or `len > 1024` | `(0, 0, ErrSnapshotHeaderInvalidLength)` — fail-closed (DoS bound) |
| No `EKVTHLC` prefix | `(0, 0, nil)` — headerless legacy, bytes preserved in `bufio.Reader` |
| Stream `< 8` bytes | `(0, 0, nil)` — short-stream fallback, bytes preserved |

### Verification

- 11 new tests in `kv/snapshot_test.go` per design §5.
- 4 pre-8a regression tests (TestSnapshot, TestFSMSnapshotPreservesCeiling, TestFSMSnapshotRestoreOldFormat, TestFSMSnapshotRestoreSmallLegacy) stay green.
- `go test -race ./kv/...` clean.
[...]

PR #893, opened 2026-05-30
docs(encryption): Stage 6E design proposal — enable-raft-envelope admin RPC + Phase-2 cutover

bootjp:
## Summary

Stage 6E design proposal — lands `enable-raft-envelope` admin RPC + the §7.1 Phase-2 raft cutover end-to-end. Unblocked by Stage 8a (#886), which shipped the v2 snapshot reader that 6E depends on for mid-Phase-2 follower catch-up.

### Why now

Every prerequisite has shipped:
- 6D — Phase-1 storage envelope cutover
- 7 — Writer registry + deterministic nonce
- 8a — Snapshot header v2 cutover carriage

6E is the missing piece between sidecar-recorded cutover and on-the-wire wrap/unwrap.

### Sliced into 3 implementation milestones

| Milestone | Scope | Why this slice |
|---|---|---|
| **6E-1** | Admin RPC + sidecar plumbing (no behavior change) | Records cutover index but no wrap/unwrap yet — safe to ship alone |
| **6E-2** | Engine unwrap + coord wrap + quiescence barrier | The three pieces are interlocked at the cluster-halt boundary; cannot split |
| **6E-3** | 6C-4 fail-closed guards (divergence, not-bootstrapped, epoch-out-of-range) | Dead code without an active Phase 2; bundles after 6E-2 |

### Load-bearing details documented

- **Why the barrier's source-tag exemption is required** — a global `Propose` mutex would deadlock on the cutover's own proposal.
- **Why dispatch is strict-greater-than `index > cutover`** — the cutover entry at `index == cutover` MUST NOT be unwrapped (chicken/egg with the sidecar update).
- **Why `decodeProposalEnvelope` MUST run before unwrap** — wrapping `entry.Data` itself would clobber the proposal-ID handoff to `resolveProposal`; every coordinator write would time out forever.
- **Why 6E-2's three pieces cannot ship separately** — either ordering produces cluster-wide halt-apply at cutover.

### Verification action items

8 + 8 + 3 tests across the three milestones, with explicit 5-lens self-review focus per milestone (data loss for 6E-2's barrier window, data consistency for 6E-3's startup divergence refusal).

## Test plan

- [ ] Design doc reviewable on its own (no implementation in this PR).
- [ ] @claude review for the milestone breakdown, the quiescence-barrier sequencing, and the strict-`>` index dispatch rule.
- [ ] Implementation lands in three follow-on PRs (6E-1 → 6E-2 → 6E-3) per CLAUDE.md design-doc-first workflow.

### DIFF
--- a/docs/design/2026_05_31_proposed_6e_enable_raft_envelope.md
+++ b/docs/design/2026_05_31_proposed_6e_enable_raft_envelope.md
@@ -1,0 +1,397 @@
+# Stage 6E — `enable-raft-envelope` admin RPC + Phase-2 raft cutover
+
+| Field | Value |
+|---|---|
+| Status | proposed |
+| Date | 2026-05-31 |
+| Parent design | [`2026_04_29_partial_data_at_rest_encryption.md`](2026_04_29_partial_data_at_rest_encryption.md) (§4.2 raft envelope, §6.3 engine apply-hook, §6.6 admin RPC, §7.1 Phase-2 cutover) |
+| Builds on | Stage 6A–6D (capability gate, storage envelope cutover, sidecar field plumbing), Stage 7 (writer registry), Stage 8a (snapshot header v2 cutover carriage) |
+| Sibling slice | Stage 8b — WAL coverage (not blocked on 6E) |
+
+## 0. Why this slice exists
+
+Today every leader proposes Raft entries as plaintext: the
+`internal/raftengine/etcd/engine.go::applyNormalEntry` path
+decodes the proposal envelope and hands the FSM payload straight
+to `kv/fsm.go::Apply`. The §4.2 raft envelope is fully specified
+(DEK shape, KEK wrap, in-flight rotation), but the on-the-wire
+activation — flipping `wrap on Propose` on the leader and the
+matching `unwrap on Apply` on every replica — has not landed.
+Without 6E, the cluster cannot reach Phase 2 even though every
+prerequisite (6D Phase-1 storage envelope, 7 writer registry, 8a
+snapshot cutover carriage) has shipped.
+
+This slice lands the actual Phase-2 raft cutover end-to-end:
+admin RPC, sidecar cutover-index recording, engine unwrap-on-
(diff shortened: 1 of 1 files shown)


bootjp:
@claude review

Specifically please verify:
- The milestone breakdown is correct (6E-1 is safe to ship alone with no behavior change; 6E-2's three pieces truly cannot ship separately; 6E-3 guards bundle correctly).
- The §7.1 6-step quiescence-barrier sequence — particularly the source-tag exemption rationale and the strict-greater-than index dispatch rule (`index > cutover`, not `>=`).
- The `decodeProposalEnvelope` ordering invariant (must run BEFORE unwrap so the proposal-ID handoff to resolveProposal stays intact).
- The sequencing claim that 8a (#886) being already shipped is sufficient to unblock 6E (snapshot v2 reader carries cutover across mid-Phase-2 follower joins).

bootjp:
Round-1 fixes in dd0beb22:

- **gemini HIGH** §2.1 (line 57) — startup-init rule for `wrapOnPropose`. The original spec left a gap where a post-cutover restart would default the flag to `false`, a new-leader plaintext propose would land at `index > cutover`, the engine apply-hook would attempt Unwrap, fail GCM, and halt apply cluster-wide. Now requires initialization to `true` iff `sidecar.RaftEnvelopeCutoverIndex != 0` at coordinator construction time. Two write paths (startup hydration, barrier step 5), zero reset paths within a process load. Added test `TestCoordinatorWrap_StartupInitFromSidecar` in §6.2.
- **gemini MEDIUM** §2.4 (line 143) — verified actual signature at `internal/raftengine/etcd/engine.go:2226` is `(entry raftpb.Entry) (any, error)`; `setApplied` and `resolveProposal` live in `applyNormalCommitted` (line 2173, 2186-2187). Updated the code snippet + commentary; 6E-2 adds the unwrap shim inside `applyNormalEntry` without changing the caller contract.
- **gemini MEDIUM** §5 header — typo: was 'is unwrapped', now 'is NOT unwrapped' (matches the body).

@claude review

PR #895, opened 2026-05-31
kv(composed1): M3 — verifyComposed1 apply-time gate + retry sentinels

bootjp:
## Summary

Third milestone (**M3**) of the Composed-1 cross-group commit-time guard per `docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md` §M3.

Wires the safety property at apply time using the M2 versioned-snapshot ring and the M1 ObservedRouteVersion field.

**Stacked on PR #894** (M2 plumbing).

## What's enforced

- **(a) Observed-version owner** — the spec-level Composed-1 check from `tla/composed/Composed.tla`. Every write key of a Composed-1-pinned txn must be owned by THIS Raft group at the txn's observed catalog version. Refusal returns `ErrComposed1Violation`.
- **(b) Current-version cross-version-read fence** — the §4.4 / §3 codex P1 trace. Even when (a) passes, a route shift between BeginTxn and Commit can leave the write landing on the OLD owner while readers at the new version route to the NEW owner and miss the write. The current-version fence refuses that case. Refusal returns `ErrComposed1Violation` with a different diagnostic prefix (`current-version` vs `observed-version`) so the M4 retry path can distinguish.
- **(c) Retention-miss fail-closed** — when the txn's observed version has been evicted from the M2 ring (long-running txn or high catalog churn), return `ErrComposed1VersionGCd`. Per design doc §4.3, the not-found semantics is a hard retryable error, NOT a soft pass, because soft-pass would let the gate be bypassed exactly in the cases where the cross-version-read hazard is most likely.

## Short-circuits

Three legacy / not-applicable cases that the gate explicitly bypasses:

- **FSM constructed without `WithRouteHistory`** (legacy / test seam): `routes == nil`, gate returns nil.
- **Request carries `ObservedRouteVersion == 0`** (unpinned — pre-M1 caller, or 2PC ABORT request that doesn't carry the version): gate returns nil. This is what keeps M3 from regressing every caller that has not migrated to set `OperationGroup.ObservedRouteVersion`.
- **`Engine.Current` returns `(zero, false)`** — engine has no history (bare-struct test seam): the (b) fence falls through.

## Changes

- **`distribution/engine.go`** — adds `Engine.Current()` returning the snapshot at the current `catalogVersion`. Adds `SetHistoryDepthForTest` as the cross-package test seam that lets kv-side tests trigger eviction without bypassing the package boundary (claude review on PR #894 — fragile-but-test-local lock contract documented inline).
- **`kv/fsm.go`** — defines `ErrComposed1Violation` and `ErrComposed1VersionGCd` sentinels with the wrapped-error diagnostic shape M4 retry will use. Adds `verifyComposed1` and `verifyOwnerFromSnapshot`. Wires `verifyComposed1` into `handleTxnRequest` at the top so PREPARE / COMMIT / ABORT / NONE all pass through the gate (ABORT carries `ObservedRouteVersion=0` and falls through naturally).
- **`kv/route_history.go`** — `RouteHistory` interface gains `Current()`. Adapter forwards to `distribution.Engine.Current()`.

## Tests (`kv/fsm_composed1_test.go`)

| Test | Maps to design doc M3 "Done when" |
|---|---|
| `TestVerifyComposed1_StaleObservedVersionWithMovedKeyFails` | (i) observed-version snapshot resolves the key to a different group than this FSM |
| `TestVerifyComposed1_ObservedVersionOlderThanRingFails` | (ii) observed version is outside the ring → `ErrComposed1VersionGCd` |
| `TestVerifyComposed1_ObservedPassesButCurrentDiffersFails` | (iii) / §3 codex P1 trace: observed-version check passes but current-version differs → `ErrComposed1Violation` (`current-version` diagnostic) |
| `TestVerifyComposed1_ObservedVersionZeroSkipsGate` | Legacy caller bypasses the gate |
| `TestVerifyComposed1_NilRouteHistorySkipsGate` | Unwired FSM bypasses the gate |

## Verification

- `go build ./...` — clean
- `go vet ./...` — clean
- `go test -race -count=1 ./kv ./distribution` — 10.2 s + 1.0 s, pass
- `make lint` — 0 issues

## Self-review (5 lenses)

1. **Data loss** — the gate refuses commits but cannot lose them. Every rejection emits a sentinel the M4 coordinator path will convert into a successful commit on the correct owner.
2. **Concurrency** — the gate runs under the FSM's apply loop (Raft applies are serial), so no new lock ordering. RouteHistory reads take the engine's read lock; `SnapshotAt` and `Current` each acquire and release in a single call.
3. **Performance** — two map lookups + a per-mutation linear scan of the snapshot's sorted routes per Composed-1-pinned commit. `OwnerOf` short-circuits via the M2 round-1 break optimisation, so the per-mutation cost is bounded by the first non-covering gap rather than the full routes slice. Legacy callers (`ObservedRouteVersion=0`) pay one branch per commit and exit.
4. **Data consistency** — closes the Composed-1 (observed) and Composed-1a (current) gaps end-to-end. Spec correspondence: `tla/composed/Composed.tla`'s `Commit` precondition is now enforced at apply time; the §4.4 fence matches `Composed1a_CommitToCurrentOwner` from PR #878.
5. **Test coverage** — 5 new unit tests cover the three design doc criteria + the two legacy short-circuit paths. Existing `handleTxnRequest` tests pass unchanged (their requests carry `ObservedRouteVersion=0` from M1's behaviour-neutral default).

## Test plan

- [x] `go test -race ./kv ./distribution`
- [x] `make lint`
- [ ] **#894 (M2) must merge first** — this PR's branch is stacked on it.
- [ ] (Follow-up PR) **M4** — coordinator retry path. When either Composed-1 sentinel returns, the coordinator re-reads the route cache, re-routes the txn against the new owning group, and re-issues it once.

## Resolves

The M3 row in the Composed-1 design doc.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents cross-group transaction commits when route ownership shifts between observed and current versions.
  * Returns a clear error when an observed version is no longer retained in route history.
  * Ensures transactions are skipped (not rejected) when the shard/group identifier is unset, avoiding false rejections.

* **Tests**
  * Added comprehensive tests covering validation across routing/version scenarios, including history GC and shard-group unset behavior.

### DIFF
--- a/distribution/engine.go
+++ b/distribution/engine.go
@@ -214,16 +214,13 @@ func (e *Engine) SnapshotAt(v uint64) (RouteHistorySnapshot, bool) {
 // avoid interleaving surprises around the eviction watermark, but
 // the write itself is lock-protected (e.mu.Lock below) so it is
 // safe to call from any goroutine that does not also expect a
-// consistent SnapshotAt view across the depth change.  Exists so
-// tests in the kv package can drive eviction-trigger scenarios
-// without adding a constructor option just for tests (claude
-// review on PR #894).
+// consistent SnapshotAt view across the depth change.
 //
-// Production code must use DefaultRouteHistoryDepth (32) or a
-// future operator-exposed config knob; this seam is build-time
-// equivalent to direct field access and exists ONLY so tests in
-// the kv package can drive eviction-trigger scenarios without
-// adding a constructor option just for tests.
+// Exists so tests in the kv package can drive eviction-trigger
+// scenarios without adding a constructor option just for tests
+// (claude review on PR #894).  Production code must use
+// DefaultRouteHistoryDepth (32) or a future operator-exposed
+// config knob.
 //
 // Fails fast on depth <= 0 (coderabbit minor on PR #895):
 // recordHistorySnapshotLocked's eviction path indexes
--- a/distribution/engine.go
+++ b/distribution/engine.go
@@ -179,6 +179,21 @@ func (s RouteHistorySnapshot) OwnerOf(key []byte) (uint64, bool) {
 	return 0, false
 }
 
+// Current returns the route catalog snapshot at the engine's current
+// catalogVersion.  Returns (zero, false) when the history ring has
+// not been initialised (bare-struct Engine).  Used by the M3
+// Composed-1 cross-version-read fence (design doc §4.4) — the gate
+// compares the txn's observed-version owner against the current
+// owner so a route shift between BeginTxn and Commit is caught
+// before it can produce a G1c anomaly across a cross-group
+// MoveRange / SplitRange.
+func (e *Engine) Current() (RouteHistorySnapshot, bool) {
+	e.mu.RLock()
+	defer e.mu.RUnlock()
+	snap, ok := e.history[e.catalogVersion]
+	return snap, ok
+}
+
 // SnapshotAt returns the route catalog snapshot recorded at version v.
 // Returns (zero, false) when v is not in the ring — either because v
 // is in the future (> catalogVersion), or because the FIFO ring has
@@ -191,6 +206,26 @@ func (e *Engine) SnapshotAt(v uint64) (RouteHistorySnapshot, bool) {
 	defer e.mu.RUnlock()
 	snap, ok := e.history[v]
 	return snap, ok
+}
+
+// SetHistoryDepthForTest overrides the FIFO ring depth from outside
+// the package.  Test-only — callers MUST set the depth before the
+// Engine is shared with any concurrent reader (no internal
+// synchronisation here for the same reason TestEngineSnapshotAt_FIFOEviction
+// does the direct field write in-package; this seam exposes the
+// equivalent capability to external test packages that need a
+// small depth to exercise eviction without overwhelming TLC-style
+// bounded scenarios — claude review on PR #894).
+//
+// Production code must use DefaultRouteHistoryDepth (32) or a
+// future operator-exposed config knob; this seam is build-time
+// equivalent to direct field access and exists ONLY so tests in
+// the kv package can drive eviction-trigger scenarios without
+// adding a constructor option just for tests.
+func (e *Engine) SetHistoryDepthForTest(depth int) {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	e.historyDepth = depth
 }
 
 // HistoryDepth returns the configured ring depth for diagnostics.
--- a/distribution/engine.go
+++ b/distribution/engine.go
@@ -222,10 +222,34 @@ func (e *Engine) SnapshotAt(v uint64) (RouteHistorySnapshot, bool) {
 // equivalent to direct field access and exists ONLY so tests in
 // the kv package can drive eviction-trigger scenarios without
 // adding a constructor option just for tests.
+//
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

bootjp:
All 4 findings addressed:

**coderabbit MAJOR** (kv/fsm.go:589 — shardGroupID==0 bypass) — verifyComposed1 now honors both `routes==nil` and `shardGroupID==0` per the documented unset-sentinel contract. Caller audit (per the loop directive's semantic-change check): only `handleTxnRequest` calls verifyComposed1 and treats nil as success — bypass-as-nil preserves caller semantics, no further wiring change. Regression test `TestVerifyComposed1_ShardGroupIDZeroSkipsGate` added BEFORE the fix per project convention, verified failing on parent / passing on fix.

**coderabbit minor** (distribution/engine.go:229 — reject depth<=0) — SetHistoryDepthForTest now panics at the seam with a clear message.

**gemini medium** (distribution/engine.go:229 — shrink panic) — SetHistoryDepthForTest now evicts excess entries and compacts the slice when shrinking depth below current ring size.

**golangci/reviewdog** (gci formatting on kv/fsm.go and kv/fsm_composed1_test.go) — auto-fixed by `make lint --fix`.

`go test -race ./kv ./distribution` pass / `make lint` 0 issues.

@claude review

bootjp:
Round-3: all remaining items addressed.

**gemini HIGH** (kv/fsm.go — ABORT bypass): Added `if r.GetPhase() == pb.Phase_ABORT { return nil }` as the FIRST guard in verifyComposed1. Production callers leave ObservedRouteVersion=0 on ABORT so today the existing observedVer==0 branch already protects them, but that's enforcement-by-convention — the new Phase_ABORT guard is enforcement-in-code. A future ABORT construction site that copies a COMMIT struct by mistake still bypasses. Caller audit: only handleTxnRequest calls verifyComposed1 and treats nil as success — semantics preserved.

**claude follow-up #1** (SetHistoryDepthForTest doc comment): Rewrote to remove the contradictory "no internal synchronisation" wording; comment now correctly states the write is lock-protected but callers should still set depth before sharing the Engine with concurrent readers to avoid interleaving surprises.

**claude follow-up #2** (happy-path test): Added `TestVerifyComposed1_ValidOwnershipPassesGate` — this FSM serves the group that owns the key at both the observed AND current version, gate returns nil. Locks down the pass direction the failure-only suite was missing.

**ABORT regression test**: `TestVerifyComposed1_AbortPhaseSkipsGate` deliberately sets ObservedRouteVersion=1 on the ABORT to defeat the existing observedVer==0 short-circuit and isolate the new Phase_ABORT guard.

**Nit**: removed the now-redundant `//nolint:unparam` on newComposed1FSM (new tests pass distinct shardGroupID values).

`go test -race ./kv ./distribution` pass / `make lint` 0 issues.

@claude review

bootjp:
Cosmetic nit (round-3 follow-up): deduped the doc-comment sentence "tests in the kv package can drive eviction-trigger scenarios without adding a constructor option" — was duplicated across two paragraphs of `SetHistoryDepthForTest`.

Doc-only; no code change. Claude round-3 confirmed all 7 findings ✅ Fixed and the table is fully green.

@claude review

PR #810, opened 2026-05-23
backup: cmd/elastickv-snapshot-decode CLI + CHECKSUMS writer

bootjp:
## Summary

Adds the Phase 0a operator-facing binary on top of the dispatcher in #806. Plus the CHECKSUMS writer / verifier and the `<uint64>.fsm` filename → snapshot-index parser the CLI needs to stamp MANIFEST.json.

Stacked on `backup/snapshot-dispatcher` — once #806 merges, this PR's diff collapses to just the CLI + helpers and I'll retarget to `main`.

## What lands

| File | Purpose |
|---|---|
| `cmd/elastickv-snapshot-decode/main.go` | Flag parsing, `DecodeSnapshot` invocation, `MANIFEST.json` + `CHECKSUMS` emission |
| `cmd/elastickv-snapshot-decode/main_test.go` | End-to-end synthetic-snapshot golden path + four flag-validation paths + `parseAdapterSet` table test |
| `internal/backup/checksums.go` | `WriteChecksums(root)` / `VerifyChecksums(root)` — sha256sum(1)-compatible |
| `internal/backup/source.go` | `SnapshotIndexFromPath` — `<uint64>.fsm` filename parser used to stamp `MANIFEST.snapshot_index` |
| `internal/backup/checksums_test.go`, `source_test.go` | Pin file shape, deterministic ordering, tamper detection, parse table |

## CLI surface

```
elastickv-snapshot-decode \
  --input  <fsm-file>          (required)
  --output <directory-root>    (required)
  [--adapter dynamodb,s3,redis,sqs|all]
  [--cluster-id <id>]
  [--include-incomplete-uploads]
  [--include-orphans]
  [--preserve-sqs-visibility]
  [--include-sqs-side-records]
  [--rename-collisions]
  [--dynamodb-bundle-mode per-item|jsonl]
```

Matches the design's §"Tooling" decoder section, minus the deferred `--input-snap-dir`, `--scope`, and `--dynamodb-bundle-size` flags (called out in the design as later additions).

## Output shape

`MANIFEST.json` carries `phase = "phase0-snapshot-decode"`, the filename-derived `snapshot_index`, the dispatcher-surfaced `last_commit_ts`, exclusion flags from the CLI, and a non-nil `Adapter{}` pointer for every enabled adapter (per-scope enumeration is a follow-up).

`CHECKSUMS` is the standard `sha256sum --binary` shape — verifiable with `sha256sum -c CHECKSUMS` from the dump root, no elastickv binary involved.

[...]

PR #896, opened 2026-06-01
docs(backup): propose Phase 0b M6 - cmd/elastickv-snapshot-encode CLI

bootjp:
## Summary

Phase 0b M6 capstone design doc. M1–M5 are merged; this proposal pins down the CLI implementation specifics the parent design doc ([`2026_05_25_proposed_snapshot_logical_encoder.md`](../blob/main/docs/design/2026_05_25_proposed_snapshot_logical_encoder.md)) left at sketch level.

## What lands

`docs/design/2026_06_01_proposed_snapshot_encode_cli.md` — new proposal covering:

- **CLI surface**: `--input`, `--output`, `--adapter` (mirrors decoder parsing), `--last-commit-ts`, `--self-test`, `--scratch-root`. Decoder-equivalent exit codes (0 success / 1 user error / 2 data error).
- **`--last-commit-ts` override**: fail-closed `T ≥ manifest.last_commit_ts` validation per parent §"MVCC re-encoding"; preserves the uniform-stamping invariant (single substitution everywhere `last_commit_ts` appears).
- **`ENCODE_INFO.json` sidecar**: encoder version, key-format version, output `.fsm` SHA-256, override record, manifest cluster_id propagation, self-test result. Sidecar (not inside the `.fsm` — EKVPBBL1 byte format is fixed).
- **Round-trip self-test**: `--self-test` runs encode → decode → structural diff against the input tree, exits 2 with `mismatch.txt` on first-N-differing-paths if the diff is non-empty. Default off (doubles encode time and disk), but CI runs it on every encoder PR.
- **Decoder cleanup folded in**: parent doc flags `Source.FSMCRC32C` as dead (native `.fsm` has no CRC32C footer); recommended path is repurpose-as-SHA-256, fallback is remove.
- **Test plan**: 8 tests covering round-trip success, override validation (lower rejected, equal+higher accepted), scope parsing, missing-manifest, corruption detection, and `ENCODE_INFO.json` round-trip + version gate.

## Why doc-first

Per CLAUDE.md "Design-doc-first workflow" — M6 touches a new entrypoint (`cmd/elastickv-snapshot-encode` binary) plus a new file format (`ENCODE_INFO.json`), so the proposal lands first; implementation is a separate follow-up PR after this is reviewed and accepted.

## Test plan

- [x] Doc reviewable on its own — every claim cites a concrete file:line (decoder CLI prior art at `cmd/elastickv-snapshot-decode/main.go`, encoder constructors at `internal/backup/encode_*.go`, manifest discipline at `WriteManifest` per gemini r1 #810).
- [x] Decision gates explicit: `--last-commit-ts T < manifest` fail-closed; self-test default-off with reasoning.
- [x] 5-pass self-review included (data loss / concurrency / performance / consistency / test coverage).
- [ ] Implementation lands in a follow-up PR (single slice; CLI + ENCODE_INFO + self-test are tightly coupled).


bootjp:
@claude review

bootjp:
v2 (fb6ba10f): addresses all 4 review findings on the design.

**Gemini medium #1 (sidecar collision):** ENCODE_INFO.json is now path-derived: `<output>.encode_info.json`. Multiple .fsm files in the same dir produce distinct sidecars.

**Gemini medium #2 (scratch-root collision):** Self-test scratch is ALWAYS a unique `encode-self-test-<random>` subdir under the resolved base, even when `--scratch-root` is operator-pinned. Concurrent encodes can share a pinned base safely.

**Gemini medium #3 (cleanup):** Eager `os.RemoveAll` on every exit path (success / mismatch / error) via defer at MkdirTemp time; stale mismatch.txt removed at start of every run.

**Codex P2 (MANIFEST.json absent from scratch):** Self-test now uses `DecodeSnapshot`'s in-memory `DecodeResult.Header.LastCommitTS` for the header check (vs. comparing scratch MANIFEST.json, which the library doesn't emit). Tree diff is adapter-subdir byte-equal; last_commit_ts is covered by the header check.

Test plan grew from 8 to 12 — 4 new tests pin each regression above. No impact on already-merged M1-M5 slices.

@claude review

bootjp:
v3 (a253d7f7): addresses both codex P2 findings on v2.

**Codex P2 #1 (write-temp-first):** New write-then-rename atomic-publish discipline. Encoder writes to `<output>.tmp-<random>`, fsync+close, runs self-test, ONLY renames to `<output>` on success. Self-test failure → temp file removed → no .fsm at the publish path. Pinned by `TestCLISelfTestFailureLeavesNoFsmAtOutputPath`.

**Codex P2 #2 (library entrypoint exposure):** Added `backup.EncodeSnapshot(EncodeOptions, io.Writer) (EncodeResult, error)` mirroring the decoder's `DecodeSnapshot` symmetry. `snapshotBuilder` stays unexported; the CLI's responsibility shrinks to flag-parse + temp-file dance + sidecar. Future callers (Phase 1 live extractor, integration tests) get a tested API surface. Pinned by `TestEncodeSnapshotLibraryRoundTrip`.

Files-to-add and test plan updated: 14 tests now (was 12).

@claude review

bootjp:
v4 (9e295f54): addresses codex P2 v3 finding.

**Codex P2 (DecodeOptions threading):** The self-test now reads MANIFEST.json's `Exclusions.*` + `DynamoDBLayout` fields and threads them into the scratch `DecodeOptions`, so a tree produced with (e.g.) `--include-incomplete-uploads` or `--dynamodb-bundle-mode jsonl` round-trips without spurious diff. Older manifests that omit the fields fall back to decoder defaults. Pinned by `TestCLISelfTestPreservesManifestDecodeOptions`.

Also fixed two duplicate step-4 numbering in the self-test sequence.

Test plan grew from 14 to 15.

@claude review

bootjp:
v5 (bae03774): addresses codex P2 v4 finding.

**Codex P2 (RenameS3Collisions not in manifest):** Pre-existing decoder-side gap — `--rename-collisions` drives `DecodeOptions.RenameS3Collisions` but the flag's value is NOT recorded in MANIFEST.json's `Exclusions`. Self-test could not perfectly round-trip dumps that used the flag.

Fix folded into M6 (same precedent as parent doc M1 "Decoder cleanup folded into M1"):
- `Exclusions` gains `RenameS3Collisions bool` (JSON tag `rename_s3_collisions`)
- Decoder CLI `emitManifest` populates the new field
- Encoder self-test reads it back and threads into scratch DecodeOptions
- Legacy manifests without the field decode as `false` (decoder default), no fail

Test plan grew from 15 to 17 — `TestCLISelfTestPreservesRenameS3Collisions` pins the full round-trip; `TestExclusionsLegacyManifestOmitsRenameS3Collisions` pins forward-compat for older manifests.

@claude review

bootjp:
v6 (be063d61): addresses 1 new medium + 12 still-open items from claude v3/v4/v5 reviews.

**Mandatory items resolved:**
- **Lifecycle rename:** `git mv` parent doc from `proposed_snapshot_logical_encoder.md` → `partial_snapshot_logical_encoder.md`. M1-M5 are merged; the rename was overdue.
- **exclusionsRequiredFields policy:** `rename_s3_collisions` intentionally NOT in required fields; doc states why (backward compat) + mechanism (JSON zero-value fallback). Asymmetry with the 4 pre-existing required fields is intentional.
- **io.Writer self-test buffer:** committed to Option A — buffer FSM in `*bytes.Buffer` when `SelfTest=true`, decode from buffer, copy to `out` only on match. No-self-test path streams straight to `out`.
- **`--s3-chunk-size`:** explicitly deferred with rationale (no in-tree deployment uses non-default).

**Polish items (all flagged across 3 review rounds):**
- Exit code 2 rationale for flag-driven data constraint
- Line 11 vestigial "plus the assembled .fsm bytes" → removed
- `DynamoDBBundleJSONL` correct field name + mapping
- `TestCLISelfTestDetectsCorruption` injection mechanism (`corruptingWriter`)
- `TestEncoderProducesLoadableSnapshot` deferred from M6 with rationale
- `encoder_version` ldflags mechanism mirrors decoder `main.go:45`
- `adapters_enabled` canonical encoder fan-out order
- Decoder cleanup M1→M6 attribution
- Parameter order asymmetry (`EncodeSnapshot(opts, out)` vs `DecodeSnapshot(r, opts)`) noted as deliberate

@claude review

bootjp:
v7 (ac7af858): addresses codex P2 v6 — sharp catch on the corruption-injection mechanism.

The v6 `corruptingWriter` wrapping `out` is incoherent with the buffering model: `EncodeSnapshot` runs self-test against its internal `*bytes.Buffer` and only copies to `out` AFTER the self-test matches. So corruption at `out` happens after self-test has already passed — the test would silently report exit 0.

New mechanism: same-package test-only hook on `EncodeOptions` (unexported `corruptBufferForTest func([]byte)`) that runs against the internal buffer between `WriteTo` and the self-test decode. Corruption reaches the decode but never reaches `out` (write-then-rename discipline preserved). Hook is unexported → external callers cannot set it → no corrupted-output API on the public surface.

@claude review

bootjp:
v8 (e2938a62): both v7 findings addressed.

**Mandatory (v6-still-open):** Line 4 `Parent:` link updated from `2026_05_25_proposed_snapshot_logical_encoder.md` to `2026_05_25_partial_snapshot_logical_encoder.md`, matching the v6 `git mv`. Added a sentence noting the lifecycle promotion happened in v6.

**Minor (v7 new):** Renamed `TestCLISelfTestDetectsCorruption` → `TestEncodeSnapshotSelfTestDetectsCorruption`. The unexported `corruptBufferForTest` hook requires same-package access (`internal/backup/encode_snapshot_test.go` as `package backup`), so the `CLI` prefix was misleading.

@claude review

bootjp:
v9 (653d928c): claude v8 minor — parent doc body header now reflects the partial-lifecycle status.

`Status: Partial (M1-M5 merged via PRs #807/#841/#847/#849/#864/#846/#892; M6 CLI in review at PR #896)` — matches the filename and CLAUDE.md's partial-doc convention.

@claude review

bootjp:
v10 (17364356): folded in the only remaining non-blocking minor — explicit note on encoder fan-out (`redis` first) vs decoder finalize order (`dynamodb` first per `decode.go:278`), with the rationale that the final `.fsm` byte sequence is determined by encoded-key sort, not fan-out order.

Claude v9 verdict was already "ready to merge"; v10 just closes the last carried-over minor so the doc is fully clean.

@claude review

PR #899, opened 2026-06-01
feat(encryption): Stage 6E-1a — FSM apply machinery for enable-raft-envelope cutover

bootjp:
## Summary

Stage 6E-1a — FSM apply machinery for the §7.1 Phase-2 raft-envelope cutover. Builds on the design merged in #893.

**Scope**: FSM-side only. The admin RPC + server + CLI ship in 6E-1b (separate PR, deferred to keep this slice off the protoc toolchain version pin in CLAUDE.md).

### What lands

- `RotateSubEnableRaftEnvelope byte = 0x05` — the previously-reserved sub-tag is promoted to live; wire decoder whitelist updated.
- `validateEnableRaftEnvelopePayload` — mirrors the storage variant for the raft slot (Purpose=PurposeRaft, empty-Wrapped length check, proposer-DEK pinning).
- `applyEnableRaftEnvelope` — structural mirror of `applyEnableStorageEnvelope`. Records the cutover via `sc.RaftEnvelopeCutoverIndex` (a uint64 index — non-zero ⇒ "Phase-2 active"; doubles as the idempotency token, no separate bool flag). Registration-before-sidecar ordering matches the storage variant's crash-recovery invariant.
- `ApplyRotation` dispatch case for the new sub-tag.

### Outcomes (mirror the storage variant)

| Case | Treatment |
|---|---|
| Malformed (purpose / Wrapped / proposer-DEK) | Halt-apply via `ErrEncryptionApply` |
| Stale DEKID (RotateDEK race) | Benign no-op; advance `RaftAppliedIndex` only |
| Already active (duplicate cutover) | Idempotent; preserve original `RaftEnvelopeCutoverIndex` |
| Fresh success | Register proposer → set cutover + advance applied inside one fsync |

### Operator-inertness in 6E-1a

Stage 6E-1a deliberately does NOT activate the §6.3 engine apply-hook unwrap or the coordinator wrap-on-propose switch — those land atomically in 6E-2 (the design notes 6E-2's three pieces cannot ship separately because either ordering produces cluster-wide halt-apply at cutover). With only 6E-1a deployed, the sidecar cutover field advances on apply but no entry is wrapped or unwrapped, matching the design's "no behavior change" guarantee for 6E-1.

### Verification

- 6 new tests under `TestApplyRotation_EnableRaftEnvelope_*`. All pass under `-race`.
- Full `internal/encryption/...` `-race` clean (32s).
- `golangci-lint --config=.golangci.yaml run ./internal/encryption/...` 0 issues.

### Self-review (5-lens) — full breakdown in commit body

1. **Data loss** — registration-before-sidecar ordering preserves crash-recovery invariant.
2. **Concurrency** — FSM apply is serial per Raft group; no race surface added.
3. **Performance** — one extra sub-tag dispatch case; hot-path unchanged.
4. **Data consistency** — strict-`>` is a 6E-2 concern; 6E-1a only writes the field. FSM determinism unaffected (each replica reaches the same value at the same apply index).
5. **Test coverage** — every validate branch + every apply state-machine branch has a dedicated test; 6D regression tests stay green.

## Test plan

- [x] `go test -race ./internal/encryption/...` clean
- [x] `golangci-lint --config=.golangci.yaml run ./internal/encryption/...` clean
- [x] 6 new tests + all pre-6E-1a tests green
- [ ] @claude review for the apply state machine, the registration-before-sidecar ordering rationale, and the "operator-inertness in 6E-1a" guarantee

## Summary by CodeRabbit

* **New Features**
  * Added support for raft-layer encryption envelope cutover with wire-layer acceptance of the new cutover operation, strict payload validation, and proposer-registration persistence.
  * Cutover is safe and idempotent: stale or duplicate cutovers preserve existing state while still advancing apply progress.

* **Tests**
  * Added comprehensive tests covering success paths, validation failures, edge-case rejections, and idempotency/race behaviors.

### DIFF
--- a/internal/encryption/applier.go
+++ b/internal/encryption/applier.go
@@ -826,11 +826,17 @@ func advanceRaftAppliedIndex(sc *Sidecar, raftIdx uint64) {
 //   - RotateSubEnableStorageEnvelope — one-shot storage-layer
 //     cutover (Stage 6D-4). Flips StorageEnvelopeActive and records
 //     StorageEnvelopeCutoverIndex inside a single sidecar fsync.
+//   - RotateSubEnableRaftEnvelope — one-shot raft-layer cutover
+//     (Stage 6E-1). Records RaftEnvelopeCutoverIndex inside a
+//     single sidecar fsync. The engine apply-hook installed by
+//     6E-2 dispatches `entry.Index > sidecar.RaftEnvelopeCutoverIndex`
+//     through the unwrap path; strict `>` makes the cutover entry
+//     itself (at index == cutover) flow through unwrap-free, which
+//     is the chicken/egg bootstrap.
 //
-// Other sub-tags (rewrap-deks, retire-dek, enable-raft-envelope)
-// land in later stages and return ErrEncryptionApply so the
-// HaltApply seam fires on an unrecognised sub-tag rather than
-// silently advancing setApplied.
+// Other sub-tags (rewrap-deks, retire-dek) land in later stages
+// and return ErrEncryptionApply so the HaltApply seam fires on an
+// unrecognised sub-tag rather than silently advancing setApplied.
 //
 // Same WithKEK / WithKeystore / WithSidecarPath trio requirement
 // as ApplyBootstrap; partial wiring returns ErrKEKNotConfigured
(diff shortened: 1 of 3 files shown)


bootjp:
@claude review

Round-1 fix in a175d097:

- **coderabbit Major** + **gemini MEDIUM** (same defect) — `applyEnableRaftEnvelope` accepted `raftIdx == 0`, would persist the proposer-registration but leave `RaftEnvelopeCutoverIndex` at 0 (Phase-2 silently inactive) and replays would re-enter fresh-success since the already-active short-circuit only fires on `!= 0`. Now fail-closed before any sidecar mutation: `errors.Wrap(ErrEncryptionApply, "applier: enable-raft-envelope requires a non-zero raft index")`.

Regression test: `TestApplyRotation_EnableRaftEnvelope_RejectsZeroRaftIndex` (added in this commit, verified to fail on the prior tree).

Caller audit (per the project's semantic-change-requires-grep directive): `ApplyRotation` has a single production caller at `kv/fsm_encryption.go:136`, which wraps any error with `haltErr(... encryption.ErrEncryptionApply)` to halt apply — the new error path matches existing payload-validation rejects exactly.

Note on storage sibling: `applyEnableStorageEnvelope` does NOT need the same check because `StorageEnvelopeActive` is a separate bool — even with a zero index, activation would flip to true. The raft variant lacks that second sentinel, hence the variant-specific guard.

PR #900, opened 2026-06-01
kv(composed1): M4 — Dispatch retry on ErrComposed1Violation / ErrComposed1VersionGCd

bootjp:
## Summary

Fourth milestone (**M4**) of the Composed-1 cross-group commit-time guard per `docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md` §M4.

Makes the M3 apply-time gate **non-disruptive to clients**: instead of surfacing `ErrComposed1Violation` or `ErrComposed1VersionGCd` back to the adapter (and thus the application), the coordinator absorbs the sentinel, re-reads the catalog version, and re-issues the txn once against the post-shift owning group.

### `ShardedCoordinator.Dispatch` — pin ObservedRouteVersion at entry

Every transactional `OperationGroup` that arrives with `ObservedRouteVersion == 0` (the M1 unpinned sentinel) gets stamped with `engine.Version()`. This makes the M3 gate effective **by default** — every txn flowing through the sharded coordinator is now gate-eligible without each adapter rediscovering the §4.1 contract.

A caller that supplies a non-zero `ObservedRouteVersion` (a future adapter migration target managing the version itself) still wins over the Dispatch-entry pin.

### New `dispatchTxnWithComposed1Retry` — retry-once on Composed-1 sentinels

Wraps the existing single-attempt `dispatchTxn` in a loop bounded at `composed1RetryAttempts = 1` (initial + 1 retry = 2 attempts).

On the sentinel:
- Re-read `engine.Version()` and stamp the fresh value on the OperationGroup
- Clear `StartTS` / `CommitTS` so the next attempt allocates a fresh HLC-correct pair via `nextStartTS` (OCC invariants require `commitTS > startTS` under the same HLC observation)
- `groupMutations` on the next dispatchTxn pass re-runs `router.ResolveGroup`, so a key whose owning group changed naturally lands on the new group's FSM

Beyond one retry the sentinel surfaces unchanged so a wrapping adapter retry harness (or the client) can decide whether to keep trying — preventing the coordinator from spinning when catalog churn outpaces the dispatch.

## Caller audit (loop directive's semantic-change check)

`ShardedCoordinator.Dispatch` is the entry point every adapter calls. The change:
- **Surfaces success more often** (turning a previous error into a successful commit)
- **Delays sentinel surfacing** by a bounded retry budget

`grep -rn 'ErrComposed1' adapter/` returns no hits — no adapter relies on a faster-than-bounded surfacing of these sentinels (M3 only landed yesterday).

## Tests (5 new)

- `TestShardedCoordinator_RetriesOnComposed1Violation` — happy path: first attempt fails, second succeeds, caller sees success.
- `TestShardedCoordinator_RetriesOnComposed1VersionGCd` — same retry path drives both sentinels.
- `TestShardedCoordinator_PersistentComposed1ViolationSurfaces` — retry budget bounds the loop: persistent failure surfaces sentinel after exactly 2 total Commit calls.
- `TestShardedCoordinator_PinsObservedRouteVersionAtDispatchEntry` — Dispatch pins from `engine.Version()` when caller leaves it at zero.
- `TestShardedCoordinator_HonorsCallerPinnedObservedRouteVersion` — caller-supplied non-zero value wins (§4.1 contract preserved).

## Verification

- `go build ./...` — clean
- `go test -race -count=1 ./kv` — 10.3s, pass
- `make lint` — 0 issues

## Self-review (5 lenses)

1. **Data loss** — retry can only convert a previously-rejected commit into success OR surface the same error after a bounded loop. Cannot lose a committed write (each attempt allocates fresh commitTS; FSM's OCC validation rejects duplicates).
2. **Concurrency** — no new locks. Retry runs serially on dispatcher goroutine; `engine.Version()` takes the existing read lock per call.
3. **Performance** — common case: single iteration + one branch + one `engine.Version()` read at entry. Sentinel case: one extra `dispatchTxn` round-trip.
4. **Data consistency** — closes §M4. With M1+M2+M3+M4 landed, a `MoveRange` / cross-group `SplitRange` between BeginTxn and Commit is now **transparent to the client**: coordinator absorbs the gate's rejection, re-routes, succeeds on the new owner.
5. **Test coverage** — 5 unit tests cover both sentinels, retry budget, Dispatch-entry pin, and caller-pin precedence. Existing tests pass unchanged.

## Test plan

- [x] `go test -race ./kv`
- [x] `make lint`
- [ ] (Follow-up) **M5** — Jepsen workload (route-shuffle generator for the DynamoDB suite, find no G1c anomaly under concurrent MoveRange).

## Resolves

The M4 row in the Composed-1 design doc.

## Summary by CodeRabbit

* **New Features**
  * Transaction dispatch now performs a single bounded retry for specific transient composed-1 errors, preserving previous commit timestamps; caller-supplied start timestamps suppress the retry. Write-only transactions may auto-pin route version at dispatch, while multi-shard commit-phase secondary requests intentionally omit the pinned route version.

* **Tests**
  * Expanded test coverage for composed-1 retry semantics, PrevCommitTS preservation, caller-supplied StartTS behavior, route-version pinning rules, and secondary commit behavior.

### DIFF
--- a/kv/sharded_coordinator.go
+++ b/kv/sharded_coordinator.go
@@ -565,14 +565,35 @@ func (c *ShardedCoordinator) dispatchTxnWithComposed1Retry(ctx context.Context, 
 		// so the retry would run with StartTS=0 — violating MVCC
 		// invariants (gemini HIGH on PR #900).
 		reqs.CommitTS = 0
-		// Drop the option-2 dedup probe.  The M3 gate fired at apply
-		// time, so the ORIGINAL attempt never committed — there is no
-		// prior landing for PrevCommitTS to dedup against on this
-		// retry.  Leaving it set would let dispatchMultiShardTxn:736
-		// short-circuit with ErrTxnDedupRequiresSingleShard if the
-		// post-shift catalog now reroutes this txn to span two groups
-		// (claude[bot] medium, 3rd review round on PR #900).
-		reqs.PrevCommitTS = 0
+		// IMPORTANT — do NOT clear reqs.PrevCommitTS here.  An earlier
+		// round of this PR did (taking claude[bot]'s reasoning at face
+		// value: "M3 fired at apply time, so the original attempt never
+		// committed → probe is stale").  That reasoning conflates two
+		// different "original attempts" and is wrong (codex P1 on
+		// 43c55dfe, PR #900):
+		//
+		//   * Within M4: attempt 1's M3 rejection does prove attempt 1
+		//     did not write anything (the FSM rejects before any state
+		//     change).
+		//   * Across adapter retries: PrevCommitTS names the
+		//     ADAPTER'S prior attempt (e.g. adapter/redis.go:3209,
+		//     :3615 send PrevCommitTS = pending.commitTS as the
+		//     option-2 ambiguous-outcome probe).  That earlier
+		//     attempt may have actually landed at the named commitTS
+		//     before the route shift that triggered our M3 sentinel.
+		//
+		// Dropping the probe in the retry would let the FSM re-apply
+		// the writes against the post-shift route, double-writing the
+		// caller's at-most-once-payload if the named prior attempt did
+		// land.  Preserving it has the cost that
+		// dispatchMultiShardTxn:749 surfaces
+		// ErrTxnDedupRequiresSingleShard when the post-shift route now
+		// spans multiple groups — which IS the faithful signal: the
+		// adapter's at-most-once retry contract cannot be transparently
+		// honoured across a shard split, and the caller must reckon
+		// with the ambiguity at a higher level (re-read, re-execute,
+		// abort).  Surfacing that error is strictly safer than silently
+		// losing dedup.
 		startTS, allocErr := c.nextStartTS(ctx, reqs.Elems)
 		if allocErr != nil {
 			// resp here is the failed first attempt's response (always
--- a/kv/sharded_coordinator.go
+++ b/kv/sharded_coordinator.go
@@ -505,10 +505,11 @@ func (c *ShardedCoordinator) Dispatch(ctx context.Context, reqs *OperationGroup[
 // maybeAutoPinObservedRouteVersion stamps reqs.ObservedRouteVersion
 // with the engine's current catalog version at Dispatch entry, but
 // ONLY when (a) the txn has no read keys AND (b) the caller did not
-// supply StartTS.  Per design doc §4.1 the version must be pinned
-// at BeginTxn (read-set capture time); the auto-pin is a
-// best-effort backfill for callers that have not yet migrated to
-// pin themselves.  Two cases skip the auto-pin:
+// supply StartTS AND (c) no element's key is claimed by a partition
+// resolver.  Per design doc §4.1 the version must be pinned at
+// BeginTxn (read-set capture time); the auto-pin is a best-effort
+// backfill for callers that have not yet migrated to pin
+// themselves.  Three cases skip the auto-pin:
 //
 //   - len(ReadKeys) != 0 — the caller already performed reads at
 //     some earlier moment (Redis MULTI/EXEC's txnContext.commit
@@ -532,6 +533,20 @@ func (c *ShardedCoordinator) Dispatch(ctx context.Context, reqs *OperationGroup[
 //     auto-pin gives false confidence (codex P1 on 144ec0ca,
 //     PR #900).
 //
+//   - any element's key is resolver-claimed — when a
+//     PartitionResolver is wired (SQS HT-FIFO partition keys
+//     etc.), ShardRouter routes via the resolver but
+//     verifyComposed1 calls route-history OwnerOf on the raw
+//     mutation key against the BYTE-RANGE engine snapshot.  The
+//     engine has no knowledge of the resolver's mapping, so the
+//     gate spuriously rejects resolver-routed commits even when
+//     the resolver picked the correct gid.  Skip the auto-pin
+//     for any resolver-recognised key; the request flows with
+//     ObservedRouteVersion=0 and the M3 gate short-circuits —
+//     restoring the pre-auto-pin behaviour for resolver-routed
+//     txns.  Resolver-aware M3 is M5+ work (codex P1 on
+//     6a458a28, PR #900).
+//
 // The non-auto-pin case (request flows with ObservedRouteVersion=0,
 // M3 gate short-circuits) is the safe non-regressing posture for
 // non-migrated callers — the gate cannot retroactively pin reads
@@ -544,7 +559,31 @@ func (c *ShardedCoordinator) maybeAutoPinObservedRouteVersion(reqs *OperationGro
 	if c.engine == nil || reqs.ObservedRouteVersion != 0 || len(reqs.ReadKeys) != 0 || callerSuppliedStartTS {
 		return
 	}
+	if c.anyResolverClaimedKey(reqs.Elems) {
+		return
+	}
 	reqs.ObservedRouteVersion = c.engine.Version()
+}
+
+// anyResolverClaimedKey reports whether any element's key is
+// claimed by the partition resolver.  Returns false when the
+// router has no resolver installed (the most common case) so the
+// auto-pin path stays inexpensive in the byte-range-only deployment.
+// Extracted from maybeAutoPinObservedRouteVersion to keep the
+// auto-pin's cyclomatic complexity in the cyclop budget.
+func (c *ShardedCoordinator) anyResolverClaimedKey(elems []*Elem[OP]) bool {
+	if c.router == nil || c.router.partitionResolver == nil {
+		return false
+	}
+	for _, e := range elems {
+		if e == nil || len(e.Key) == 0 {
+			continue
+		}
+		if c.router.partitionResolver.RecognisesPartitionedKey(e.Key) {
+			return true
+		}
+	}
+	return false
 }
 
 func (c *ShardedCoordinator) dispatchTxnWithComposed1Retry(ctx context.Context, reqs *OperationGroup[OP], callerSuppliedStartTS bool) (*CoordinateResponse, error) {
--- a/kv/sharded_coordinator.go
+++ b/kv/sharded_coordinator.go
@@ -851,7 +851,13 @@ func (c *ShardedCoordinator) dispatchMultiShardTxn(ctx context.Context, startTS,
 		return nil, errors.WithStack(err)
 	}
 
-	maxIndex = c.commitSecondaryTxns(ctx, startTS, primaryGid, primaryKey, grouped, gids, commitTS, maxIndex, observedRouteVersion)
+	// observedRouteVersion is intentionally NOT propagated to
+	// commitSecondaryTxns — secondary commits drop the M3 gate to
+	// avoid the best-effort-secondary + fail-closed-gate silent
+	// partial commit hazard (codex P1 on 6202b964, PR #900).  See
+	// the Request construction inside commitSecondaryTxns for the
+	// full reasoning.
+	maxIndex = c.commitSecondaryTxns(ctx, startTS, primaryGid, primaryKey, grouped, gids, commitTS, maxIndex)
 	return &CoordinateResponse{CommitIndex: maxIndex}, nil
 }
 
@@ -991,7 +997,7 @@ func (c *ShardedCoordinator) commitPrimaryTxn(ctx context.Context, startTS uint6
 	return primaryGid, r.CommitIndex, nil
 }
 
-func (c *ShardedCoordinator) commitSecondaryTxns(ctx context.Context, startTS uint64, primaryGid uint64, primaryKey []byte, grouped map[uint64][]*pb.Mutation, gids []uint64, commitTS uint64, maxIndex uint64, observedRouteVersion uint64) uint64 {
+func (c *ShardedCoordinator) commitSecondaryTxns(ctx context.Context, startTS uint64, primaryGid uint64, primaryKey []byte, grouped map[uint64][]*pb.Mutation, gids []uint64, commitTS uint64, maxIndex uint64) uint64 {
 	// Secondary commits are best-effort. If a shard is unavailable after the
 	// primary commits, read-time lock resolution will commit the remaining
 	// secondaries based on the primary commit record. Retry a few times to
@@ -1006,11 +1012,30 @@ func (c *ShardedCoordinator) commitSecondaryTxns(ctx context.Context, startTS ui
 			continue
 		}
 		req := &pb.Request{
-			IsTxn:                true,
-			Phase:                pb.Phase_COMMIT,
-			Ts:                   startTS,
-			Mutations:            append([]*pb.Mutation{meta}, keyMutations(grouped[gid])...),
-			ObservedRouteVersion: observedRouteVersion,
+			IsTxn:     true,
+			Phase:     pb.Phase_COMMIT,
+			Ts:        startTS,
+			Mutations: append([]*pb.Mutation{meta}, keyMutations(grouped[gid])...),
+			// ObservedRouteVersion intentionally omitted (zero) on
+			// secondary commits.  commitSecondaryTxns is best-effort:
+			// errors are logged and the dispatch acknowledges success
+			// while the lock-resolution backstop is expected to commit
+			// the secondary later.  If we propagated the auto-pinned
+			// version here, a route shift between primary-COMMIT and
+			// secondary-COMMIT would let the secondary FSM return
+			// ErrComposed1Violation, which commitSecondaryTxns swallows
+			// — turning route churn into a silent partial commit
+			// visible to no readers on the new owner (codex P1 on
+			// 6202b964, PR #900).  The FSM's verifyComposed1
+			// short-circuits on ObservedRouteVersion=0 (fsm.go:609), so
+			// dropping the field here restores the pre-66ad5b0 "no
+			// gate on commit-phase secondaries" behaviour without
+			// affecting prewrite (still gated) or primary commit
+			// (still gated; abortPreparedTxn cleans up on failure).
+			// Until a fatal-secondary 2PC protocol or
+			// lock-resolution-bypass design lands (M5+), this is the
+			// honest posture: gate where we can cleanly abort, skip
+			// where we cannot.
 		}
 		r, err := commitSecondaryWithRetry(ctx, g, req)
 		if err != nil {
--- a/kv/sharded_coordinator.go
+++ b/kv/sharded_coordinator.go
@@ -134,6 +134,16 @@ const (
 
 	txnSecondaryCommitRetryAttempts = 3
 	txnSecondaryCommitRetryBackoff  = 20 * time.Millisecond
+
+	// composed1RetryAttempts bounds how many times Dispatch re-issues a
+	// txn after the FSM's M3 verifyComposed1 gate returned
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

bootjp:
Both critical-class findings addressed:

**gemini CRITICAL / codex P1** (line 503/519 — bumped StartTS breaks SSI for read-write txns): Now surfaces `ErrComposed1Violation` / `ErrComposed1VersionGCd` IMMEDIATELY (no retry) when `len(reqs.ReadKeys) > 0`. The client already performed reads at the original StartTS — silently bumping it would let concurrent writes between old/new timestamps escape OCC validation, violating SSI. Write-only txns (no reads to invalidate) remain safe to retry, so M4 is still client-transparent for the most common cross-group SplitRange / MoveRange case.

**gemini HIGH** (line 526 — nil-clock guard left StartTS=0): `nextStartTS` already handles nil clock internally; removed the `if c.clock != nil` gate so the retry allocates StartTS unconditionally, matching the unconditional call pattern Dispatch itself uses.

**Caller audit** (loop directive): `grep -rn 'ErrComposed1' adapter/` → no hits. M3 only landed yesterday and no adapter matches on these sentinels, so the earlier surfacing for read-write txns is semantics-safe. When adapters do start matching, the new behaviour is what they'd want (entire txn re-execution, not a coordinator-side bump).

**Regression test** (added before fix per convention): `TestShardedCoordinator_DoesNotRetryReadWriteTxnOnComposed1` — read-write txn with ErrComposed1Violation surfaces immediately, exactly 1 Commit call. Verified failing on parent / passing on fix.

`go test -race ./kv` pass / `make lint` 0 issues.

@claude review

bootjp:
Codex P1 addressed:

The Dispatch-entry auto-pin no longer fires for read-write txns. Per design doc §4.1, pinning must happen at BeginTxn (read-set capture); for read-write txns Dispatch is strictly LATER than BeginTxn, so the coordinator cannot retroactively reconstruct the route view the reads were taken against. Stamping the catalog version at EXEC time would associate post-shift routes with pre-shift reads — exactly the gate-bypass codex flagged for Redis MULTI/EXEC.

Fix: gate the auto-pin on `len(reqs.ReadKeys) == 0`. Write-only txns continue to be auto-pinned (Dispatch == BeginTxn equivalently). Read-write callers must pin ObservedRouteVersion themselves at BeginTxn time — until they migrate, the observedVer=0 sentinel surfaces to M3 which short-circuits, matching the pre-feature behaviour byte-for-byte.

**Caller audit** (loop directive): `grep -rn 'ObservedRouteVersion' adapter/` → no hits. No adapter currently sets or reads the field, so the auto-pin shrink is invisible.

**Regression test** (added before fix per convention): `TestShardedCoordinator_DoesNotAutoPinObservedRouteVersionForReadWriteTxn` — read-write OperationGroup with ObservedRouteVersion=0 against engine pinned at v=5 must surface observedVer=0 (auto-pin must NOT fire). Verified failing on parent / passing on fix.

`go test -race ./kv` pass / `make lint` 0 issues.

@claude review

bootjp:
Pushed as 10123c5a — the @claude review request above was posted before the commit actually landed (a pre-commit cyclop lint failure on my first attempt — fixed by extracting the auto-pin into `maybeAutoPinObservedRouteVersion`). The semantic fix is unchanged from the previous comment; please pick up 10123c5a.

@claude review

bootjp:
Pushed as 43c55dfe — claude[bot]'s three items on 10123c5a addressed.

**Medium — PrevCommitTS leak (3rd review round):** retry block now zeros `reqs.PrevCommitTS` alongside `reqs.CommitTS`. The M3 gate fires at apply time, so the original attempt never committed → the option-2 dedup probe is invalid on retry. Without the clear, a write-only retry that the post-shift catalog now reroutes across two groups would hit `ErrTxnDedupRequiresSingleShard` on a stale signal.

**Caller audit** (semantic change, loop directive): `grep -rn 'PrevCommitTS' adapter/` → adapter/redis.go:3209, :3615. Both sites construct OperationGroup inline, pass to Dispatch, branch on the result. Neither reads `reqs.PrevCommitTS` back; the adapter-owned `pending.commitTS` is unmodified by Dispatch. Clearing inside the M4 retry is invisible to current callers.

**Minor — empty goroutine (3rd review round):** removed; intent promoted to a plain block comment in front of Dispatch pointing at the observedSeen[1] assertion that enforces the re-read contract.

**Nit — stale resp on allocErr:** `return nil, allocErr` (resp was always nil for Composed-1 sentinels, but explicit is better).

**Regression test** (added before fix per convention): `TestShardedCoordinator_ClearsPrevCommitTSOnComposed1Retry` — decodes the encoded TxnMeta from each attempt's pb.Request to observe on-wire PrevCommitTS, asserts attempt 1=17 (caller's probe unchanged) / attempt 2=0 (probe dropped). Verified failing on parent / passing on fix.

`go test -race ./kv` pass (full pkg, 11s) / `make lint` 0 issues.

@claude review

bootjp:
Pushed as 57da8886 — codex P1 on 43c55dfe addressed.

**Codex P1 was correct, claude[bot]'s prior medium was wrong.** 43c55dfe took claude[bot]'s reasoning at face value ("M3 fired at apply time → original attempt never committed → probe is stale") and zeroed `reqs.PrevCommitTS` on the M4 retry. That conflates two different "original attempts":

- **Within M4**: attempt 1's M3 rejection proves attempt 1 wrote nothing. ✓
- **Across adapter retries**: PrevCommitTS names the **adapter's** prior attempt (adapter/redis.go:3209, :3615 send `PrevCommitTS = pending.commitTS` as the option-2 ambiguous-outcome probe). That earlier attempt may have actually **landed** before the route shift that triggered this M4's M3 sentinel.

Dropping the probe would let the FSM re-apply against the post-shift route, double-writing the at-most-once payload if the prior attempt landed. The at-most-once contract is the whole point of option-2 dedup — M4 must not break it.

**Trade-off accepted**: post-shift-multi-shard case still surfaces `ErrTxnDedupRequiresSingleShard` from dispatchMultiShardTxn:749. That's the faithful signal — at-most-once retry cannot transparently cross a shard split; the caller must reckon with the ambiguity at a higher level. Surfacing is strictly safer than silent double-write.

**Caller audit** (loop directive): no caller reads `reqs.PrevCommitTS` back after Dispatch. Restoring the pre-43c55dfe wire contract is invisible to current callers.

**Regression test** replaces the misnamed prior one:
- `TestShardedCoordinator_PreservesPrevCommitTSOnComposed1Retry` — decodes TxnMeta from each per-attempt pb.Request mutation (same bytes the FSM reads), asserts attempt 1 **AND** attempt 2 both carry PrevCommitTS=17. This is the inverse assertion of what 43c55dfe pinned, encoding codex P1's at-most-once correctness contract.

`go test -race ./kv` pass (full pkg, 11s) / `make lint` 0 issues.

@claude review

bootjp:
Pushed as 144ec0ca — codex P1 on 57da8886 addressed (plus claude[bot] stale-comment nit).

**Codex P1 — caller-supplied StartTS surfaces.** The M4 retry path's StartTS bump was unsafe for ALL caller-supplied StartTS values, not just the already-gated ReadKeys-non-empty case. A code-base grep found **13+ adapter sites** that supply `StartTS = readTS` without populating ReadKeys (their pre-Dispatch metadata read is invisible to the coordinator):

- adapter/s3.go:573 (createBucket), s3_admin_objects.go:144 / :348
- adapter/sqs_messages.go:541 / :1113 / :1332 / :1487 / :1614
- adapter/sqs_messages_batch.go:222, sqs_fifo.go:241, sqs_purge.go:146, sqs_tags.go:179
- adapter/dynamodb.go:4366 / :8274

Under the pre-fix retry, any Composed-1 sentinel they hit would have been silently absorbed by the bumped-StartTS retry, letting any commit landed in (originalStartTS, newStartTS) bypass the FSM's `latest > startTS` OCC check.

**Fix:** capture `callerSuppliedStartTS := reqs.IsTxn && reqs.StartTS != 0` at Dispatch entry (BEFORE the coordinator-allocates-on-zero branch mutates the field), thread it into `dispatchTxnWithComposed1Retry`, surface the sentinel without retry when set. Coordinator-allocated StartTS remains retryable — by definition no read has happened at that timestamp.

**Caller audit** (loop directive): the 13+ adapter sites are the BENEFICIARIES of this fix. Pre-fix they were silently exposed to stale-snapshot OCC bypass under route shifts; post-fix they correctly surface ErrComposed1Violation through their existing HTTP error path. No caller depended on the silent retry succeeding (M4 was added in this PR; no adapter is yet wired to take advantage). DynamoDB sites with `StartTS: 0` (dynamodb.go:922, :981, etc.) continue to use the coordinator-allocated path — M4 retry remains available for those.

**Regression test:** `TestShardedCoordinator_SurfacesComposed1OnCallerSuppliedStartTS` — write-only OG with StartTS=42, ReadKeys empty, asserts the sentinel surfaces unchanged and exactly ONE Commit attempt fires. Failing on parent (retry absorbed); passing on fix.

**Existing M4-retry tests** updated to omit StartTS (was `StartTS: 10` test fixture, now coordinator-allocated path so retry path is eligible). Tests that exercise read-write surfacing keep StartTS=10 — ReadKeys non-empty surfaces first, so the new gate is never reached.

**claude[bot] nit on 57da8886:** stub's PrevCommitTS-decode comment still said "assert the field is cleared" (the now-disproven 43c55dfe-era claude[bot] medium reasoning). Updated to "PRESERVED across retry attempts" with codex P1 (43c55dfe) reference.

`go test -race ./kv` pass (full pkg, 11s) / `make lint` 0 issues.

@claude review

bootjp:
Pushed as 6202b964 — codex P1 on 144ec0ca addressed.

**Codex P1 — skip auto-pin for caller-supplied StartTS.** The auto-pin at Dispatch entry stamped `reqs.ObservedRouteVersion = engine.Version()` even for callers that supplied StartTS=readTS with empty ReadKeys (the same 13+ S3/SQS/DynamoDB sites from 144ec0ca). That stamps the CURRENT catalog version on a request whose read happened at an EARLIER catalog version — a route shift in (readTime, dispatchTime) becomes invisible to verifyComposed1, masking the failure mode the gate exists to catch.

**Why this is worse than no auto-pin:** ObservedRouteVersion=0 → gate short-circuits → operator knows protection is not in place. Spurious auto-pin → gate verifies a FALSE claim → false confidence. claude[bot]'s 144ec0ca framing of "partial coverage" is illusory because the auto-pin LIES about when the read happened — it's a corrupt signal, not a partial one.

**Fix:** thread the already-captured `callerSuppliedStartTS` one level down into `maybeAutoPinObservedRouteVersion`. The skip condition now reads:
```go
if engine == nil || ObservedRouteVersion != 0 || len(ReadKeys) != 0 || callerSuppliedStartTS { return }
```

**Caller audit** (loop directive): the 13+ caller-supplied-StartTS adapter sites lose the auto-pin's false-positive coverage but no real protection — they were never honestly covered. DynamoDB sites with `StartTS: 0` (dynamodb.go:922 / :981 / :1760 / :4686 / :4703 / :4722 / :4735) are unaffected — coordinator-allocated path still auto-pins, which is honest because the allocation IS the read-time-equivalent moment. Adapters that want genuine M3 protection must pin at read time per §4.1 (M5+ migration).

**Regression test:** `TestShardedCoordinator_DoesNotAutoPinObservedRouteVersionForCallerSuppliedStartTS` — StartTS=42, ReadKeys empty, engine at v=9, asserts observed=0. Failing on parent (was v=9) / passing on fix.

**Existing test** `PinsObservedRouteVersionAtDispatchEntry` updated to omit StartTS so it stays on the coordinator-allocated branch (still auto-pins). `HonorsCallerPinnedObservedRouteVersion` unchanged (its non-zero ObservedRouteVersion short-circuits the auto-pin on the existing condition). `DoesNotAutoPinForReadWriteTxn` unchanged (ReadKeys non-empty short-circuits first).

`go test -race ./kv` pass (full pkg, 10s) / `make lint` 0 issues.

@claude review

bootjp:
Pushed as d8487672 — codex P1 on 6202b964 addressed.

**Note on flaky CI:** `test (ubuntu-latest)` failed on 6202b964 in `TestHotKeysAggregatorRaceFree` (keyviz pkg, unrelated to this PR — I haven't touched keyviz). Reproduced locally 5/5 PASS. Likely a known flake. The other `test` job on 6202b964 passed.

**Codex P1 — drop M3 on secondary commits.** `commitSecondaryTxns` is best-effort by design (logs per-secondary errors and continues, then `dispatchMultiShardTxn` returns success while lock resolution backstops unavailable secondaries). Propagating ObservedRouteVersion to secondary commit Requests composes with M3's fail-closed semantic to yield a silent partial commit: a route shift between primary-COMMIT and secondary-COMMIT makes the secondary FSM return ErrComposed1Violation; the error is swallowed; the caller-visible Dispatch ACK still says success; the secondary's write is invisible to readers on the new owner.

**Coverage matrix** the post-fix code satisfies:
| Step | observedVer | Why |
|---|---|---|
| Prewrite (both shards) | pinned | gate-active before lock placement; clean abort path |
| Primary commit | pinned | linearization point; abortPreparedTxn cleans up on failure |
| Secondary commit | **0** | best-effort step has no rollback; gate would silently mask |

**Fix:** drop the `observedRouteVersion` parameter from `commitSecondaryTxns` entirely; per-secondary Requests are built with `ObservedRouteVersion` unset (zero). FSM short-circuits verifyComposed1 on observedVer==0 (fsm.go:609).

**Caller audit** (loop directive): change is entirely internal to commitSecondaryTxns. No external callers affected. Auto-pinned write-only multi-shard txns (DynamoDB BatchWriteItem etc.) lose the silent-partial-commit hazard; explicit-caller-pin callers also lose secondary-side coverage (the fundamental hazard is structural to 2PC + fail-closed gate; M5+ work for a fatal-secondary 2PC variant or lock-resolution-bypass).

**Regression test:** `TestShardedCoordinator_DropsObservedRouteVersionOn2PCSecondaryCommit` — 2-shard fixture, asserts the coverage matrix above. Failing on parent (g2 commit carried v=7) / passing on fix.

**Existing test** `TestShardedCoordinatorDispatchTxn_CrossShardPropagatesObservedRouteVersion` (PR #881 contract) updated to encode the refined contract — prewrite + primary commit carry pin, secondary commit drops it. The original "propagate everywhere" assertion was exactly what codex P1 identifies as the bug.

`go test -race ./kv` pass (full pkg, 8s) / `make lint` 0 issues.

@claude review

bootjp:
Pushed as 6a458a28 — codex P1 on d8487672 addressed.

**The d8487672 fix was the wrong interpretation.** Codex P1 on 6202b964 flagged "fail-closed gate + best-effort swallow = silent partial commit (missing-write direction)" and recommended "make secondary errors fatal/recoverable." I picked the wrong option ("disable the gate") instead of the explicit recommendation ("make fatal"). Codex P1 on d8487672 correctly showed that the disabled-gate posture trades one silent partial commit for another (stale-owner direction): the secondary FSM short-circuits verifyComposed1 and applies the write on a group that's no longer the catalog's owner.

**Correct fix — make secondary Composed-1 fatal with a distinct sentinel.**

1. Restored `ObservedRouteVersion` propagation to secondary commits.
2. New `ErrTxnSecondaryRouteShiftedAfterPrimaryCommit` sentinel — distinct from `ErrComposed1Violation` so the M4 retry path doesn't loop (the primary is durable; re-prewriting would conflict).
3. `commitSecondaryTxns` now returns `(uint64, error)`. Composed-1 from a per-secondary commit surfaces as the new sentinel; non-Composed-1 errors stay best-effort (logged+continued) preserving lock-resolution-backstop semantic.
4. `dispatchMultiShardTxn` propagates the sentinel to the caller.

**Coverage matrix** (now restored to PR #881's original):
| Step | ObservedRouteVersion | Composed-1 outcome |
|---|---|---|
| Prewrite (both shards) | pinned | abort cleanly |
| Primary commit | pinned | abortPreparedTxn cleans up |
| Secondary commit | **pinned** | surface ErrTxnSecondaryRouteShiftedAfterPrimaryCommit (fatal, no retry) |

**Caller audit** (loop directive): `commitSecondaryTxns` has only one caller (`dispatchMultiShardTxn`). The new sentinel is NOT in `isComposed1RetryableError`'s set, so M4 retry doesn't match. Adapters propagate the error through their HTTP/RPC error path — semantic change is **strict correctness improvement**: pre-fix silent data loss → post-fix visible error.

**Tests:**
- NEW `TestShardedCoordinator_SurfacesFatalErrorOn2PCSecondaryComposed1` — 2-shard fixture, secondary persistently fails Composed-1 on commit; asserts Dispatch returns the new sentinel and NOT `ErrComposed1Violation`. Failing on parent / passing on fix.
- INVERTED `TestShardedCoordinator_PropagatesObservedRouteVersionOn2PCSecondaryCommit` (was `Drops…` in d8487672) — secondary commit's observedVer=7 (restored, not 0).
- REVERTED `TestShardedCoordinatorDispatchTxn_CrossShardPropagatesObservedRouteVersion` to original PR #881 contract — every 2PC envelope on every shard carries pinned version.

`go test -race ./kv` pass (full pkg, 9s) / `make lint` 0 issues.

**Note on d8487672 CI failure:** `TestRedis_StreamXReadLatencyIsConstant` (adapter/redis, 48.35s) — latency-sensitive perf test on loaded ubuntu-latest runner, unrelated to this PR.

@claude review

bootjp:
Pushed as e08d44fb — codex P1 on 6a458a28 addressed.

**Codex P1 — skip auto-pin for resolver-claimed keys.** When a PartitionResolver is wired (SQS HT-FIFO partition keys etc.), ShardRouter routes via the resolver but verifyComposed1 calls route-history OwnerOf against the BYTE-RANGE engine snapshot. The engine has no knowledge of the resolver's mapping, so the gate spuriously rejects resolver-routed commits even when the resolver picked the correct gid. M4 retry re-reads engine.Version() (unchanged) and fails the same way — spurious double failure.

**Fix:** extend `maybeAutoPinObservedRouteVersion`'s skip conditions with a fifth case: "any element's key is recognised by the partition resolver." Extracted `anyResolverClaimedKey` helper that returns false fast when no resolver is wired (the byte-range-only deployment), iterates Elems otherwise, and calls `resolver.RecognisesPartitionedKey` per non-empty key.

**Skip-condition summary:**
| # | Skip when | Reason |
|---|---|---|
| 1 | engine == nil | no catalog to pin against |
| 2 | ObservedRouteVersion != 0 | caller already pinned |
| 3 | len(ReadKeys) != 0 | caller has pre-shift reads |
| 4 | callerSuppliedStartTS | caller-owned snapshot via readTS |
| 5 | **resolver-claimed key** | byte-range engine doesn't reflect resolver routing |

**Caller audit** (loop directive): change is internal to `maybeAutoPinObservedRouteVersion`. Only caller is `dispatchTxnWithComposed1Retry` (unchanged). Production resolver consumers (adapter/sqs_messages.go, adapter/sqs_messages_batch.go, etc.) now correctly commit instead of being spuriously rejected — strict correctness improvement. Byte-range-only callers unaffected.

**Regression test:** `TestShardedCoordinator_DoesNotAutoPinObservedRouteVersionForResolverRoutedTxn` — installs stubPartitionResolver claiming prefix "part:", dispatches a write-only txn with key "part:q1:1234", asserts observedVer=0 (auto-pin must NOT fire). Failing on parent / passing on fix.

`go test -race ./kv` pass (full pkg, 10s) / `make lint` 0 issues.

@claude review

PR #905, opened 2026-06-02
docs(design): Composed-1 M5 — Jepsen route-shuffle workload proposal

bootjp:
## Summary

- Design doc for **Composed-1 M5** — Jepsen route-shuffle workload, the integration-level proof from the parent doc's milestone table.
- **Doc only** — no implementation in this PR. Per CLAUDE.md design-doc-first workflow.
- M5 itself is two phases: M5a (CLI + workload setup, mergeable on its own), M5b (route-shuffle nemesis + cadence tuning).

## Companion to

- Parent: [`docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md`](../blob/main/docs/design/2026_05_29_proposed_composed1_cross_group_commit_guard.md)
- M1+M2+M3+M4 implementation: #900 (currently awaiting human merge after 9 codex review rounds)

## Forward-looking posture (matches parent)

Today's `SplitRange` is same-group only (per CLAUDE.md and `adapter/distribution_server.go`), so the Composed-1 hazard the M3/M4 guard catches cannot yet be *triggered* in production. M5 ships the scaffolding so:

1. The current `SplitRange` is exercised under realistic concurrent multi-shard write load and proved non-regressing — workload finds zero G1c, which is the baseline M4 contract.
2. When a future PR introduces a route-mutating RPC that DOES shift ownership across groups (cross-group `SplitRange`, `MoveRange`, online rebalancer), the M5 workload — with a one-line nemesis swap — becomes the integration-level proof that M3+M4 hold under cross-group churn.

## Open questions for review

The doc tracks four open questions in §5. The two that benefit most from reviewer input:

- **OQ-2** — Ship M5a + M5b in a single PR or two? (Two if M5a's CLI work runs ≥150 lines; one if M5a fits in a single screen.)
- **OQ-4** — Lifecycle of the parent doc post-#900 merge: rename `*_proposed_*.md` → `*_partial_*.md` (since only M5 remains)? Tentative yes.

## Test plan

- [x] `gofmt` / lint clean (doc-only, no Go changes)
- [ ] Reviewer: confirm M5a + M5b phasing matches existing repo conventions
- [ ] Reviewer: confirm `cmd/elastickv-split` as a standalone binary is preferable to a `cmd/elastickv-admin` subcommand (the doc tentatively chooses standalone; see §3.1 "Alternative considered")

## Summary by CodeRabbit

* **Documentation**
  * Added design doc for the Composed-1 M5 milestone introducing a DynamoDB route-shuffle testing workload.
  * Describes multi-shard/multi-group transaction coverage, setup verification, pass/fail validation criteria (including rejection of specific anomalies), default shuffle cadence, milestone breakdown, and remaining open questions.


bootjp:
Pushed as f5d2ad7a — all four findings on ffb9c73f addressed.

**Codex P1 — correctly identified the core flaw.** Verified against `kv/shard_key.go:94-124`: every DynamoDB table-metadata, item, and GSI key normalises to a SINGLE per-table route key (`!ddb|route|table|<tableSegment>`). Splitting inside a single-table workload's item keyspace cannot put two items on different shards, so the 2PC path (`dispatchMultiShardTxn`, secondary commits, the new `ErrTxnSecondaryRouteShiftedAfterPrimaryCommit` sentinel) would never fire — invalidating G2.

**Fix:** §3.3 rewritten. New workload variant `dynamodb-append-multi-table-workload` creates N=4 tables (`jepsen_append_t1` … `jepsen_append_t4`) and writes to ≥2 distinct tables per `TransactWriteItems`. The router maps each table to its own route key, so cross-table txns naturally fan out across shards. The setup hook splits at `!ddb|route|table|jepsen_append_t2` so tables 1 lives on one shard, tables 2-4 on another from t=0.

**Gemini medium R1 (Lexicographical split)** — fixed by anchoring split keys to the table-route prefix `!ddb|route|table|...` so they land INSIDE the active workload route range. The prior `/split/<int>` was lexicographically smaller than the workload's keyspace (`/` < `0`), so every workload key ended up on the rightmost shard.

**Gemini medium R2 (Route ID resolution)** — fixed by having the nemesis re-query `ListRoutes` on every `:start` invocation, walk the snapshot to find the route covering the chosen split key, and use that route's ID + `snapshot.version` as `expected_catalog_version`. Catalog drift surfaces as `ErrCatalogVersionMismatch` and the nemesis refreshes on the next tick. A cached route ID was indeed always going to be stale — every successful split deletes the parent and creates two fresh children.

**Gemini medium R3 (Setup hook gating)** — fixed by gating on `(when (= node (first (:nodes test))) ...)` so only the first node attempts the initial split. Jepsen's `db/setup!` runs on every node by default.

**Also:**
- §4 milestone table updated — M5a now ships the new workload variant (not just a setup hook), so it's meaningfully bigger than the original row suggested.
- Added OQ-5 (is N=4 the right default?) and OQ-6 (first-node gate semantics) as implementation-time follow-ups.
- Resolved OQ-4: PR #900 has merged, so the parent doc rename `*_proposed_*.md` → `*_partial_*.md` should now land as a separate small doc-only PR.

@claude review

bootjp:
Pushed as 24812867 — all six findings on f5d2ad7a addressed (2 codex P1, 2 claude P2, 2 claude P3).

**codex P1 #1 — encode table segments before split keys.** My split key used the raw table name; actual route key uses `dynamoRouteTableKey(encodeDynamoSegment(tableName))` — base64 RawURLEncoding (`adapter/dynamodb.go:8337-8365`). Split at raw name sorts after all encoded segments → all tables on one side. **Fix:** §3.2 now specifies a Clojure `encode-dynamo-segment` helper mirroring the Go encoding.

**codex P1 #2 — same-group SplitRange can't seed cross-group routes.** Current `SplitRange` only creates children with the parent's `GroupID`. `dispatchTxn` enters `dispatchMultiShardTxn` only when mutations group to ≥2 Raft **groups**, not ≥2 routes. Setup-hook SplitRange would leave tables in the same group → single-shard path → 2PC never fires. **Fix:** §3.3 substantially rewritten. **M5a now requires multi-group cluster startup** — extend `scripts/run-jepsen-local.sh` to pass `--shardRanges` so cluster launches with ≥2 Raft groups, table-route keys for t1..t4 statically split across groups. Jepsen `db/setup!` no longer issues SplitRange — calls `ListRoutes` (gated to first node) and **verifies** the multi-group routing is in place, failing fast if not. Added a "Why the nemesis is still useful" paragraph: same-group SplitRange churns catalog versions/route IDs, exercising M3+M4 drift detection — a non-regression check, not a real anomaly hunt until cross-group routing lands.

**claude[bot] P2 #1 — success criterion would always pass.** My `(filter #(= (:type %) :G1c) (:anomalies results))` was wrong: `:anomalies` is a map (keyword → seq), iterating yields `[k v]` vectors with no `:type`. **Fix:** §3.4 uses `(nil? (get (:anomalies results) :G1c))`.

**claude[bot] P2 #2 — lexicographic direction inverted.** `!` is ASCII 33, `/` is 47 → `/split/` is LARGER than `!ddb|...`. Workload keys would land on LEFT shard, not rightmost. **Fix:** §3.2 explanation corrected.

**claude[bot] P3 #1 — NG3 contradicts §3.3.** Tightened: "no new operation types or generators **beyond** the multi-table variant and the route-shuffle nemesis."

**claude[bot] P3 #2 — setup hook route ID resolution.** Resolved differently: §3.3's rewrite means the hook no longer mutates routes (only verifies), so no route-ID lookup needed for mutation. Just ListRoutes for verification.

**Added OQ-7** for the `--shardRanges` boundary key encoding helper (tentative: small Go binary `cmd/elastickv-route-key` so the encoding stays in Go).

@claude review

bootjp:
Pushed as 673bee8c — coderabbitai minor lint findings on 24812867 addressed:

- **MD027** (multiple-space-after-blockquote) at lines 17, 21 — trimmed double-spaces after `>` to a single space in the upfront posture block.
- **MD040** (fenced-code-language) at three plain ``` blocks — added language tags: `bash` for the `elastickv-split` CLI usage, `clojure` for both the route-shuffle nemesis sketch and the success-criterion snippet.

Doc-only, no semantic changes.

@claude review

bootjp:
Pushed as 3ca2a7f7 — claude[bot] P2 + P3 on 24812867 addressed.

**P2 — launch-script scope underestimated.** Confirmed against `shard_config.go:61-99`: without `--raftGroups`, every `--shardRanges` entry collapses into the default single group 1 and `groupMutations` returns `gids = [1]` for every key. Single-shard path → 2PC never fires. My "only `--shardRanges` is new" was wrong.

**Fix:** §3.3 table cell rewritten to list **five** coordinated launch-script changes M5a must make:
1. `--raftGroups` declaring ≥2 groups
2. `--shardRanges` placing t1-t2 in group 1, t3-t4 in group 2
3. Cluster topology decision (tentative: 2 single-node groups for CI speed; scale to 6 nodes if needed)
4. Per-group `--raftBootstrapMembers`
5. Expanded `--raftDynamoMap` + matching port allocations

This documents the topology gap upfront so the M5a implementer doesn't discover it mid-PR.

**P3 — Clojure Base64 nuance.** Go uses `base64.RawURLEncoding` (URL-safe charset **without** `=` padding). Java's three Base64 variants differ:
- `Base64/getEncoder` (standard `+`/`/`, with padding) — **wrong**
- `Base64/getUrlEncoder` (URL-safe, **with** padding) — **wrong**
- `Base64/getUrlEncoder` + `.withoutPadding` — **correct**

Failure mode is silent: wrong encoding produces split keys outside every route's range; ListRoutes returns no match; nemesis logs and skips every `:start`; the run appears clean while the nemesis is a no-op.

**Fix:** §3.2 now calls out `.withoutPadding` explicitly and documents the silent-no-op failure mode.

@claude review

bootjp:
Pushed as 87dcfe3f — claude[bot]'s only remaining minor on 673bee8c addressed.

**claude[bot] LGTM ("Ready to merge") on 673bee8c** with one minor doc-clarity note: the bottom-of-§4 line *"M5a is a small, focused PR (Go CLI + Clojure setup hook + docs)"* was accurate for the original M5a scope but became understated after the codex P1 #2 revision expanded M5a to include the multi-table workload variant and the five-change multi-group launch script.

**Fix:** §4 trailer rewritten to reflect actual M5a scope:
- `cmd/elastickv-split` (Go CLI)
- `dynamodb-append-multi-table-workload` (Clojure)
- `scripts/run-jepsen-local.sh` multi-group launch extension (5 coordinated changes per §3.3 table)
- Setup-hook verification (read-only ListRoutes + assert)
- Docs

Likely lands as a single PR for atomicity (workload variant only makes sense alongside multi-group launch + verification) but reviewable as four roughly independent diff sections.

claude[bot]'s review of 3ca2a7f7 (which carried the launch-script-scope + Base64-nuance fixes themselves) is still in flight at the time of this comment — I don't expect new findings there since the fixes followed claude[bot]'s own P2/P3 recommendations.

@claude review

bootjp:
Pushed as b752a894 — codex P2 + claude[bot] P2 (same defect) and codex P2 #2 on 3ca2a7f7 addressed.

**codex P2 + claude[bot] P2 (same defect, independently flagged) — `--raftBootstrapMembers` rejected on multi-group startup.** Confirmed against `main.go:735-741`: `resolveBootstrapServers` returns `ErrBootstrapMembersRequireSingleGroup` whenever `--raftBootstrapMembers` is set and `--raftGroups` parses to more than one group. My §3.3 item 4 ("Per-group `--raftBootstrapMembers`") would abort startup before Jepsen could run.

**Fix:** §3.3 item 4 now correctly specifies `--raftBootstrap` (boolean) for the tentative 2-single-node-groups topology. Single-member groups need no peer discovery → `--raftBootstrap` is necessary and sufficient. The existing `--raftBootstrapMembers` in `run-jepsen-local.sh` must be **removed** for multi-group processes — only valid for single-group multi-node formations.

**codex P2 #2 — split key reuse becomes no-op.** My §3.2 example showed `!ddb|route|table|<encode("jepsen_append_t2")>` as the split key, but after the setup-time split that key is the **`Start`** of the right child route. `adapter/distribution_server.go:357-372`'s `validateSplitKey` rejects `splitKey == parent.Start/End` with "split key at route boundary" — every subsequent nemesis tick would fail. In a 5-min run expecting ≥10 shuffles, the nemesis would no-op silently and the workload would appear clean while catalog churn never lands.

**Fix:** §3.2 split into TWO distinct split-key constructions:
- **(a) Setup-time boundary key** (one-shot, in `run-jepsen-local.sh`) — IS a route boundary by design (the static partition between groups).
- **(b) Per-shuffle interior key** (every nemesis `:start`) — MUST lie strictly between the current route's `Start` and `End`. Strategy: append a base64-alphabet byte plus a fresh counter to `route.start`; the result sorts after start but before any non-base64 byte (e.g. ASCII `|`=124).

Implementation details (counter persistence, collision recovery) deferred to M5b; the design contract is: each shuffle's split key is strictly interior to the route's current range.

claude[bot] gave LGTM on **87dcfe3f** "Ready to merge" with no new findings — this commit (b752a894) addresses two findings from claude[bot]'s **earlier** 3ca2a7f7 review (which arrived between the 87dcfe3f review starting and finishing) plus codex's two P2s on 87dcfe3f.

@claude review

bootjp:
Pushed as 6f024aaa — addresses the only remaining note from claude[bot]'s "Ready to merge" review of b752a894.

**claude[bot] minor (non-blocking) on b752a894**: `ListRoutes` protobuf response fields may be Java byte arrays rather than strings depending on deserialization. Calling Clojure `(str byte-array)` silently yields `"[B@…"` (Java object reference) instead of the bytes — a silent mis-routing footgun.

**Fix:** §3.2's existing "Implementation details deferred to M5b" list now explicitly includes "byte-vs-string coercion of `ListRoutes` response fields" alongside "counter persistence" and "collision recovery." The footgun is now visible at design-doc-read time so the M5b implementer can budget for it rather than discover it through debugging silent mis-routing.

**State of PR #905:**
- claude[bot] LGTM "Ready to merge" on both **87dcfe3f** and **b752a894** (back-to-back review cycles)
- Codex bot: rate-limited (will not review further until reset)
- CI on b752a894 (and 6f024aaa once it kicks): build/test/lint all pass
- 9 commits of refinement across 5 substantive review rounds (codex + claude + gemini + coderabbit)

Loop exit condition reached. Stopping the 5-minute cron.

@claude review

bootjp:
Pushed as f92a029e — codex P1 on 6f024aaa addressed.

**codex P1 — proposed topology not supported by the current server.** Confirmed against the codebase:
- `shard_config.go:399-410` (`validateShardRanges`) requires every `--shardRanges` group ID to appear in that process's `--raftGroups`.
- `main.go:764` (`buildShardGroups`) starts a local Raft runtime for every group listed in `--raftGroups`.

Consequence: my proposed "2 processes, each hosting one group" is **rejected** — Process A with `--raftGroups "1=…"` but a `--shardRanges` entry pointing to group 2 fails validation. Configuring both processes with `--raftGroups "1=…,2=…"` makes both try to host both groups and race on the Raft listeners.

**Fix:** §3.3 items 3-5 rewritten for the SUPPORTED topology — **one process hosting both single-member groups**, two Raft addresses (`50051` for group 1, `50054` for group 2), shared Dynamo endpoint (e.g. `63801`). Each group is single-member so `--raftBootstrap` works for both. Cross-group txns go through the in-process router → 2PC fires across the two co-located groups.

**Trade-off explicitly documented:** because both groups live in one process, partition/kill nemeses can't separate them — only the route-shuffle nemesis (API-level) exercises the cross-group path meaningfully. True distributed multi-group (multi-process) requires server-side support for "remote-only groups in `--raftGroups`" or equivalent — **M6+ work, explicitly out of M5's scope**.

This is a strictly correct narrowing of the topology choice (removed the unsupported "2 separate processes" and the speculative "6 nodes" expansion; the only supported shape is now spelled out). The M5a Done-when criterion still holds: cluster starts with tables 1-2 in group 1 and tables 3-4 in group 2, workload exercises `dispatchMultiShardTxn`.

@claude review

bootjp:
Pushed as 58d6ac5e — coderabbit Major on f92a029e addressed.

**coderabbit Major — §4 row was inconsistent with §3.3's multi-group contract.** The M5a milestone row said the launch extension passes `--shardRanges` so the cluster launches with ≥2 Raft groups, but §3.3 explicitly states (per the claude[bot] P2 on 24812867 that drove that rewrite) that `--shardRanges` alone collapses every range to the single default group 1 — both `--raftGroups` AND `--shardRanges` are required.

**Fix:** §4's M5a Scope cell now explicitly says "pass BOTH `--raftGroups` (declaring ≥2 groups) AND `--shardRanges` (placing tables across those groups)" and cross-references §3.3 for the "shardRanges alone collapses to group 1" explanation. An implementer reading §4 in isolation will no longer miss the `--raftGroups` requirement.

Doc-only, no semantic changes — just internal consistency between the §4 row and the §3.3 table.

@claude review

PR #907, opened 2026-06-02
feat(encryption): Stage 6E-1b — EnableRaftEnvelope admin RPC + elastickv-admin CLI

bootjp:
## Summary

Stage 6E-1b — operator-facing trigger for the Phase-2 raft cutover, mirroring the 6D-6 EnableStorageEnvelope path. The FSM apply machinery shipped in #899 (Stage 6E-1a); this PR adds the proto RPC, server method, and `elastickv-admin enable-raft-envelope` subcommand so operators can actually invoke the cutover.

Operator-inert by design: with only 6E-1 (a + b) deployed, the cutover entry advances the sidecar's `RaftEnvelopeCutoverIndex` but no Raft proposal is wrapped and no apply is unwrapped — those land atomically in 6E-2.

## What changed

- `proto/encryption_admin.proto` — new `EnableRaftEnvelope` RPC + `EnableRaftEnvelopeRequest` / `EnableRaftEnvelopeResponse` messages. The response omits the storage variant's `cutover_index_unknown` defensive field because `RaftEnvelopeCutoverIndex != 0` IS the active sentinel — there's no separate bool to defend against, and 6E-1a already fail-closes on `raftIdx == 0` before any sidecar mutation.
- `adapter/encryption_admin.go` — `EnableRaftEnvelope` server method. Same semaphore / pre-check / fan-out / propose / post-check sequence as `EnableStorageEnvelope`. Variant-specific differences: target `Purpose=PurposeRaft`, source slot `Active.Raft`, active-sentinel field `RaftEnvelopeCutoverIndex`.
- `cmd/elastickv-admin/` — `enable-raft-envelope` subcommand (parser/runner/printer). Output prefixes (`enabled` / `already-active`) match the storage subcommand for script compatibility. No warning line — no defensive `cutover_index_unknown` field.
- `docs/design/2026_05_31_proposed_6e_enable_raft_envelope.md` → `partial`: milestone 6E-1 (both sub-milestones) is now complete. 6E-2 and 6E-3 remain.

## Protoc version skew (documented)

The pinned protoc toolchain (`libprotoc 29.3`) is not available in this dev environment; locally installed is `libprotoc 34.0`. Empirical comparison: a no-op regen against the unmodified proto file with both versions produces byte-identical output **except** the `// protoc vX.Y.Z` version comment in the file header. The version comment in this PR has been reverted to the pinned 29.3 output, so the diff is purely the new additive content. CI/release should re-run `make gen` with the pinned 29.3 toolchain to confirm the generated artifacts.

## Test plan

- [x] `go test ./adapter/ ./cmd/elastickv-admin/ ./proto/ ./internal/encryption/` (without `-race`): all pass.
- [x] Focused `-run 'EnableStorageEnvelope|EnableRaftEnvelope'` race-tests in adapter: pass (0.571s).
- [x] `go test ./kv/...` (FSM apply chain): pass.
- [x] `golangci-lint --config=.golangci.yaml run` on all impacted packages: 0 issues.
- [ ] Note: full `go test -race ./adapter/...` times out at 10 min on this hardware due to existing slow integration tests in adapter; not specific to this PR. The new tests themselves are quick unit tests.

1. **Data loss** — no replication / FSM apply path changes; new server method only proposes through the existing `proposeEncryptionEntry` seam (same as `EnableStorageEnvelope`).
2. **Concurrency** — `acquireCutoverSemaphore` reused verbatim from the storage variant; ctx-aware wait, pre-canceled short-circuit, and serialization properties inherited (those branches are tested by the storage suite, which exercises the same code paths).
3. **Performance** — one extra entry in proto type registry; no hot-path impact.
4. **Data consistency** — proposal shape (`Purpose=PurposeRaft`, `DEKID=Active.Raft`, `len(Wrapped)==0`) re-validated at apply by 6E-1a's `validateEnableRaftEnvelopePayload`. Server and applier agree on the constraints.
5. **Test coverage** — 5 new server tests (zero proposer ID, oversized epoch, not-bootstrapped, idempotent retry, happy path) + 5 new CLI tests + 1 dispatch test. The cancellation / serialization / fan-out branches reuse code paths already covered by the storage variant's test suite.

## Caller audit (semantic-change directive)

No semantic changes to existing functions — purely new additions. The new `EnableRaftEnvelope` server method has a single transport-layer caller (gRPC dispatch), and the new CLI subcommand has a single dispatch entry in `encryptionSubcommands`. Both are exercised by the new tests.

## Release Notes

* **New Features**
  * Added `enable-raft-envelope` CLI command enabling Raft envelope encryption cutover.
  * New RPC endpoint supports idempotent cutover operations with per-member capability tracking.

* **Documentation**
  * Updated implementation status documentation reflecting Phase-1 completion for Raft envelope encryption feature.

### DIFF
--- a/adapter/encryption_admin.go
+++ b/adapter/encryption_admin.go
@@ -1032,6 +1032,21 @@ func freshCutoverResponse(sc *encryption.Sidecar, proposedIdx uint64, fanoutResu
 	}
 }
 
+// raftEnvelopeWrapEnabled gates the EnableRaftEnvelope RPC until
+// the Stage 6E-2 wrap-on-propose / unwrap-on-apply / §7.1
+// proposal-quiescence-barrier triple ships. With only 6E-1
+// (admin RPC + FSM-apply machinery) deployed, accepting a cutover
+// proposal records RaftEnvelopeCutoverIndex=N in the sidecar, but
+// no entry between N and the eventual 6E-2 upgrade is wrapped —
+// they all remain cleartext. The future engine apply-hook
+// (designed in §6.3) dispatches `entry.Index > cutover` through
+// the unwrap path, so a 6E-2 upgrade against a sidecar where N
+// was recorded under a 6E-1-only build would treat every
+// pre-upgrade cleartext entry above N as an envelope and halt
+// apply cluster-wide. Gate fails closed until 6E-2 atomically
+// flips this to true alongside the wrap/unwrap/barrier wiring.
+const raftEnvelopeWrapEnabled = false
+
 // EnableRaftEnvelope is the Stage 6E Phase 2 cutover — flips Raft
 // proposals from cleartext to §4.2-envelope. Structural mirror of
 // EnableStorageEnvelope; the differences are:
@@ -1049,6 +1064,13 @@ func freshCutoverResponse(sc *encryption.Sidecar, proposedIdx uint64, fanoutResu
 // The semaphore, pre-check / fan-out / propose / post-check
 // sequence, and error mapping match the storage variant verbatim;
 // see EnableStorageEnvelope for the full design rationale.
+//
+// **Gated**: refuses with FailedPrecondition until 6E-2 ships the
+// wrap-on-propose / unwrap-on-apply / §7.1 barrier (see
+// raftEnvelopeWrapEnabled for the rationale). The pre-gate
+// validation surface (leader, semaphore acquire, request shape)
+// still fires so operators get fast feedback on wiring problems,
+// but no Raft proposal is composed and no sidecar mutation occurs.
 func (s *EncryptionAdminServer) EnableRaftEnvelope(ctx context.Context, req *pb.EnableRaftEnvelopeRequest) (*pb.EnableRaftEnvelopeResponse, error) {
 	if err := s.acquireCutoverSemaphore(ctx); err != nil {
 		return nil, err
@@ -1060,6 +1082,15 @@ func (s *EncryptionAdminServer) EnableRaftEnvelope(ctx context.Context, req *pb.
 	}
 	if earlyResp != nil {
 		return earlyResp, nil
+	}
+	if !raftEnvelopeWrapEnabled {
+		// Without 6E-2 wrap-on-propose, recording the cutover
+		// index here would let cleartext entries land at indexes
+		// > N. A 6E-2 upgrade would then treat those cleartext
+		// entries as envelopes and halt apply. Refuse before the
+		// fan-out and propose so no sidecar state changes.
+		return nil, grpcStatusError(codes.FailedPrecondition,
+			"encryption: enable-raft-envelope is gated until Stage 6E-2 ships wrap-on-propose / unwrap-on-apply / §7.1 proposal-quiescence-barrier; accepting the cutover now would brick the cluster on the next upgrade")
 	}
 	fanoutResult, err := s.runCutoverFanout(ctx)
 	if err != nil {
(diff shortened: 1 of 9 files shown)


bootjp:
@claude review

Round-1 fixes in 1facaa14:

- **codex P1** (cluster-bricking hazard) — Added a fail-closed gate (package-level `const raftEnvelopeWrapEnabled = false` in `adapter/encryption_admin.go`). `EnableRaftEnvelope` refuses fresh cutover proposals with `FailedPrecondition` until 6E-2 ships the wrap-on-propose / unwrap-on-apply / §7.1 proposal-quiescence-barrier triple. The pre-check (leader gate, bootstrap gate, idempotent-retry short-circuit) still fires for fast operator feedback; only the propose path is refused. 6E-2 flips the const atomically with the wrap/unwrap/barrier wiring so the open-gate-without-wrap window cannot exist. Replaced the `HappyPath` test with `GatedUntil6E2`, which pins (1) `FailedPrecondition` with a 6E-2 hint in the message, (2) `proposer.calls` empty, (3) sidecar `RaftEnvelopeCutoverIndex` remains 0 — i.e., no state mutation can occur before 6E-2 lifts the gate, so a future upgrade is safe.

- **gemini MEDIUM** — Removed the dead `if appliedIndex == 0` defensive branch in `freshRaftCutoverResponse`. The raft variant uses the cutover index itself as the active sentinel, so `raftCutoverPostcheck` refuses earlier on a zero index; reaching `freshRaftCutoverResponse` guarantees non-zero. The storage variant's hand-edit hedge has no analogue here (the storage variant's `appliedIndex == 0` branch covers `StorageEnvelopeActive=true ∧ StorageEnvelopeCutoverIndex=0`, a two-field-divergence state that cannot exist when active and index are the same field).

- **Caller audit** (per the project's semantic-change-requires-grep directive): `EnableRaftEnvelope`'s sole production caller is the auto-generated gRPC dispatcher in `proto/encryption_admin_grpc.pb.go`; the CLI wraps any error with `errors.Wrap(err, "EnableRaftEnvelope")`, which surfaces the gate message cleanly. `freshRaftCutoverResponse`'s sole caller is `raftCutoverPostcheck`, which I already audited as guaranteeing non-zero. No other callers.

- **Design doc** — added §3.3 documenting the gate, the codex P1 rationale (the engine apply-hook's `entry.Index > cutover` semantics and why cleartext entries above N would brick a 6E-2 upgrade), and the 6E-2 atomic-flip contract.

The orphaned `assertRaftCutoverProposalShape` helper was dropped (its only caller was the old `HappyPath` test). When 6E-2 lands, a new HappyPath test reintroduces the proposal-shape assertion.

bootjp:
@claude review

Round-2 follow-up (7c4403ba + cafe9bf0):

- **claude round-1 cosmetic #1-3** — fixed in cmd/elastickv-admin/encryption.go: usage string and unknown-subcommand error now both list `enable-raft-envelope`; the stale block comment saying "enable-raft-envelope lands in Stage 6E" was replaced with a concise shipping-order list that notes 6E-1b is gated until 6E-2 ships wrap-on-propose.
- **coderabbit round-1 Major / Minor** — addressed in cafe9bf0: §0 prose clarified to defer engine unwrap + coord wrap + barrier to 6E-2 (planned, not shipped); "this PR" placeholder replaced with `#907` reference.

All claude round-1 findings now closed:
- Substantive items (gate, dead branch, tests, §3.3) — confirmed LGTM in the previous review.
- Cosmetic items #1 (usage), #2 (error list), #3 (block comment), #4 (doc placeholder) — all addressed.

Tests pass on cmd/elastickv-admin (0.340s); lint clean.

PR #911, opened 2026-06-03
feat(cmd): M5a — elastickv-split + elastickv-route-key CLI tools

bootjp:
**Draft — first slice of Composed-1 M5a.** Based on the M5 design doc at `docs/design/2026_06_02_proposed_composed1_m5_jepsen_route_shuffle.md` (merged via #905). I'll iterate the remaining M5a pieces as separate commits on this branch and unmark draft once the workload passes locally.

### 1. `cmd/elastickv-split` (design doc §3.1)

Single-shot CLI that invokes `Distribution.SplitRange` against a running cluster. Flags: `--address`, `--route-id`, `--split-key`, `--expected-version`. Prints `catalog_version` + `left`/`right` child route descriptors; non-zero exit on any error so the Jepsen nemesis sees failures verbatim.

Smoke test covers the four required-flag rejection cases + the all-present accept case. The real coverage is the Jepsen run itself per the design doc.

### 2. `cmd/elastickv-route-key` (design doc OQ-7)

Prints the table-route key (`!ddb|route|table|<base64(name)>`) for a given DynamoDB table name. Used by the launch script (still TODO) to compute `--shardRanges` boundary keys without inlining the base64 encoding in shell — codex P1 #1 on #905 flagged that getting the encoding wrong silently mis-routes.

Test pins the output against the adapter's actual encoding (`adapter/dynamodb.go:8441` `encodeDynamoSegment` → `base64.RawURLEncoding`) for all four M5 tables plus a single-char case.

## Remaining M5a work (separate commits on this branch)

- [ ] `jepsen/src/elastickv/dynamodb_multi_table_workload.clj` — new workload variant; N=4 tables; each `TransactWriteItems` writes to ≥2 tables
- [ ] `scripts/run-jepsen-local.sh` (or new `scripts/run-jepsen-m5-local.sh`) — single-process two-group launch per design doc §3.3 (must pass BOTH `--raftGroups` AND `--shardRanges`)
- [ ] Setup-hook verification — `ListRoutes` gRPC client in Clojure, asserts multi-group routing; gated to first node via `(when (= node (first (:nodes test))) …)`
- [ ] Done-when check: workload exercises `dispatchMultiShardTxn` (via server-side log marker or probe metric) and Elle finds zero G1c

## Test plan

- [x] `go test -race ./cmd/elastickv-split ./cmd/elastickv-route-key`
- [x] `golangci-lint run` (0 issues)
- [ ] `./scripts/run-jepsen-m5-local.sh` (once that script + workload land)
- [ ] Elle finds zero G1c on the multi-table workload

### DIFF
--- a/cmd/elastickv-route-key/main.go
+++ b/cmd/elastickv-route-key/main.go
@@ -1,0 +1,58 @@
+// elastickv-route-key prints the byte-string of a DynamoDB
+// table-route key for a given table name.  Used by the Composed-1
+// M5 launch script (`scripts/run-jepsen-local.sh`) and OQ-7
+// (docs/design/2026_06_02_proposed_composed1_m5_jepsen_route_shuffle.md)
+// to compute `--shardRanges` boundary keys without inlining the
+// base64-encoding logic in shell — the encoding is part of the
+// routing surface and any drift would silently mis-route.
+//
+// Usage:
+//
+//	elastickv-route-key jepsen_append_t1
+//	# → !ddb|route|table|amVwc2VuX2FwcGVuZF90MQ
+//
+// The output matches `dynamoRouteTableKey(encodeDynamoSegment(name))`
+// (kv/shard_key.go:117 + adapter/dynamodb.go:8441) byte-for-byte.
+// A test pin (main_test.go) covers the contract against
+// representative names.
+//
+// The output is written WITHOUT a trailing newline (raw bytes) so
+// shell composition (e.g. `--shardRanges "$(elastickv-route-key
+// jepsen_append_t2)=2"`) works without `tr -d '\n'`.
+package main
+
+import (
+	"encoding/base64"
+	"fmt"
+	"os"
+)
+
+// dynamoRoutePrefix mirrors kv.dynamoRoutePrefix (unexported in
+// the kv package).  Keep this in sync — the M5 launch script's
+// --shardRanges boundary keys must match the routing layer's
+// expectations exactly.
+const (
+	dynamoRoutePrefix = "!ddb|route|table|"
+	// usageExitCode is the conventional shell exit code for
+	// "command-line usage error" (sysexits.h EX_USAGE = 64; many
+	// CLIs simplify to 2).  Distinct from exit 1 used for runtime
+	// failures so shell callers can branch.
+	usageExitCode = 2
+)
+
+func main() {
+	if len(os.Args) != 2 || os.Args[1] == "-h" || os.Args[1] == "--help" {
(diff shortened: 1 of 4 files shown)


bootjp:
Pushed as e7920003 — gemini medium on 0a102e02 addressed.

**Two valid issues** in `cmd/elastickv-split/main.go`:

1. **Speculative / unused code** — `dialCtx`, `dialCancel`, and `dialTimeout` were declared but never used. `grpc.NewClient` is non-blocking and does not take a context; the dial timeout was a YAGNI placeholder for a hypothetical future enhancement. Dropped.
2. **Silent `conn.Close()`** — `defer conn.Close()` ignored its error return, potentially masking resource leaks or half-closed-stream conditions.

**Fix:** dropped the unused dial-context plumbing; replaced `defer conn.Close()` with a `defer func()` that logs any close error to stderr. The close error does **not** fail the process — by that point the `SplitRange` response is already in hand, and a noisy close shouldn't mask a successful RPC. The stderr line is the visibility hook for resource-leak / half-close diagnosis.

@claude review

bootjp:
Pushed as f793d664 — two trivial nits from claude[bot]'s "Nothing blocking" review of e7920003 addressed inline:

- **nit 2** — `--expected-version` flag description now says "required, must be ≥ 1 — catalog version is 1-based" so `--help` matches the `validateFlags` rejection of 0.
- **nit 4** — `errors.Newf` with no format args → `errors.New`.

**Deferred** (claude[bot]'s informational notes, no action needed now):
- Finding 1 — `dynamoRoutePrefix` duplication has no compile-time guard against drift from `kv.dynamoRoutePrefix`. Acceptable given the test pin; future refactor when the `kv` package exports the constant cleanly.
- Finding 3 — `%q` on byte-slice route `Start`/`End` in `printResponse`. Fine for human inspection; flagged for the future Jepsen Clojure script author so they don't expect raw unquoted bytes.
- Finding 5 — `_ = flag.CommandLine` defensive guard. claude[bot] itself said "probably fine as-is."

@claude review

bootjp:
Pushed as 19decc6e — the only nit from claude[bot]'s LGTM review of f793d664 addressed.

claude[bot]'s 'Finding 1' on f793d664: stale `errors.Newf` reference in the `resetFlags` doc comment. Updated to match the actual `errors.New` call. Comment-only change.

No further inline review items outstanding on the CLI tools — three deferred-by-claude items (`dynamoRoutePrefix` duplication, `%q` on byte-slice, `_ = flag.CommandLine`) remain non-blocking and will be tracked at the next refactor opportunity.

@claude review

PR #912, opened 2026-06-03
backup: #35 - validate name/dir consistency in S3 + DDB reverse encoders

bootjp:
## Summary
- S3 (`encodeBucket`) and DDB (`encodeTable`) now check that the on-disk subdirectory matches `EncodeSegment([]byte(name))` from `_bucket.json` / `_schema.json` before emitting any records.
- Closes task #35 (the M4 + M5 follow-up the original session-summary tracked).

## Background

The reverse encoders use two identifiers for the same logical entity:
- On-disk subdir: `s3/<bucketDir>/`, `dynamodb/<tableDir>/` — decoder spells as `EncodeSegment([]byte(name))`.
- Embedded name in `_bucket.json` / `_schema.json` — encoder uses for MVCC keys (`s3keys.BucketMetaKey(pub.Name)`, base64'd `tableName`).

Until now, both sides were trusted independently. A tampered or hand-rebuilt dump that renamed the directory without touching the JSON (or vice versa) would silently emit records keyed by the JSON name while pulling object / item bytes from a mismatched filesystem path — a name/dir consistency violation that splits a single restore artifact across two identities.

## Fix

After the existing empty-name guard, both encoders verify `EncodeSegment([]byte(name)) == <dir>`. Mismatch fails closed with the existing per-adapter sentinel (`ErrS3EncodeInvalidBucket` / `ErrDDBEncodeInvalidSchema`), and the error message names both the JSON value and the encoded form so the operator can correct the dump.

## Caller audit (CLAUDE.md semantic-change rule)

- `encodeBucket` (S3): success path gains one new fail-closed check; sole production caller is `S3RecordEncoder.Encode`.
- `encodeTable` (DDB): success path gains one new fail-closed check; sole production caller is `DynamoDBEncoder.Encode`.
- Both checks reuse existing per-adapter sentinels, so `runAdapterEncoders`' `errors.Mark(ErrEncodeAdapterData)` wrap still routes failures through the CLI's exit-2 classification — no new sentinel needed in `classifyEncodeError`.

## Self-review of code changes (5 lenses)

1. **Data loss** — Stricter validation; no commit/apply path touched. No data-loss surface.
2. **Concurrency / distributed failures** — Encoder is single-process / single-goroutine on the input tree. No new concurrency surface.
3. **Performance** — One extra `EncodeSegment` call and string compare per bucket / per table. Negligible.
4. **Data consistency** — This IS the data-consistency improvement. The check enforces the round-trip invariant `<dir> == EncodeSegment(name)` for every published record.
5. **Test coverage** — Two new tests (one per adapter) pin the rejection. Existing positive tests use `EncodeSegment` for the dir spelling, so they remain green; existing rejection tests trip earlier guards, also green.

## Test plan
- [x] `go test -race ./internal/backup/... ./cmd/elastickv-snapshot-encode/...`
- [x] `golangci-lint --config=.golangci.yaml run`
- [ ] Bot review cycle (claude / codex / CodeRabbit)

### DIFF
--- a/internal/backup/encode_dynamodb.go
+++ b/internal/backup/encode_dynamodb.go
@@ -119,6 +119,19 @@ func (e *DynamoDBEncoder) encodeTable(b *snapshotBuilder, root *os.Root, tableDi
 	if tableName == "" {
 		return errors.Wrapf(ErrDDBEncodeInvalidSchema, "%s/_schema.json: empty table_name", tableDir)
 	}
+	// Name/dir consistency: the decoder spells the table subdir as
+	// EncodeSegment([]byte(name)) (dynamodb.go:213); a tampered or
+	// hand-rebuilt dump could rename one without the other and the
+	// encoder would otherwise emit DDBTableMeta records under
+	// tableName while reading items from a mismatched on-disk path.
+	// Fail closed so the round-trip invariant
+	// `<tableDir> == EncodeSegment(tableName)` holds for every table
+	// the encoder publishes (#35 follow-up).
+	if want := EncodeSegment([]byte(tableName)); want != tableDir {
+		return errors.Wrapf(ErrDDBEncodeInvalidSchema,
+			"%s/_schema.json: table_name %q encodes to dir %q, not %q (dump tampered or rebuilt by hand?)",
+			tableDir, tableName, want, tableDir)
+	}
 	// A table must have a hash key; fail early here rather than letting
 	// an empty primary key propagate into the item encoder (which
 	// builds item keys from the hash/range attributes).
--- a/internal/backup/encode_dynamodb_test.go
+++ b/internal/backup/encode_dynamodb_test.go
@@ -238,10 +238,15 @@ func TestDDBEncodeRejectsDuplicateGSIName(t *testing.T) {
 
 // TestDDBEncodeRejectsEmptyHashKey pins that a schema with no primary
 // hash key fails closed rather than propagating into the item encoder.
+// The dir intentionally matches EncodeSegment(table_name) so the new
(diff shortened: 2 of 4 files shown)


bootjp:
@claude review

This is the implementation PR for follow-up task #35 (name/dir consistency check). Adds a fail-closed validation in both `encodeBucket` (`encode_s3.go`) and `encodeTable` (`encode_dynamodb.go`) that verifies `EncodeSegment(pub.Name) == bucketDir` (and the DDB equivalent), so a hand-edited dump where the on-disk dir name doesn't match what the encoded name would produce fails closed via `ErrS3EncodeInvalidBucket` / `ErrDDBEncodeInvalidSchema`.

Tests pinned:
- `TestS3EncodeRejectsNameDirMismatch`
- `TestDDBEncodeRejectsNameDirMismatch`

Gemini posted "no feedback" earlier; no other bot reviews yet.

bootjp:
@claude review

Fixes the claude review finding on the prior revision:

**Finding 1 (TestDDBEncodeRejectsEmptyHashKey silent regression):** The test used `dir="tbl"` with `table_name="x"`. After the new name/dir consistency check at `encodeTable` L130 (`EncodeSegment("x")="x" != "tbl"`), the name/dir guard fired first with the same `ErrDDBEncodeInvalidSchema` sentinel, so the test silently exercised the wrong guard. Now uses `dir=EncodeSegment("x")` so the gate ordering reaches the empty-hash-key check at L138. Added a comment documenting the gate ordering so a future inserted guard can't silently break this test again.

**Finding 2 (cosmetic redundancy):** Left as-is per claude's note that it's cosmetic-only.

No production code change — encoder gates are unchanged; only the test fixture is realigned.

PR #915, opened 2026-06-03
feat(snapshot-skip B2): plumb metaAppliedIndex through raft-Apply + both snapshot persist sites

bootjp:
## Summary

Implements **Branch 2** of the cold-start snapshot-restore skip optimisation designed in PR #910. After this lands the `metaAppliedIndex` Pebble meta key is durably written on every raft-Apply data mutation AND at every snapshot persist — but the skip gate itself (Branch 3) is NOT yet wired, so behaviour is observationally identical to `main` except for the new meta key in fsm.db. Branch 2 is meant to soak in production for at least one release before Branch 3 enables the skip; this PR is intentionally a no-op-from-the-outside change with comprehensive plumbing.

## Reading order (6 commits, designed to review one-at-a-time)

| # | commit | scope |
|---|---|---|
| 1 | `2339a6f2` | raftengine: opt-in interfaces (`AppliedIndexReader` / `AppliedIndexWriter`) |
| 2 | `525fc152` | pebbleStore: `metaAppliedIndex` const + `LastAppliedIndex` + `SetDurableAppliedIndex` (with `pebble.Sync` UNCONDITIONALLY) |
| 3 | `aa9b8acc` | `MVCCStore` interface extension: `ApplyMutationsRaftAt` / `DeletePrefixAtRaftAt` overloads, threading appliedIndex through `applyMutationsWithOpts` + `deletePrefixAtWithOpts` |
| 4 | `7cd72bda` | kvFSM seam wiring: `AppliedIndexReader()` / `SetDurableAppliedIndex()` accessors + all 7 data-Apply leaves switched to `*RaftAt` with `f.pendingApplyIdx` |
| 5 | `f1e8748c` | engine hooks at BOTH snapshot persist sites: `persistCreatedSnapshot` + `e.persistLocalSnapshotPayload` call `SetDurableAppliedIndex` BEFORE `persist.SaveSnap` |
| 6 | `2c42f7d6` | tests (10 new tests across store + engine) |

## Design constraints honoured

All from `docs/design/2026_06_02_idempotent_snapshot_restore.md`:

- **§2 "Why both leaves"**: meta key bundle in BOTH `applyMutationsWithOpts` AND `deletePrefixAtWithOpts` so DEL_PREFIX entries don't silently leave `LastAppliedIndex` behind. Tested by `TestDeletePrefixAtRaftAt_BundlesMetaAppliedIndex`.
- **§3 `dbMu.RLock()`**: both `LastAppliedIndex` and `SetDurableAppliedIndex` acquire the read-lock, matching the lock-ordering discipline at `lsm_store.go:153 / :553 / :675`.
- **§4 fallback policy**: `AppliedIndexReader()` returns nil when the store doesn't implement the seam; `LastAppliedIndex` returns `(0, false, nil)` for missing OR truncated meta key. Branch 3 will then fall back to full restore conservatively.
- **§6 `ELASTICKV_FSM_SYNC_MODE=nosync` mode**: `SetDurableAppliedIndex` is **pinned to `pebble.Sync` unconditionally**. Rationale documented at length in the method's doc-comment — once `persist.SaveSnap` returns, WAL compaction discards every log entry ≤ `snap.Metadata.Index`, so there's no source to replay the meta key bump from. +1 fsync per snapshot persist (rare; default `SnapshotCount=10000`). Tested by `TestSetDurableAppliedIndex_UsesPebbleSync`.
- **§6 "HLC lease entries — checkpoint at snapshot persist"**: BOTH `persistCreatedSnapshot` (config snapshots) AND `e.persistLocalSnapshotPayload` (steady-state `SnapshotCount`-triggered hot path) call the hook. Both crash-ordering tested by `TestPersistCreatedSnapshot_*`.
- **§8 compatibility**: `StateMachine.Apply`'s public signature is unchanged. New interfaces are opt-in. Old call sites (`ApplyMutationsRaft` without `*At`) still work, just pass `appliedIndex=0` to opt out of the meta key bump.

## Test results

```
go vet ./...                                                    → 0 issues
go test ./store/ -short                                         → ok 29.4s
go test ./kv/    -short                                         → ok 10.4s
go test ./internal/raftengine/... -short                        → ok 32.8s
go test ./store/ -run 'TestLastAppliedIndex|TestSetDurable...|TestApply...|TestDelete...' → ok 1.6s
go test ./internal/raftengine/etcd/ -run 'TestRecording|TestPersistCreatedSnapshot_' → ok 0.03s
```

10 new tests added (see commit `2c42f7d6` for the full inventory).

## What this does NOT do

- **Does NOT enable the skip gate.** `restoreSnapshotState` still always restores. Branch 3 wires the `fsmAlreadyAtIndex` check + `applyHeaderStateOnSkip` + the two-phase `SnapshotHeaderApplier` seam.
- **Does NOT change `HEALTH_TIMEOUT_SECONDS=300`.** Branch 4 lowers it once Branch 3 has soaked.
- **Does NOT touch the snapshot-install hot path** (`Engine.applySnapshot`) per Non-Goals in the design.

## Soak plan

Branch 2 should run in production for at least one release before Branch 3 opens. Operators can verify the meta key is being written via:

```bash
# Inspect a pebble fsm.db (read-only)
ldb --db=/var/lib/elastickv/n3/fsm.db get '_meta_applied_index' --hex
# Expected: 8 little-endian bytes equal to the current applied index
```

## Refs

- PR #910 (design) — round 1..7 design history + retraction sections explaining the design constraints this PR honours
- PR #909 — `HEALTH_TIMEOUT_SECONDS` band-aid that this series eventually obviates

## Summary by CodeRabbit

* **New Features**
  * Durable tracking of Raft-applied indexes to ensure consistent snapshot/save ordering.

* **Bug Fixes**
  * Improved snapshot persistence reliability by pinning durable applied index before snapshot writes.
  * Stronger durability for writes bundled with Raft entry indices, reducing restore/recovery surprises.

* **Tests**
  * Added comprehensive tests covering applied-index ordering, failure handling, and persistence behavior.

### DIFF
--- a/docs/design/2026_06_02_idempotent_snapshot_restore.md
+++ b/docs/design/2026_06_02_idempotent_snapshot_restore.md
@@ -236,20 +236,36 @@ type AppliedIndexReader interface {
 }
 ```
 
-`kvFSM` exposes its store through a typed accessor so wal_store.go can
-reach it without importing the concrete pebble type:
+`kvFSM` directly satisfies `raftengine.AppliedIndexReader` by
+forwarding to its underlying store (PR #915 round-5 — round-1 had a
+factory `AppliedIndexReader() AppliedIndexReader` method intended to
+be called through a separate `AppliedIndexReporter` interface, but
+that pattern would require the skip gate to know about the reporter
+shim; the direct interface satisfaction is simpler and a compile-time
+guard catches future signature drift):
 
 ```go
-type AppliedIndexReporter interface {
-    AppliedIndexReader() AppliedIndexReader
-}
-func (f *kvFSM) AppliedIndexReader() AppliedIndexReader {
-    if r, ok := f.store.(AppliedIndexReader); ok {
-        return r
+// kv/fsm.go
+func (f *kvFSM) LastAppliedIndex() (uint64, bool, error) {
+    r, ok := f.store.(raftengine.AppliedIndexReader)
+    if !ok {
+        return 0, false, nil
     }
-    return nil
+    idx, present, err := r.LastAppliedIndex()
+    if err != nil {
+        return 0, false, errors.WithStack(err)
+    }
+    return idx, present, nil
 }
+
+// kv/fsm_applied_index_iface_check.go (compile-time guard)
+var _ raftengine.AppliedIndexReader = (*kvFSM)(nil)
+var _ raftengine.AppliedIndexWriter = (*kvFSM)(nil)
 ```
+
+The compile-time guard means any future rename or signature drift
+fails `go build` immediately — the soak investment is protected at
+the compiler level.
 
 ### 4. Conditional restore (with conservative error fallback)
 
@@ -279,16 +295,20 @@ func restoreSnapshotState(fsm StateMachine, snapshot raftpb.Snapshot, fsmSnapDir
 
 // fsmAlreadyAtIndex returns true ONLY when we can prove the FSM is
 // already at or past `want`. Any uncertainty -- FSM doesn't expose
-// the reporter, store doesn't expose the reader, read error, or
-// missing meta key -- returns false so we fall back to the full
-// restore. A stale-but-incorrect skip is far worse than a wasteful
-// full restore; the fallback errs strictly toward restoring.
+// the reader interface, read error, or missing meta key -- returns
+// false so we fall back to the full restore. A stale-but-incorrect
+// skip is far worse than a wasteful full restore; the fallback errs
+// strictly toward restoring.
+//
+// Direct type-assert against raftengine.AppliedIndexReader (PR #915
+// round-5): kvFSM satisfies the interface directly via its
+// LastAppliedIndex method, so no separate AppliedIndexReporter shim
+// is needed. The compile-time guard in kv/fsm_applied_index_iface_check.go
+// keeps this stable.
 func fsmAlreadyAtIndex(fsm StateMachine, want uint64) bool {
-    reporter, ok := fsm.(AppliedIndexReporter)
+    r, ok := fsm.(raftengine.AppliedIndexReader)
     if !ok { return false }
-    reader := reporter.AppliedIndexReader()
-    if reader == nil { return false }
-    have, present, err := reader.LastAppliedIndex()
+    have, present, err := r.LastAppliedIndex()
     if err != nil || !present { return false }
     return have >= want
 }
@@ -814,8 +834,12 @@ present.
 - `ApplyIndexAware` is **already** in `main`; this design only adds
   consumers.
 - The new opt-in interfaces (`AppliedIndexReader`,
-  `AppliedIndexReporter`, `SnapshotHeaderApplier`) are additive.
+  `AppliedIndexWriter`, `SnapshotHeaderApplier`) are additive.
   FSMs that don't implement them fall back to the current behaviour.
+  (Round-1 / round-2 of this doc mentioned an `AppliedIndexReporter`
+  factory-method shim; PR #915 round-5 superseded it by having
+  `kvFSM` satisfy `AppliedIndexReader` directly via its
+  `LastAppliedIndex` method.)
 - `metaAppliedIndexBytes` is new.  Older fsm.db files don't have it.
   The `present=false` branch makes the first restart after upgrade
   fall back to full restore, populating the meta key from the next
@@ -863,7 +887,7 @@ restoreSnapshotState skipped (FSM at index %d, snapshot at %d, ceiling=%d, cutov
 | Branch | Content | Behaviour change |
 |---|---|---|
 | **B1** (this PR) | Design doc | None |
-| **B2** | `ApplyMutationsRaftAt` / `DeletePrefixAtRaftAt` overloads + meta-key bundling in both leaves + `pebbleStore.LastAppliedIndex()` (under `dbMu.RLock()`) + `pebbleStore.SetDurableAppliedIndex()` (under `dbMu.RLock()`, **`pebble.Sync` unconditionally**) + `kvFSM.AppliedIndexReader()` accessor + `kvFSM.SetDurableAppliedIndex` forwarding + thread `f.pendingApplyIdx` into the data-Apply leaves + BOTH `persistCreatedSnapshot` (`engine.go:2679`) AND `e.persistLocalSnapshotPayload` (`engine.go:4032`, the SnapshotCount-triggered hot path) call `SetDurableAppliedIndex` BEFORE the corresponding `persist.SaveSnap` | Meta key starts being written on every data Apply AND at every snapshot persist (both config-snapshot and steady-state local-snapshot paths). Skip is still disabled. Soak in production for one release. |
+| **B2** | `ApplyMutationsRaftAt` / `DeletePrefixAtRaftAt` overloads + meta-key bundling in both leaves + `pebbleStore.LastAppliedIndex()` (under `dbMu.RLock()`) + `pebbleStore.SetDurableAppliedIndex()` (under `dbMu.RLock()` + `applyMu.Lock()` RMW monotonic guard, **`pebble.Sync` unconditionally**) + `kvFSM.LastAppliedIndex()` directly satisfies `raftengine.AppliedIndexReader` (compile-time guard in `kv/fsm_applied_index_iface_check.go`) + `kvFSM.SetDurableAppliedIndex` forwarding + thread `f.pendingApplyIdx` into the data-Apply leaves + BOTH `persistCreatedSnapshot` (`engine.go:2679`) AND `e.persistLocalSnapshotPayload` (`engine.go:4032`, the SnapshotCount-triggered hot path) call `SetDurableAppliedIndex` BEFORE the corresponding `persist.SaveSnap` | Meta key starts being written on every data Apply AND at every snapshot persist (both config-snapshot and steady-state local-snapshot paths). Skip is still disabled. Soak in production for one release. |
 | **B3** | `restoreSnapshotState` skip gate + `applyHeaderStateOnSkip(snapPath, tok.CRC32C)` orchestrating size + footer-vs-tokenCRC + full-body-CRC verification using `internal/raftengine/etcd`'s existing helpers (matching `openAndRestoreFSMSnapshot`'s safety contract) + two-phase `SnapshotHeaderApplier` seam on `kvFSM` (`ParseSnapshotHeader(r io.Reader) (ceiling, cutover, err)` + pure `ApplySnapshotHeader(ceiling, cutover)`) + metrics + INFO log | **User-visible cold-start win.** |
 | **B4** | Lower `HEALTH_TIMEOUT_SECONDS` default once production data shows steady-state skip rate ≥ 90 % | Tighter ceiling; the env override remains honoured. |
 
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/aws/smithy-go v1.25.1
 	github.com/cockroachdb/errors v1.13.0
 	github.com/cockroachdb/pebble/v2 v2.1.5
+	github.com/coreos/go-semver v0.3.1
 	github.com/emirpasic/gods v1.18.1
 	github.com/getsentry/sentry-go v0.46.2
 	github.com/goccy/go-json v0.10.6
@@ -58,7 +59,6 @@ require (
 	github.com/cockroachdb/redact v1.1.5 // indirect
 	github.com/cockroachdb/swiss v0.0.0-20251224182025-b0f6560f979b // indirect
 	github.com/cockroachdb/tokenbucket v0.0.0-20230807174530-cc333fc44b06 // indirect
-	github.com/coreos/go-semver v0.3.1 // indirect
 	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dustin/go-humanize v1.0.1 // indirect
--- a/internal/raftengine/etcd/engine.go
+++ b/internal/raftengine/etcd/engine.go
@@ -2676,25 +2676,37 @@ func (e *Engine) createConfigSnapshot(index uint64, confState raftpb.ConfState, 
 	}
 }
 
+// bumpDurableAppliedIndexBeforeSave pins the FSM's durable applied
+// index to `index` BEFORE the engine calls persist.SaveSnap, so a
+// successful snapshot persist always implies LastAppliedIndex >=
+// snap.Metadata.Index — closes the HLC-lease-only / encryption-only
+// fallback (PR #910 design §6).
+//
+// FSMs that do not expose raftengine.AppliedIndexWriter silently
+// no-op; the skip optimisation falls back to full restore for them
+// (legacy test fakes, in-memory backends). pebble.Sync is forced on
+// the writer side regardless of ELASTICKV_FSM_SYNC_MODE — once
+// persist.SaveSnap returns, WAL compaction discards every log entry
+// at or before snap.Metadata.Index, so there is no source to replay
+// the meta key bump from.
+//
+// Used by BOTH snapshot persist sites: persistCreatedSnapshot (this
+// file) and e.persistLocalSnapshotPayload (the steady-state
+// SnapshotCount-triggered hot path).
+func (e *Engine) bumpDurableAppliedIndexBeforeSave(index uint64) error {
+	w, ok := e.fsm.(raftengine.AppliedIndexWriter)
+	if !ok {
+		return nil
+	}
+	return errors.WithStack(w.SetDurableAppliedIndex(index))
+}
+
 func (e *Engine) persistCreatedSnapshot(snap raftpb.Snapshot) error {
 	if etcdraft.IsEmptySnap(snap) || e.persist == nil {
 		return nil
 	}
-	// Pin metaAppliedIndex to snap.Metadata.Index BEFORE SaveSnap so a
-	// successful snapshot persist always implies LastAppliedIndex >=
-	// snap.Metadata.Index — closes the HLC-lease-only / encryption-only
-	// fallback (PR #910 design §6). FSMs that do not expose
-	// raftengine.AppliedIndexWriter silently no-op; the skip
-	// optimisation falls back to full restore for them. pebble.Sync
-	// is forced on the writer side (see lsm_store.SetDurableAppliedIndex)
-	// regardless of ELASTICKV_FSM_SYNC_MODE — once SaveSnap returns,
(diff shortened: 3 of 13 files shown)


bootjp:
@claude review

PR #910 design Branch 2 implementation. 6 commits, no behaviour change yet (skip gate lands in B3); B2 is the soak phase.

Review checklist for this PR:
- §2 "why both leaves": is the metaAppliedIndex bundle present in BOTH `applyMutationsWithOpts` (1324) AND `deletePrefixAtWithOpts` (1393)? (verified with grep + tests)
- §3 dbMu.RLock(): `LastAppliedIndex` + `SetDurableAppliedIndex` both acquire the read-lock matching the discipline at lsm_store.go:153
- §4 strictly-additive fallback: `AppliedIndexReader()` returns nil for non-seam stores; `LastAppliedIndex` returns (0, false, nil) for missing OR truncated; never propagates an error that would abort cold start
- §6 nosync durability: `SetDurableAppliedIndex` uses `pebble.Sync` UNCONDITIONALLY (not `raftApplyWriteOpts()`) — the rationale is in the method comment; round-7 retraction in design doc explains why
- §6 both persist sites hooked: `persistCreatedSnapshot` (engine.go:2679) AND `e.persistLocalSnapshotPayload` (engine.go:4048), BEFORE `persist.SaveSnap` each
- crash-ordering: `TestPersistCreatedSnapshot_BumpErrorAborts` verifies that a failed bump prevents SaveSnap (no observable state where 'snapshot pointer durable, metaAppliedIndex stale')
- semantic-change audit: all 7 `f.store.ApplyMutationsRaft` / `DeletePrefixAtRaft` call sites in kv/fsm.go switched to `*RaftAt` with `f.pendingApplyIdx`; the existing non-At variants stay backwards-compatible (`appliedIndex=0` no-ops the meta key bump)

PR description has full reading order + soak plan + ldb verification command.

bootjp:
Round-2 push: `7001a934`. All blocking items from round-1 resolved.

## 🔴 claude[bot] BLOCKER — Site 2 crash-ordering test gap

Closed via two new tests:

- `TestPersistLocalSnapshotPayload_BumpsAppliedIndex` — asserts the recorded event sequence is exactly `[bump@123, save@123]`
- `TestPersistLocalSnapshotPayload_BumpErrorAborts` — mirror of Site 1: failed `SetDurableAppliedIndex` MUST surface the error AND prevent the free-function `persistLocalSnapshotPayload` from running

Discovery from writing the happy-path test: `buildLocalSnapshot` (`wal_store.go:544`) calls `storage.Term(applied)`. A fresh `etcdraft.MemoryStorage` has no entries, so without seeding, `Term` returns `ErrUnavailable` and the free function short-circuits before `SaveSnap`. `localSnapshotEngine(t, rec, fsm, applied)` now seeds the storage with `applied` entries to reach the actual SaveSnap path.

## 🟢 reviewdog golangci-lint (3 issues → 0)

- `lsm_store.go` gci: re-aligned the `metaAppliedIndex` const block (`golangci-lint fmt` picked it up automatically)
- `kv/fsm.go` wrapcheck on `SetDurableAppliedIndex` return: wrapped with `errors.WithStack`
- `engine.go` cyclop on `persistLocalSnapshotPayload` (12 > 10): extracted three helpers and folded the three "snapshot already moved on" error cases into a single switch arm:
  - `bumpDurableAppliedIndexBeforeSave(index)` — shared by BOTH persist sites, deduplicates the writer-seam type assertion and inline doc-comment (round-1 had it duplicated)
  - `handleLocalSnapshotPersistResult(err)` — collapses the four-arm switch
  - `purgeAfterLocalSnapshot()` — extracted the on-success disk-side cleanup

`golangci-lint run ./store/ ./kv/ ./internal/raftengine/etcd/` → **0 issues**.

## 🟢 Defensive forwards (push-back on gemini HIGH false positive)

`LeaderRoutedStore` and `ShardStore` gain `LastAppliedIndex` / `SetDurableAppliedIndex`. **The gemini HIGH finding was a false positive in current `origin/main`** — verified by grep:

```bash
$ grep -rn 'NewKvFSMWithHLC(' --include='*.go' . | grep -v _test.go
./cmd/server/demo.go:453:  fsm := kv.NewKvFSMWithHLC(st, hlc)         # st = *pebbleStore
./main.go:851:             sm := kv.NewKvFSMWithHLC(st, clock, ...)   # st = *pebbleStore
```

The kvFSM holds a `*pebbleStore` directly in every production path. `LeaderRoutedStore` is used by adapter/server code for read-routing (`cmd/server/demo.go:265-378` for grpc/redis/s3/dynamodb adapters); `ShardStore` is the coordinator-facing fanout wrapper. **Neither is `f.store`.** The skip optimisation works correctly in production.

The defensive forwards are added anyway because a future refactor that wraps the FSM store would otherwise silently degrade the skip to "full-restore-always" with no failure signal. `ShardStore.LastAppliedIndex` aggregates as **MIN over groups** so the cold-start skip gate refuses to skip whenever ANY group lags (conservative "over-restore beats under-restore" per design §4).

## 🟢 Other round-1 nits

- **claude[bot] note on non-monotonic SetDurableAppliedIndex**: kept the existing comment for now; the "logical correctness" reasoning (snapshot at M still satisfies its own invariant) is sound and a strictly-monotonic write would mean adding a read-modify-write under dbMu.Lock (not RLock), changing the lock-discipline more invasively than this round warrants. Worth a follow-up if production data shows the rare backward-write actually causes meaningful regression.
- **claude[bot] note on ShardStore multi-group convention**: the inline comment at `shard_store.go:1294` already documents the "tests MUST pass `appliedIndex=0`" convention; not strengthening to a runtime guard because no production caller hits this path.

## Verification

```text
go vet ./store/ ./kv/ ./internal/raftengine/...                    → exit 0
golangci-lint run ./store/ ./kv/ ./internal/raftengine/etcd/       → 0 issues
go test ./store/ ./kv/ ./internal/raftengine/... -short            → ok 53s
go test ./internal/raftengine/etcd/ -run 'TestPersistLocal...'     → ok 25ms (Site 2 new tests pass)
```

@claude review

bootjp:
Round-3 push: `b957e1aa` — addressing codex round-2 P2.

## 🟡 codex P2 line 632 — "Avoid lowering the durable apply index"

You're right that round-2's unconditional write can rewind `metaAppliedIndex` when the background `persistLocalSnapshot` goroutine fires after `ApplyMutationsRaftAt` has already advanced the key past the snapshot's index. claude[bot] round-1 had flagged this as a "design note"; the round-2 P2 promotion is the right call — even though the skip invariant (`LastAppliedIndex >= snap.Metadata.Index`) still holds for the snapshot itself, the rewind defeats the soak/verification value of the meta key (the future Branch 3 gate would unnecessarily fall back until another data Apply repairs it).

### Round-3 fix

`SetDurableAppliedIndex` is now a read-modify-write under `applyMu` serialisation:

```go
func (s *pebbleStore) SetDurableAppliedIndex(idx uint64) error {
    s.dbMu.RLock()
    defer s.dbMu.RUnlock()
    s.applyMu.Lock()
    defer s.applyMu.Unlock()

    existing, present, err := s.readAppliedIndexLocked()
    if err != nil { return err }
    if present && existing >= idx {
        return nil // skip-invariant still holds: existing >= idx
    }

    batch := s.db.NewBatch()
    defer func() { _ = batch.Close() }()
    if err := setPebbleUint64InBatch(batch, metaAppliedIndexBytes, idx); err != nil {
        return err
    }
    return errors.WithStack(batch.Commit(pebble.Sync))
}
```

Lock ordering: `dbMu.RLock` THEN `applyMu.Lock` matches `applyMutationsWithOpts` / `deletePrefixAtWithOpts` discipline (both acquire in that order), so no deadlock. The `applyMu.Lock` serialises against the per-Apply meta key writes — between the read and the batch commit, no concurrent Apply can advance the key.

`readAppliedIndexLocked` is the locked variant of the decode logic (factored out of `LastAppliedIndex`); same `(0, false, nil)` semantics for absent / truncated meta keys.

## Tests

Two new tests in `store/lsm_store_applied_index_test.go`:

- `TestSetDurableAppliedIndex_Monotonic` — simulates the exact race in single-threaded form: `ApplyMutationsRaftAt(applied=250)` then `SetDurableAppliedIndex(snapIdx=100)`. Asserts `LastAppliedIndex` still reads 250 after the older snapshot write.
- `TestSetDurableAppliedIndex_AdvancesOnlyForward` — `Set(50)` → `Set(100)` advances → `Set(75)` no-op → `Set(100)` no-op. Final state: 100.

```text
go vet ./store/                                                       → exit 0
golangci-lint run ./store/                                            → 0 issues
go test ./store/ -run 'TestSetDurable...|TestLast...|TestApply...'    → ok 0.59s
go test ./store/ ./kv/ ./internal/raftengine/... -short               → ok ~55s
```

@claude review

bootjp:
Round-4 push: `56848574`. claude[bot] round-3 nits + reviewdog gci. codex P2 line 1575 push-back below.

## 📝 codex round-3 P2 line 1575 — false positive (push-back)

> "deletePrefixAtWithOpts never takes that lock"

Factually incorrect. Verified directly:

```bash
$ grep -n 'func .s \*pebbleStore. deletePrefixAtWithOpts\|s\.applyMu' store/lsm_store.go | head -8
637:    s.applyMu.Lock()
638:    defer s.applyMu.Unlock()
1532: func (s *pebbleStore) deletePrefixAtWithOpts(...) error {
1536:   s.applyMu.Lock()
1537:   defer s.applyMu.Unlock()
```

`deletePrefixAtWithOpts:1536` acquires `s.applyMu.Lock()` immediately after `s.dbMu.RLock()` — same discipline as `applyMutationsWithOpts:637` and now `SetDurableAppliedIndex:637`. All three writers contend on the same `applyMu`, so the snapshot pin and a concurrent DEL_PREFIX are serialised; the read-modify-write inside `SetDurableAppliedIndex` (round-3) sees the post-DEL_PREFIX value of `metaAppliedIndex` and no-ops if `existing >= idx`.

The race codex describes ("the snapshot pin can still commit its stale lower N afterward and rewind the durable applied index") cannot occur with the current locking. The DEL_PREFIX path participates in the same monotonic/max protection by virtue of the shared `applyMu` and the round-3 RMW guard.

No code change for this finding.

## 🟢 claude[bot] round-3 nits — addressed

### #1 — `applyMu` missing from struct-level lock-ordering comment

`store/lsm_store.go:170-181` was stale (still listed `maintenanceMu → dbMu → mtx`). Round-4 extends it to `maintenanceMu → dbMu → applyMu → mtx` and documents `applyMu`'s purpose (serialises conflict-check → batch-commit for the raft-apply leaves and the `SetDurableAppliedIndex` RMW guard).

### #4 — Site 2 happy-path test should use exact slice match

`TestPersistLocalSnapshotPayload_BumpsAppliedIndex` now uses

```go
require.Equal(t, []orderEvent{
    {kind: "bump", index: index},
    {kind: "save", index: index},
}, rec.snapshot(), ...)
```

matching Site 1's style. A spurious third event (double SaveSnap, double bump) would fail the assertion — the previous `GreaterOrEqual + positional` would not. Also closes the coderabbit round-2 nit.

### #2 — `LastAppliedIndex` duplicates `readAppliedIndexLocked`

Deferred: claude[bot] explicitly called this out as "pre-existing; not a round-3 regression". `LastAppliedIndex` and `readAppliedIndexLocked` have identical Pebble-Get + error-handling bodies; the public API does `dbMu.RLock` only, the private helper assumes `dbMu.RLock + applyMu.Lock` already held. Refactoring `LastAppliedIndex` to call `readAppliedIndexLocked` would either require acquiring `applyMu` on every read (perf regression) or splitting the helper further (extra surface area). Worth doing but in a follow-up.

### #3 — `applyMu` held across `pebble.Sync` fsync

Doc-only note; same trade-off `applyMutationsWithOpts` already makes. Acceptable for soak-phase B2; revisit if production data shows tail-latency regression.

## 🟢 reviewdog gci on `store/lsm_store_applied_index_test.go:182`

Auto-corrected via `golangci-lint fmt`. Whitespace-only diff in the test's comment block.

## Verification

```text
go vet ./store/ ./kv/ ./internal/raftengine/...                         → exit 0
golangci-lint run ./store/ ./kv/ ./internal/raftengine/etcd/            → 0 issues
go test ./store/ ./internal/raftengine/etcd/ -run '<applied-index suite>' → ok
```

@claude review

bootjp:
Round-5 push: `a8e2cf52`. One real codex P2 fixed, one push-back.

## 🟡 codex P2 kv/fsm.go:145 — real bug, fixed

You're right. Round-1's `AppliedIndexReader() raftengine.AppliedIndexReader` was a factory method, intended to be called through an `AppliedIndexReporter` interface — but that interface was never added, and Branch 3's planned skip gate will do `fsm.(raftengine.AppliedIndexReader)` directly. The factory method does NOT satisfy that interface; the skip would always fall back even after meta keys are populated.

Renamed `kvFSM.AppliedIndexReader()` → `kvFSM.LastAppliedIndex() (uint64, bool, error)`, inlining the type-assert forward to `f.store.(raftengine.AppliedIndexReader)`. kvFSM now directly satisfies `raftengine.AppliedIndexReader`.

Added a separate compile-time guard file (`kv/fsm_applied_index_iface_check.go`):

```go
var _ raftengine.AppliedIndexReader = (*kvFSM)(nil)
var _ raftengine.AppliedIndexWriter = (*kvFSM)(nil)
```

so any future rename or signature drift fails at build time rather than silently degrading the skip optimisation. No existing production callers — verified by grep before renaming.

## 📝 codex P2 engine.go:4077 — out-of-scope for B2 (push-back with rationale)

> "the snapshot artifact does not contain the `_meta_applied_index` value just written here"

Correct observation, but the invariant claimed in PR #910 design §6 is **local-only**:

> "After a successful snapshot persist, LastAppliedIndex >= snapshot.Metadata.Index holds unconditionally"

This is a claim about the LIVE pebble store on the node that persisted the snapshot, NOT about the snapshot artifact's serialized state. The local cold-start skip path (Branch 3) reads the live store, not the artifact, so the invariant holds where it's used.

The codex finding's actual consequence:

| Path | Effect of stale metaAppliedIndex in artifact |
|---|---|
| Local cold-start | None — uses live store, which has post-bump value |
| Peer InstallSnapshot → restart | Peer's restored store has artifact's stale value; B3's skip gate would conservatively fall back to full restore |

The peer-after-InstallSnapshot path is **explicitly listed as a Non-Goal** in the design (`docs/design/2026_06_02_idempotent_snapshot_restore.md`):

> "Not changing the snapshot-install hot path (`Engine.applySnapshot`, `engine.go:1641`)."

When B3 wires the skip gate, the peer-after-InstallSnapshot scenario will conservatively fall back to full restore — which IS the strictly-additive guarantee the design promises (§4 fallback policy). Not a correctness bug; a missed optimisation.

**Why not bump before `fsm.Snapshot()`?** Because the FSM Snapshot is taken via `s.db.NewSnapshot()` which captures a consistent point-in-time view BEFORE the bump can occur. Bumping the live store first doesn't help: the pebble snapshot has already been captured. Closing this gap requires either:

1. Refactoring `maybePersistLocalSnapshot` to bump BEFORE calling `e.fsm.Snapshot()`. This changes the snapshot worker's flow significantly (the bump would need to be synchronous with the `e.applied` read, before queuing) and risks introducing other races.
2. Having `Engine.applySnapshot` (the InstallSnapshot receiver) explicitly call `SetDurableAppliedIndex(snap.Metadata.Index)` after `fsm.Restore` returns. Symmetric with `persistLocalSnapshotPayload` and parallel to the design's HLC ceiling / cutover restore handoff (PR #910 §5).

Option 2 is the right answer, but it belongs in Branch 3 (the path that consumes the meta key) or as a parallel follow-up — not in B2 (the soak-only PR). Filing as a TODO with a sentinel comment in `engine.applySnapshot` rather than expanding B2's scope.

## Verification

```text
go vet ./kv/ ./store/ ./internal/raftengine/...                    → exit 0
golangci-lint run ./kv/ ./store/ ./internal/raftengine/etcd/       → 0 issues
go test ./kv/ ./store/ ./internal/raftengine/etcd/ -short          → ok 56s
```

The new compile-time interface guard means any future regression would fail the build immediately:

```bash
$ go build ./kv/
$ # If kvFSM stops satisfying AppliedIndexReader:
$ go build ./kv/
./kv/fsm_applied_index_iface_check.go:7:5: cannot use (*kvFSM)(nil) ...
```

@claude review

bootjp:
Round-6 push: `39513379`. Closes both round-5 actionable items. Doc + comment only.

## 🟢 Item #1 — TODO sentinel in `applyReadySnapshot` ✅

Added a multi-line comment block right above the `openAndRestoreFSMSnapshot` call in `engine.go` (was around `:2001`, now `:2001-:2009` after the insert):

```go
// B3/follow-up: also call SetDurableAppliedIndex(tok.Index) here
// after Restore so peer-after-InstallSnapshot populates the meta
// key. The local-snapshot persist path already bumps the live
// store (engine.persistLocalSnapshotPayload), but the receiving
// node's restored store inherits the pre-bump value embedded in
// the snapshot artifact. Design Non-Goals §
// docs/design/2026_06_02_idempotent_snapshot_restore.md:71-74
// scopes this out of Branch 2; see PR #915 round-4/5 codex P2 on
// engine.go:4077 for the rationale.
```

Future B3 implementor will find the planned hook site + the round-4/5 rationale without re-discovering it.

## 🟠 Item #2 — Design doc stale pseudocode ✅

Three places in `docs/design/2026_06_02_idempotent_snapshot_restore.md` referenced the round-1 `AppliedIndexReporter` factory pattern that PR #915 round-5 superseded:

| Where | Before | After |
|---|---|---|
| §3 "kvFSM exposes its store" | factory `AppliedIndexReader() AppliedIndexReader` + `AppliedIndexReporter` interface | direct `LastAppliedIndex()` + compile-time guard from `kv/fsm_applied_index_iface_check.go` |
| §4 `fsmAlreadyAtIndex` pseudocode | `reporter, ok := fsm.(AppliedIndexReporter); reader := reporter.AppliedIndexReader()` | `r, ok := fsm.(raftengine.AppliedIndexReader)` direct |
| Implementation Plan B2 row | "kvFSM.AppliedIndexReader() accessor" + no mention of monotonic guard | "kvFSM.LastAppliedIndex() directly satisfies raftengine.AppliedIndexReader (compile-time guard)" + applyMu.Lock() RMW from round-3 |
| §8 compatibility list | `AppliedIndexReporter` listed as new opt-in | replaced by `AppliedIndexWriter` (actually shipped) + retrospective note |

Three retrospective mentions of `AppliedIndexReporter` are kept as "this is NOT what we use" — so future readers understand the design history without re-deriving it.

## Verification

```text
go vet ./kv/ ./store/ ./internal/raftengine/...                    → exit 0
golangci-lint run ./kv/ ./store/ ./internal/raftengine/etcd/       → 0 issues
```

No code semantics changed in round-6. The compile-time interface guards from round-5 remain the structural protection.

@claude review

PR #916, opened 2026-06-03
feat(jepsen): M5a — dynamodb-append-multi-table-workload

bootjp:
**Draft** — second slice of Composed-1 M5a per `docs/design/2026_06_02_proposed_composed1_m5_jepsen_route_shuffle.md` §3.3. Follows #911 (CLI tools, merged); blocks on launch script + setup-hook verification (remaining M5a slices) before unmark-draft.

## What this PR ships

`jepsen/src/elastickv/dynamodb_multi_table_workload.clj` — a new workload variant of the DynamoDB list-append test that exercises cross-table 2PC, which the single-table `elastickv.dynamodb-workload` cannot do.

### Why a separate workload exists

`kv/shard_key.go`'s `dynamoRouteKey` normalises every DynamoDB table-meta / item / GSI key for one table to a single per-table route key (`!ddb|route|table|<base64(name)>`). Single-table workloads always route to one shard regardless of partition-key value, so `dispatchMultiShardTxn` / `commitSecondaryTxns` / the new `ErrTxnSecondaryRouteShiftedAfterPrimaryCommit` sentinel never fire (codex P1 on #905 ffb9c73f).

### What this workload does

- **N=4 tables** (`jepsen_append_t1` … `jepsen_append_t4`) created in `setup!`.
- **Deterministic key routing**: Elle's integer keys `k` route via `(mod k N)+1 → table-idx`, so each Elle key always lands at the same `(table, pk)` storage location (Elle's append checker still sees a single key namespace).
- **Default 4-mop txns with key-count=12** distribute evenly across all 4 tables — a 4-key txn touches all 4 tables and therefore both Raft groups under the launch script's planned `--shardRanges` layout (t1-t2 → group 1, t3-t4 → group 2).
- **`TransactGetItems`** for atomic cross-table pre-read; **`TransactWriteItems`** for atomic cross-table writes (both DynamoDB APIs natively support multi-table txns).
- **Same OCC condition expressions** as the single-table workload so any G1c or G-single anomaly observed is attributable to the multi-shard 2PC path rather than to a workload-shape difference.

### Why self-contained

The file intentionally duplicates small helpers from `elastickv.dynamodb-workload` rather than extracting into a shared namespace. Refactoring would couple the workloads through a maintained API and obscure the side-by-side comparability that is the point of running both — relevant because #905 commits to keeping the single-table workload as-is for trend comparison.

## Tests (`jepsen/test/elastickv/dynamodb_multi_table_workload_test.clj`)

7 tests / 17 assertions. The load-bearing ones:

- **`key-routing-distributes-across-all-tables`** — 12 keys distribute 3-per-table across 4 tables
- **`key-routing-table-name-matches-prefix`** — wraparound at k=4
- **`key-routing-pks-disambiguate_colliding_keys`** — k=0 and k=4 share table t1 but have distinct pks
- **`multi-op-txn-spans_multiple_groups`** — default 4-key txn touches BOTH the planned group-1 tables (1,2) and group-2 tables (3,4) — without this, `dispatchMultiShardTxn` never fires and the whole workload is a no-op

The routing helpers are private; tests access via `var-get` on `#'workload/key->table-idx` etc. so a future change to the routing surfaces in this test rather than as a silent G1c during a Jepsen run.

## Remaining M5a work (separate PRs after this lands)

- [ ] `scripts/run-jepsen-m5-local.sh` — single-process two-group launch with `--raftGroups` + `--shardRanges` using `elastickv-route-key` to compute boundary keys
- [ ] Setup-hook verification — Clojure `ListRoutes` gRPC client asserting multi-group routing is in place
- [ ] Done-when check — workload exercises `dispatchMultiShardTxn` (server-side log marker or probe) and Elle finds zero G1c

## Test plan

- [x] `lein test elastickv.dynamodb-multi-table-workload-test` → Ran 7 tests / 17 assertions, 0 failures, 0 errors
- [x] `lein test elastickv.dynamodb-workload-test` (existing single-table suite) → Ran 3 tests / 5 assertions, 0 failures (no regression)
- [ ] End-to-end Jepsen run (gated on launch script + setup hook landing)

### DIFF
--- a/jepsen/src/elastickv/dynamodb_multi_table_workload.clj
+++ b/jepsen/src/elastickv/dynamodb_multi_table_workload.clj
@@ -292,6 +292,13 @@
       (assoc this :ddb (make-ddb-client host port))))
 
   (setup! [this _test]
+    ;; No DescribeTable poll loop is needed: elastickv's adapter
+    ;; returns TableStatus=ACTIVE synchronously in the CreateTable
+    ;; response (adapter/dynamodb.go:779-783), so the table is
+    ;; queryable as soon as create-one-table! returns.  Real DynamoDB's
+    ;; CreateTable is async and would need a poll; if this workload is
+    ;; ever pointed at a real DynamoDB endpoint, add a DescribeTable
+    ;; loop here (claude[bot] design note on PR #916 c46b2b5e).
     (create-all-tables! ddb))
 
   (teardown! [_this _test]
--- a/jepsen/src/elastickv/dynamodb_multi_table_workload.clj
+++ b/jepsen/src/elastickv/dynamodb_multi_table_workload.clj
@@ -145,6 +145,31 @@
   (doseq [t table-names]
     (create-one-table! ddb t)))
 
+(defn- delete-one-table!
+  "Delete one of the N tables; ignore ResourceNotFoundException so
+   teardown! is idempotent across cluster states (e.g. a re-run after
+   a partial setup failure)."
+  [ddb tname]
+  (try
+    (ddb-invoke! ddb :DeleteTable {:TableName tname})
+    (catch clojure.lang.ExceptionInfo e
+      (when-not (= "ResourceNotFoundException" (:type (ex-data e)))
+        (throw e)))))
+
+(defn- delete-all-tables!
+  "Delete every table this workload owns.  Idempotent.
+
+   Without this, leftover Items from a previous run remain in the four
+   tables across runs; the next run's TransactGetItems pre-read picks
+   them up as snapshot values, the txn's :r mops report those values
+   to Jepsen, and Elle sees a history that contains writes it never
+   generated — leading to spurious G-single / G1c false positives
+   (claude[bot] medium on PR #916).  Cleaning up here keeps each run's
+   history closed under the writes Elle actually generated."
+  [ddb]
+  (doseq [t table-names]
+    (delete-one-table! ddb t)))
+
 (defn- list-attr
   "Convert a Clojure vector of longs to a DynamoDB {:L [...]} attribute."
   [vs]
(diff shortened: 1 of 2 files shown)


bootjp:
Workload draft + tests are stable; gemini's first pass returned zero findings. Could use a proper code review before iterating the remaining M5a slices (launch script + setup-hook verification).

@claude review

bootjp:
Pushed as c46b2b5e — claude[bot]'s findings on aded8b96 addressed.

**Medium — stale table data across runs.** `teardown!` was a no-op, so the next run's `TransactGetItems` pre-read picked up leftover Items as snapshot, `:r` mops reported them to Jepsen, and Elle saw a history containing writes it never generated.

**Fix:** added `delete-all-tables!` mirroring `create-all-tables!`'s exception handling (ignore `ResourceNotFoundException` for idempotency), and called it from the client's `teardown!`.

**Low — TransactGetItems positional alignment unverified.** One-line `assert` on `(count ks-vec) = (count :Responses)` so a future cognitect SDK that normalises the response surfaces loudly.

**Low — `key-count` must be multiple of `num-tables`.** `--key-count 10` would silently give some tables fewer keys, breaking the multi-group-span invariant. Construction-time `assert` fails-fast; two new tests pin both reject (10, 7) and accept (4, 8, 16) cases.

**Trivial — deftest naming.** Two test names had stray underscores; renamed to project's hyphen-case convention.

**Deferred** (claude[bot] explicitly tagged as low-priority):
- `ConditionalCheckFailedException` dead-code arm — mirrors single-table workload, kept for parity until a single-table cleanup lands.
- `setup!` doesn't wait for ACTIVE state — real DynamoDB's `CreateTable` is async but elastickv's may be synchronous; will verify at the end-to-end Jepsen run and add a `DescribeTable` poll if needed.

`lein test elastickv.dynamodb-multi-table-workload-test` → Ran 9 tests / 22 assertions, 0 failures, 0 errors.

@claude review

bootjp:
You linked claude[bot]'s initial review of aded8b96 (the 6-finding pass). All 4 actionable findings from that review were addressed in c46b2b5e and confirmed by claude[bot]'s follow-up at 14:54:50 — both deferrals were explicitly OK'd:

- `ConditionalCheckFailedException` dead-code arm — kept for parity with the single-table workload; claude said "Fine; a future cleanup can remove both together."
- `setup!` doesn't poll for ACTIVE state — claude said "gating this on the end-to-end Jepsen run is correct."

The second deferral was the only one with an unverified assumption: claude[bot] flagged it would matter "if elastickv's `CreateTable` is async like real DynamoDB." Locked that down with actual code evidence in cf07ec01 — `adapter/dynamodb.go:779-783` returns `TableStatus=ACTIVE` synchronously in the response body, so the table is queryable as soon as `create-one-table!` returns. Added a comment at the `setup!` call site so a future reader (or someone retargeting this workload at real DynamoDB) doesn't have to re-do the adapter inspection.

Doc-only; `lein test` still 9 / 22 passing.

@claude review

PR #908, opened 2026-06-02
feat(encryption): Stage 6E-2a — ErrEnvelopeCutoverInProgress typed error

bootjp:
## Summary

First foundational slice for Stage 6E-2 (engine unwrap + coordinator wrap + §7.1 barrier atomic flip). Adds the typed sentinel `ErrEnvelopeCutoverInProgress` that `Engine.Propose` (and the coordinator wrap-on-propose path) will return while the §7.1 proposal-quiescence barrier owned by `EnableRaftEnvelope` is open.

## Why this is split out

Stage 6E-2 has four interlocked pieces that must ship in one PR before the gate flips (engine unwrap path exists since Stage 3; coordinator wrap-on-propose; §7.1 barrier in admin handler; gate flip). This slice is a **purely additive foundation** — no existing code path changes, no behavior change. It exists so the upcoming 6E-2b (Propose source-tag plumbing) and 6E-2c (EnableRaftEnvelope barrier) PRs have a stable typed-error contract to import.

## What lands

- `internal/raftengine/etcd/encryption.go` — new `ErrEnvelopeCutoverInProgress` sentinel, alongside the existing `ErrRaftUnwrapFailed`. Doc comment captures:
  - The barrier-admitted source classes (cutover entry + ConfChange-time `RegisterEncryptionWriter`, both via `source="encryption_admin"`), explaining the deadlock-on-own-proposal exemption.
  - Caller policy: retryable; barrier completes in drain-time.
  - Distinction from `ErrRaftUnwrapFailed` (process-fatal).
- `internal/raftengine/etcd/encryption_test.go` — `TestErrEnvelopeCutoverInProgress_DistinctFromUnwrapFailure` pins the `errors.Is` asymmetry so future error-chain refactors cannot silently merge the two sentinels.

## Test plan

- [x] `go test -run 'ErrEnvelopeCutoverInProgress' ./internal/raftengine/etcd/` — pass.
- [x] `golangci-lint run` on the touched package — 0 issues.
- [x] No existing test affected (purely additive).

## Self-review (per CLAUDE.md)

1. Data loss — none; new variable, no execution path changes.
2. Concurrency — none; the sentinel is a constant.
3. Performance — none; package-level `errors.New` runs once at init.
4. Data consistency — none; the sentinel will gate retryable vs fatal classification in future slices.
5. Test coverage — one focused test pinning the cross-sentinel `errors.Is` asymmetry.

## Caller audit (semantic-change directive)

[...]

PR #917, opened 2026-06-03
feat(encryption): Stage 6E-2b — ProposeAdmin sibling on raftengine.Proposer

bootjp:
## Summary

Stage 6E-2b: extend `raftengine.Proposer` with a sibling `ProposeAdmin` method, implement on every existing implementer (the etcd engine + 14 test mocks + the kv `wrappedProposer`), and migrate the two production control-plane call sites to it.

In this build `Propose` and `ProposeAdmin` are operationally identical — the etcd engine routes through a shared private `propose()` impl, and the kv wrapper's `ProposeAdmin` skips the configured `RaftPayloadWrapper` (the cutover entry must remain cleartext at `index == cutover` for the §6.3 strict-`>` dispatch to find it).

Stage 6E-2d will diverge the two paths: `Propose` gains a §7.1 quiescence-barrier check that rejects with the `ErrEnvelopeCutoverInProgress` sentinel (already defined in 6E-2a, #908), while `ProposeAdmin` remains admissible. This PR is the prerequisite migration — call sites that should bypass the barrier must move to `ProposeAdmin` **before** the barrier ships, otherwise they would fail closed at the moment 6E-2d lands.

## Production callsite migration

- `adapter/encryption_admin.go`'s `proposeEncryptionEntry` helper — composes BootstrapEncryption, EnableStorageEnvelope, the §6E EnableRaftEnvelope cutover marker, and every other admin op. All exempt from the barrier.
- `main_encryption_registration.go`'s local ConfChange-time writer registration (Stage 7c §3.1). Without admin-path exemption, a new member joining mid-barrier could not register its writer entry.

## Tests added

- `TestProposeAdmin_EquivalentToProposeToday` — pins that both methods commit identical entries through the same Raft path, reach the FSM verbatim, and advance `CommitIndex`. This is the bridge property the migration relies on.
- `TestProposerInterface_RequiresProposeAdmin` — compile-time guard against silent reverts of the interface extension.
- `TestWrappedProposer_ProposeAdminBypassesWrap` — pins the wrap-bypass invariant: even with a wrap closure configured, `ProposeAdmin` payloads reach `inner.ProposeAdmin` cleartext (not `inner.Propose` and not wrapped).

## Five-lens self-review

- **Data loss**: no FSM apply path changed; `Propose` body refactored to `return e.propose(ctx, data)`; `proposalRequest` layout unchanged; no new error swallow.
- **Concurrency / distributed**: `ProposeAdmin` reuses the same `proposalRequest` channel and the same `nextID()` atomic; no new shared state.
- **Performance**: one extra function-call frame per propose (the extracted helper), inlined by the compiler; no new allocations, no new locks.
- **Data consistency**: both methods exercise the identical raft proposal path so consensus / commit-index invariants are preserved. The 6E-2d barrier will live above `propose()`, not inside it.
- **Test coverage**: engine-level equivalence + interface compile-time guard + wrap-bypass pin. Every test mock updated in lockstep; `recordingProposer` / `applyingProposer` in the encryption-admin suite collapse `ProposeAdmin` to `Propose` so the 50+ existing assertions on `.calls` continue to work after the production callsite switch.

## Caller audit (per the `/loop` semantic-change rule)

The interface gained a method; every implementer either picks up the new method explicitly (production) or forwards to `Propose` (test mocks). 2 production callsites migrated to `ProposeAdmin`; 15 implementers carry the method; no caller can silently regress because the interface itself enforces it at compile time.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `golangci-lint run` clean (0 issues; pre-commit hook ran)
- [x] `go test -race ./internal/raftengine/etcd/` passes
- [x] `go test -race ./kv/` passes (scoped to the affected tests)
- [x] `go test -race ./internal/raftadmin/` passes
- [x] Encryption-admin / EnableRaftEnvelope / EnableStorageEnvelope / Bootstrap / Rotation / Registration tests pass in `./adapter/`
- [x] Root-package test `go test -race .` passes

Refs: `docs/design/2026_05_31_partial_6e_enable_raft_envelope.md` §7.1

### DIFF
--- a/adapter/encryption_admin.go
+++ b/adapter/encryption_admin.go
@@ -1343,11 +1343,24 @@ func (s *EncryptionAdminServer) proposeEncryptionEntry(ctx context.Context, opco
 	entry = append(entry, body...)
 	// proposeEncryptionEntry composes control-plane entries
 	// (BootstrapEncryption, EnableStorageEnvelope, the §6E
-	// EnableRaftEnvelope cutover marker, etc.). All of these are
-	// exempt from the §7.1 quiescence barrier that Stage 6E-2d
-	// installs on Propose: most carry no user-data, and the
-	// cutover entry itself would deadlock the barrier on its own
-	// proposal. ProposeAdmin is the exempt path.
+	// EnableRaftEnvelope cutover marker, RotateDEK,
+	// RegisterEncryptionWriter, etc.). The §7.1 quiescence barrier
+	// Stage 6E-2d installs on Propose would reject these — most
+	// trivially, the cutover entry itself would deadlock on its
+	// own barrier — so admin ops route through the
+	// barrier-exempt ProposeAdmin path.
+	//
+	// ProposeAdmin is barrier-exempt only; the wrap layer above
+	// the engine (kv.wrappedProposer, when configured) still
+	// applies its wrap closure to ProposeAdmin payloads, so a
+	// post-cutover RotateDEK or RegisterEncryptionWriter
+	// committed at `index > raftEnvelopeCutoverIndex` carries the
+	// AEAD envelope the §6.3 strict-`>` apply hook expects. The
+	// EnableRaftEnvelope cutover marker (at `index == cutover`)
+	// must remain cleartext for strict-`>` to leave it alone;
+	// today s.proposer is wired to the raw engine (see
+	// main_encryption_admin.go), so the marker reaches Raft
+	// without the wrap layer in the path at all.
 	res, err := s.proposer.ProposeAdmin(ctx, entry)
 	if err != nil {
 		return 0, proposeErrorToStatus(err, opcode)
--- a/adapter/encryption_admin.go
+++ b/adapter/encryption_admin.go
@@ -1341,7 +1341,14 @@ func (s *EncryptionAdminServer) proposeEncryptionEntry(ctx context.Context, opco
 	entry := make([]byte, 0, 1+len(body))
 	entry = append(entry, opcode)
 	entry = append(entry, body...)
-	res, err := s.proposer.Propose(ctx, entry)
+	// proposeEncryptionEntry composes control-plane entries
+	// (BootstrapEncryption, EnableStorageEnvelope, the §6E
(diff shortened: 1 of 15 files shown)


bootjp:
@claude review

Round-1 round-up:
- **Gemini medium** (engine_test.go:243 — non-idiomatic compile-time interface assertion wrapped in test function): fixed by moving `var _ raftengine.Proposer = (*Engine)(nil)` to package scope so the guard fires at test-package compile time rather than being deferred to a test target that nobody selects. f2fd97a2.

Head: f2fd97a2.

bootjp:
@claude review

Round-1 round-up:

- **Codex P1** (adapter/encryption_admin.go: post-cutover admin entries via ProposeAdmin would land cleartext above cutover, halting apply): root cause was the round-1 design making `wrappedProposer.ProposeAdmin` skip the wrap layer. Corrected: ProposeAdmin's sole divergence from Propose is the §7.1 barrier exemption — wrap behaviour is identical. The EnableRaftEnvelope cutover marker stays cleartext via call-site routing (raw engine reference bypasses the wrap layer entirely), not via a method-level bypass.
- **Codex P1** (main_encryption_registration.go: same shape on leader-restart registration): same root cause, same fix. The current cleartext-via-raw-engine path is correct pre-cutover; the post-cutover wrap is a known gap documented as 6E-2c/2e wiring work in the call-site comment.
- **Gemini medium** (round-0): compile-time `var _ raftengine.Proposer = (*Engine)(nil)` moved from inside a test function to package scope per Go idiom. Already addressed in f2fd97a2.

Semantic-change audit: 3 production ProposeAdmin sites; no behavioural change in this build (the two raw-engine sites are unchanged; wrappedProposer is not on any admin path today). The fix removes a wrong design lock-in so the 6E-2c/2e wiring cannot copy-paste the round-1 wrap-bypass.

Test `TestWrappedProposer_ProposeAdminBypassesWrap` was pinning the wrong behaviour; renamed to `TestWrappedProposer_ProposeAdminAppliesWrap` and inverted.

Head: 4530e642.

PR #920, opened 2026-06-03
fix(adapter/dynamodb): option-2 one-phase dedup for single-item writes

bootjp:
## Summary

Fixes the `:duplicate-elements` anomaly that the 2026-06-03 scheduled Jepsen run [26856696842](https://github.com/bootjp/elastickv/actions/runs/26856696842) surfaced on the **DynamoDB** list-append workload — value `72` stored twice in key 18.

This is the **same anomaly class** as the Redis bug behind PR #796/#884/#887 and the parent design [`2026_05_21_proposed_txn_secondary_idempotency.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_05_21_proposed_txn_secondary_idempotency.md), on a different adapter. The option-2 dedup landed for Redis only and is gated off; the parent doc states plainly *"DynamoDB / S3 / SQS do not route through the dedup loop."* This PR extends it to the DynamoDB single-item write path.

## Root cause (evidence)

- The generator emitted `72` to key 18 **exactly once** (all 250 invoke-appends to key 18 are unique). The op is a single-micro-op `[[:append 18 72]]` (process 21, `00:47:49.875` → `00:47:50.661`), routed through `dynamo-append!` → one `UpdateItem` with `SET #v = list_append(...)`.
- The demo-cluster log shows a **Raft leader-election storm** (terms 7→8→9→10→11) within `00:47:49`–`00:47:50`, with `hlc lease renewal failed: ... not leader` at the same instant.
- The DynamoDB single-item write (`retryItemWriteWithGeneration`) is a **read-modify-write that recomputes on retry**. Attempt 1 commits `[A,B,72]` at `C1` but, because the caller-supplied `StartTS=readTS` is not reset by the coordinator's leader-change retry, the new leader's OCC sees `item@C1 > StartTS` and returns a **self-inflicted `WriteConflict`** (`store/mvcc_store.go:533`). `isRetryableTransactWriteError` treats it as retryable, the retry re-reads `[A,B,72]` and re-appends → `[A,B,72,72]`.

## Mechanism (gated behind `onePhaseTxnDedup`, default off)

1. The adapter now allocates `commitTS` locally via `Clock().NextFenced()` (HLC-4 fence; `ErrCeilingExpired` is non-retryable → surfaces to the client), so it knows the value to carry as `PrevCommitTS` on a retry.
2. On a retryable write error the retry **reuses** the failed attempt's write set (`reusableItemWrite`) under a fresh `commitTS` carrying `PrevCommitTS`; the existing FSM probe `dedupProbeOnePhase` (`kv/fsm.go`) no-ops the apply if the prior attempt landed at exactly that ts.
3. A genuine `WriteConflict` (the adapter-side self-conflict guard's `CommittedVersionAt(probeKey, commitTS)` misses) drops the pending attempt and recomputes from a fresh read (outcome 3) — no duplicate, no lost update.
4. `kv.PrimaryKeyForElems` is exported so the adapter-side guard probes the **same key** the FSM uses as `meta.PrimaryKey` (the lexicographically-smallest write key; for the GSI-free `jepsen_append` table that is the itemKey).

**Gate off ⇒ `retryItemWriteWithGeneration` is byte-identical** (the dedup is a single `if d.onePhaseTxnDedup { ... }` early return on a default-false field). Enable via `WithDynamoOnePhaseTxnDedup` / `ELASTICKV_DYNAMODB_ONEPHASE_DEDUP=1` only after the whole cluster runs a probe-aware binary (R5).

## Scope

- **In:** the shared single-item loop `retryItemWriteWithGeneration` (backs `UpdateItem` / `PutItem` / `DeleteItem`).
- **Out (follow-up):** `TransactWriteItems` (separate path; already surfaces conflicts as a definite `TransactionCanceledException` → `:fail`), and S3 / SQS (same latent class).

## Design doc

`docs/design/2026_06_03_proposed_dynamodb_onephase_dedup.md` (proposed; committed first in this PR).

## Tests (`adapter/dynamodb_onephase_dedup_test.go`)

Reuse the OCC-aware `dedupTestCoordinator` (real `store.ApplyMutations` OCC + FSM-probe mimic), so a passing dedup test proves the probe is load-bearing.

| Test | Scenario | Pins |
|---|---|---|
| `LandedPriorAttempt_NoDuplicate` | attempt 1 lands then errors | reuse FSM-probe hits; list has one `3`; dispatches=2, probeNoOps=1 |
| `PriorAttemptDidNotLand_Applies` | attempt 1 pre-rejects | reuse applies fresh; one `3` |
| `SelfInflictedReuseConflict_ReturnsSuccess` | reuse lands then errors | adapter self-conflict guard probes fresh commitTS; cached result; no recompute |
| `GenuineConflictRecomputes` | concurrent write advances item past snapshot | reuse OCC-conflicts; probe misses; recompute appends once on top → `[1,2,99,3]` |
| `DisabledKeepsLegacyPath` | gate off | characterizes the duplicate `[1,2,3,3]` the gate fixes |

## Validation

- `go test ./adapter/ -run 'TestItemWriteDedup'` — pass (5/5).
- `go test ./adapter/ -run 'UpdateItem|PutItem|DeleteItem|TransactWrite|BatchWrite|Dedup|Migration'` — pass (item-write paths touched by the change).
- `go test ./kv/ ./store/` — pass.
- `golangci-lint run ./adapter/... ./kv/...` — 0 issues.
- Full `./adapter/...` and `-race` left to CI (the suite is slow locally; the change is gate-off-identical so existing tests exercise unchanged code).

## Self-review (5 passes)

1. **Data loss** — No committed write lost. Reuse re-dispatches the same write set; the FSM no-op only fires when the prior attempt's version exists at exactly `PrevCommitTS` (exact-ts, deterministic). Genuine conflict recomputes rather than dropping the append. Gate-off path unchanged.
2. **Concurrency / distributed** — The self-inflicted vs genuine conflict discriminator is the exact-ts probe + OCC check (R3). `NextFenced` keeps commitTS above the physical ceiling so a stale leader cannot mint a colliding ts. Race-freedom (E1 applies before E2 or never) inherited from the parent doc's Raft-log-order argument.
3. **Performance** — Hot path (first attempt, `PrevCommitTS==0`) untouched: one local `NextFenced` (already on the write path) and no probe. The extra point-read only happens on the retry path. No added Raft round-trips.
4. **Data consistency** — New data is written only at a fresh monotonic `commitTS` (SI preserved); the stale `commitTS` is read-only in the probe (dedup identity). Adapter guard and FSM probe agree on the probe key via `kv.PrimaryKeyForElems` (R4).
5. **Test coverage** — All three outcomes + self-conflict guard + gate-off legacy duplicate are pinned against real OCC + the real probe.

Caller audit (semantic-change rule): the three callers of `retryItemWriteWithGeneration` (`updateItemWithRetry`, `putItemWithRetry`, `deleteItemWithRetry`) all return `(*itemWritePlan, error)` with the same contract; the result is invariant across reuse (R1), so the cached `plan` is correct on a dedup no-op for all three. Gate-off ⇒ identical behavior.

## Summary by CodeRabbit

* **New Features**
  * DynamoDB now supports optional write deduplication for single-item operations (UpdateItem, PutItem, DeleteItem). Enable via `ELASTICKV_DYNAMODB_ONEPHASE_DEDUP=1` to prevent duplicate writes on retryable failures during network disruptions or leadership transitions.

* **Tests**
  * Added comprehensive test coverage for DynamoDB write deduplication scenarios.

### DIFF
--- a/adapter/dynamodb.go
+++ b/adapter/dynamodb.go
@@ -141,7 +141,7 @@ type DynamoDBServer struct {
 	// under a fresh commit_ts and carries prev_commit_ts so the FSM no-ops a
 	// commit that already landed under leadership churn (the :duplicate-elements
 	// anomaly). It MUST stay off until every node runs a probe-aware binary —
-	// see R5 in docs/design/2026_06_03_proposed_dynamodb_onephase_dedup.md.
+	// see R5 in docs/design/2026_06_03_partial_dynamodb_onephase_dedup.md.
 	// Default off; enabled via WithDynamoOnePhaseTxnDedup or the
 	// ELASTICKV_DYNAMODB_ONEPHASE_DEDUP env var.
 	onePhaseTxnDedup bool
@@ -1261,10 +1261,34 @@ func (d *DynamoDBServer) retryItemWriteWithGeneration(
 	// reuse the failed attempt's write set under a fresh commit_ts + prev_commit_ts
 	// so the FSM no-ops a commit that already landed under leadership churn,
 	// instead of re-reading and re-appending (the :duplicate-elements anomaly).
-	// See docs/design/2026_06_03_proposed_dynamodb_onephase_dedup.md.
-	if d.onePhaseTxnDedup {
+	// See docs/design/2026_06_03_partial_dynamodb_onephase_dedup.md.
+	//
+	// Leader-only (codex P1, PR #920): the dedup path allocates commit_ts from
+	// the LOCAL HLC and carries it as prev_commit_ts, so that timestamp MUST be
+	// leader-issued to stay globally unique — otherwise two frontends could mint
+	// the same commit_ts in one millisecond and the exact-ts probe would dedup
+	// against the wrong writer's version, losing an update. On the leader the
+	// single HLC issues monotonic unique values, and NextFenced's physical-ceiling
+	// fence keeps a deposed leader's window disjoint from its successor's. A
+	// non-leader (reachable only when no leaderMap HTTP proxy forwards follower
+	// ingress) falls back to the legacy path, where Coordinator.Dispatch redirects
+	// to the leader and the LEADER allocates commit_ts — never this follower's HLC.
+	if d.onePhaseTxnDedup && d.coordinator.IsLeader() {
 		return d.retryItemWriteWithGenerationDedup(ctx, tableName, exhaustedMessage, prepare)
 	}
+	return d.retryItemWriteWithGenerationLegacy(ctx, tableName, exhaustedMessage, prepare)
+}
+
+// retryItemWriteWithGenerationLegacy is the pre-dedup retry loop: it recomputes
+// the write set from a fresh read on every retryable error. It is the active
+// path whenever the dedup gate is off or this node is not the leader, so it
+// stays byte-identical to the pre-feature behavior.
+func (d *DynamoDBServer) retryItemWriteWithGenerationLegacy(
+	ctx context.Context,
+	tableName string,
+	exhaustedMessage string,
+	prepare func(readTS uint64) (*itemWritePlan, error),
+) (*itemWritePlan, error) {
 	backoff := transactRetryInitialBackoff
 	deadline := time.Now().Add(transactRetryMaxDuration)
 	for range transactRetryMaxAttempts {
@@ -1309,7 +1333,7 @@ func (d *DynamoDBServer) retryItemWriteWithGeneration(
 // leadership churn: attempt 1 commits at C1 but returns a WriteConflict, the
 // retry re-reads the now-larger list and appends again. Reuse + the FSM's
 // exact-ts dedup probe close that. See option 2 in
-// docs/design/2026_06_03_proposed_dynamodb_onephase_dedup.md.
+// docs/design/2026_06_03_partial_dynamodb_onephase_dedup.md.
 type reusableItemWrite struct {
 	// plan holds the reused OperationGroup (plan.req: Elems + fixed StartTS) and
 	// the captured current/next item. The client-visible result
@@ -1353,8 +1377,10 @@ func (d *DynamoDBServer) retryItemWriteWithGenerationDedup(
 			plan, pending, err = d.itemWriteFirstAttempt(ctx, tableName, prepare)
 		}
 		if err != nil {
+			// commitItemWrite already wraps dispatch errors; the attempt helpers
+			// return them raw, so return raw here too (no double WithStack).
 			if !isRetryableTransactWriteError(err) {
-				return nil, errors.WithStack(err)
+				return nil, err
 			}
 		} else if plan != nil {
 			return plan, nil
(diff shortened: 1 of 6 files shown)


bootjp:
@claude review

bootjp:
## Round-1 review fixes pushed (force-push: rebased onto main)

Addressed all findings; also **removed 3 unrelated encryption Stage 6E-2b commits** that had leaked in from a local-ahead `main` (rebased `--onto origin/main`). The PR is now 3 commits / 5 files, all DynamoDB-dedup-scoped.

### codex P1 — leader-issued commit timestamps (FIXED)
The dedup path allocates `commitTS` from the **local** HLC, so it is only safe on the leader. Gated the dedup path on `d.coordinator.IsLeader()`; a non-leader falls back to the legacy path where `Coordinator.Dispatch` redirects to the leader and the **leader** allocates `commitTS`. On the leader the single HLC issues unique monotonic values, and `NextFenced`'s physical-ceiling fence keeps a deposed leader's window disjoint from its successor's (so a TOCTOU leadership loss after the check still cannot collide). In the normal multi-node deployment the adapter already HTTP-proxies follower ingress to the leader, so the dedup path runs on the leader anyway; the guard only matters for the no-`leaderMap` topology. New test: `TestItemWriteDedup_NonLeaderFallsBackToLegacy`.

### claude (substantive) — probe error silently dropped in `resolveReuseWriteConflict` (FIXED)
Now **fails closed**: when `CommittedVersionAt` returns an error, surface it (`errors.Wrap`) instead of recomputing, so a transient read error cannot trigger a blind recompute that double-appends. Matches the FSM-side `dedupProbeOnePhase`.

### gemini (medium ×4) — redundant `errors.WithStack` (FIXED)
The attempt helpers return the already-wrapped dispatch error raw (`commitItemWrite` wraps once); no double `WithStack`. Retry classification is unaffected (`errors.Is` is wrap-transparent).

### claude (minor) — `ErrTxnLocked`-on-reuse untested (FIXED)
Added `TestItemWriteDedup_TxnLockedOnReuseAdvancesProbe`, exercising the `pending.commitTS = commitTS` advance on the ambiguous-retryable branch.

### claude (minor) — design-doc lifecycle (FIXED)
`git mv` proposed → `2026_06_03_partial_dynamodb_onephase_dedup.md`, `Status: Partial` (M1 ships here; M2 validation + S3/SQS remain open). Updated all code comment references.

### Refactor
Extracted the legacy loop into `retryItemWriteWithGenerationLegacy` to keep `retryItemWriteWithGeneration` under the cyclop limit after the leader guard.

Caller audit (semantic-change rule): the error-handling change (probe fail-closed + raw error return) flows to `writeDynamoErrorFromErr`, which classifies via `errors.As` (wrap-transparent) → a probe error surfaces as a 500 (client retries), never a double-apply. The 3 callers (`updateItemWithRetry` / `putItemWithRetry` / `deleteItemWithRetry`) are unaffected (gate-off ⇒ byte-identical legacy path).

Validation: `go test ./adapter/ -run 'Dedup|UpdateItem|PutItem|DeleteItem'`, `./kv`, `./store` pass; `golangci-lint ./adapter/... ./kv/...` 0 issues; `go build ./...` clean.

@claude review

bootjp:
## Round-2 review fix pushed

### coderabbit (🟠 Major) — generation fence on the landed reuse path (FIXED)
`resolveReuseWriteConflict` returned `pending.plan` directly when the self-inflicted-conflict probe found the reuse landed, skipping the table-generation fence + cleanup that the normal success path runs. Now routes the landed path through `finishItemWriteAttempt` (threaded `tableName` in), so a table dropped/recreated under us cleans up the landed write and recomputes — matching the non-reuse success path. Covered transitively by the shared `finishItemWriteAttempt` (exercised by the item-write + self-conflict tests).

### Already-resolved / moot
- **coderabbit (probe failures treated as misses)** — fixed in round-1 (fail-closed); comment is now outdated.
- **coderabbit (stale "no implementation yet" banner)** — fixed in round-1 (doc renamed to `*_partial_*`, banner updated); outdated.
- **gemini (errors.WithStack ×4)** — removed in round-1; the inline comments re-anchored to new line numbers after the force-push but the `WithStack` is already gone (helpers return the raw, once-wrapped dispatch error).
- **coderabbit nitpick (`adapter/encryption_admin_test.go` ProposeAdmin observability)** — **moot**: those were unrelated encryption Stage 6E-2b commits that leaked in from a local-ahead `main`; rebased out in round-1, so that file is no longer in this PR.

Validation: `go test ./adapter/ -run 'TestItemWriteDedup'` (7/7) pass; `golangci-lint ./adapter/...` 0 issues; `go build ./...` clean.

@claude review

PR #921, opened 2026-06-03
ci(jepsen): add DynamoDB workload to the dedup-mode scheduled workflow (M2)

bootjp:
## Summary

M2 of [`docs/design/2026_06_03_partial_dynamodb_onephase_dedup.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_06_03_partial_dynamodb_onephase_dedup.md): extend the option-2 **dedup-mode** Jepsen workflow (`jepsen-test-scheduled-dedup.yml`) to also exercise the DynamoDB single-item write path under the gate that shipped in PR #920. This starts the 7-day-green validation clock for the DynamoDB dedup path.

## Changes (`.github/workflows/jepsen-test-scheduled-dedup.yml`)

- **`ELASTICKV_DYNAMODB_ONEPHASE_DEDUP: "1"`** added at the job env. The demo cluster reads it via `adapter.NewDynamoDBServer` (`cmd/server/demo.go` → `setupDynamo`), so the single-item write path routes through `retryItemWriteWithGenerationDedup` on the leader.
- **Launch step**: now waits on the dynamo listeners (63801-63803) in addition to redis, and **fail-closed-asserts** the DynamoDB gate (`exit 2` if not `"1"`), mirroring the existing Redis assertion.
- **New step** `Run DynamoDB Jepsen workload (dedup mode)` — `lein run -m elastickv.dynamodb-workload --local --dynamo-ports 63801,63802,63803` against the gated cluster.
- Store artifact renamed `jepsen-store-redis-dedup` → `jepsen-store-dedup` (now covers both workloads); header/scope comments refreshed.

## What this does NOT do (by design)

- **Does not flip default-on.** Per the design's ship-reader-before-writer sequencing (R5), `DynamoDBServer.onePhaseTxnDedup` stays default-off; default-on is a separate follow-up PR once the 7-consecutive-days-green criterion is met. This PR only starts the clock.
- **Does not touch S3/SQS** — they don't route through the dedup loop yet (separate follow-up).
- The general 6-hourly workflow keeps running DynamoDB **gate-off**, so the legacy path stays covered.

## Validation

- `python3 -c "import yaml; yaml.safe_load(...)"` parses clean — 15 steps, env has all three keys, both Redis + DynamoDB workload steps present.
- Demo-cluster wiring verified: `setupDynamo` → `NewDynamoDBServer` reads `os.Getenv("ELASTICKV_DYNAMODB_ONEPHASE_DEDUP")`.
- No Go code change; `make lint` clean.

## Self-review (5 passes)

1. **Data loss** — n/a (CI-only change; no storage/replication code touched).
2. **Concurrency / distributed** — n/a; the workload exercises the existing gated path. The fail-closed gate assertions prevent a silently-misconfigured run from reporting a vacuous green.
3. **Performance** — adds one ~5-min DynamoDB workload to a daily off-peak job; within the job timeout.
4. **Data consistency** — the workflow is the *validator* for the consistency property (no `:duplicate-elements` under the gate); it changes no semantics.
5. **Test coverage** — this PR *is* the integration coverage for the DynamoDB dedup path under leader churn.

### DIFF
--- a/.github/workflows/jepsen-test-scheduled-dedup.yml
+++ b/.github/workflows/jepsen-test-scheduled-dedup.yml
@@ -1,16 +1,19 @@
 # Jepsen Scheduled Stress Test — Option-2 Dedup Mode
 #
-# Daily run with ELASTICKV_REDIS_ONEPHASE_DEDUP=1 so the demo cluster
-# exercises the option-2 idempotency path. The criterion for design doc
-# §M4 closure is 7 consecutive days of green runs with no
-# :duplicate-elements / :G-single-item-realtime anomalies in the Redis
-# workload's analysis output.
+# Daily run with ELASTICKV_REDIS_ONEPHASE_DEDUP=1 and
+# ELASTICKV_DYNAMODB_ONEPHASE_DEDUP=1 so the demo cluster exercises the
+# option-2 idempotency path on both adapters. The criterion for
+# default-on (docs/design/2026_05_21 §M4 and
+# docs/design/2026_06_03_partial_dynamodb_onephase_dedup.md §M2) is 7
+# consecutive days of green runs with no :duplicate-elements /
+# :G-single-item-realtime anomalies in each workload's analysis output.
 #
-# Scope: Redis workload only. The dedup feature only ships behind the
+# Scope: Redis + DynamoDB workloads. The dedup feature ships behind the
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

PR #922, opened 2026-06-03
feat(encryption): Stage 6E-2c — dynamic raft envelope wrap on ShardGroup

bootjp:
## Summary

Stage 6E-2c: install the Stage 6E-2 wrap-on-propose plumbing on every shard group without yet activating any cipher. Every `ShardGroup` gets an `atomic.Pointer[RaftPayloadWrapper]` cell; the proposer chain consults that pointer on every `Propose` / `ProposeAdmin` call; `SetRaftPayloadWrap` is the hot-swap surface Stage 6E-2d will use when `EnableRaftEnvelope` commits. The wrap closure factory itself ships in Stage 6E-2e/2f.

A startup hook reads the sidecar and logs a `WARN` if `RaftEnvelopeCutoverIndex != 0` — operator-visible refusal of a deployment that crosses the cutover line under a binary without the cipher factory wired.

## Pieces

- `dynamicWrappedProposer` (kv/raft_payload_wrapper.go): wraps a `raftengine.Proposer` with an `atomic.Pointer.Load` on every call. Mirrors `wrappedProposer`'s Propose / ProposeAdmin routing and wrap semantics — the wrap layer is NOT a barrier exemption; Propose and ProposeAdmin apply the same wrap. The cutover marker stays cleartext via the raw-engine reference at the call site, not at the method level.
- `ShardGroup.raftPayloadWrap` + `SetRaftPayloadWrap` + `RaftPayloadWrap` (kv/sharded_coordinator.go): the published hot-swap surface.
- `NewLeaderProxyForShardGroup` + `withDynamicRaftPayloadWrap` (kv/sharded_coordinator.go, kv/transaction.go): the production wiring path. `main.go`'s `buildShardGroups` now constructs every shard group's `LeaderProxy` via this helper so the dynamic wrap chain is universal.
- `noteRaftEnvelopeCutoverStartup` (main_encryption_write_wiring.go): process-start sidecar read. Silent on the universal pre-cutover case and on any read error (the Stage 6C-1 startup guard is the authoritative refuser). Logs `WARN` only when the cutover index is non-zero but the binary does not (yet) ship the cipher factory.

## Tests (8 new)

- `TestDynamicWrappedProposer_NilPointerReturnsInnerVerbatim` — defensive degradation for nil `*atomic.Pointer`.
- `TestDynamicWrappedProposer_NilStoredIsPassThrough` — Stage 3 default (verbatim routing for Propose and ProposeAdmin separately).
- `TestDynamicWrappedProposer_LoadsCurrentWrapEveryCall` — hot-swap contract (install / clear observed by the next proposal).
- `TestDynamicWrappedProposer_ProposeAdminAppliesCurrentWrap` — admin-path mirror; confirms `inner.ProposeAdmin` (not `Propose`) is reached when wrap is active.
- `TestShardGroup_SetRaftPayloadWrap_RoundTrip` — Set/Get round-trip including the nil-clears semantics.
- `TestNoteRaftEnvelopeCutoverStartup_*` (×4) — empty path / missing file / zero cutover / non-zero cutover (asserts the hook is read-only).

## Five-lens self-review

- **Data loss**: the dynamic wrap proposer is a pass-through when the pointer holds nil (the Stage 6E-2c default). Production behaviour is unchanged today — every proposal still reaches the engine cleartext via the existing path. No new error swallow, no new fall-through.
- **Concurrency / distributed**: hot-swap uses `atomic.Pointer.Store` / `Load`; reads are lock-free and observe a fully-published closure (either nil or non-nil) on every call. Last-writer-wins on concurrent stores; no torn read.
- **Performance**: one extra `atomic.Pointer.Load` per Propose / ProposeAdmin on the dynamic path; no allocation, no contention. Wrap closure invocation cost is paid only when non-nil; matches the existing `wrappedProposer` profile.
- **Data consistency**: the wrap pointer's lifetime is owned by the `ShardGroup` that owns the engine; the proposer holds an `*atomic.Pointer` reference back into `ShardGroup`. No cross-shard mixing — each shard has its own pointer, so a future per-shard rotation policy works without changes here.
- **Test coverage**: 8 new tests cover the wrap-proposer mechanics, the ShardGroup hot-swap surface, and the startup hook input matrix.

- `kv.NewLeaderProxyWithEngine` (kept; non-wrap-aware): no behaviour change. Test fixtures and any shard group that opts out of encryption continue to use it.
- `kv.NewLeaderProxyForShardGroup` (new; wrap-aware): used by `main.go`'s `buildShardGroups` for every production shard group. Audit: the only construction site for production `ShardGroup` instances now routes through this helper.
- ShardGroup struct literal in `main.go`: switched from one-shot literal with `Txn` set inline to two-step build (struct literal, then `Txn = NewLeaderProxyForShardGroup(sg, opts...)`). Necessary because `NewLeaderProxyForShardGroup` needs the `&sg.raftPayloadWrap` pointer, which is only addressable after the struct exists.
- `SetRaftPayloadWrap` callers: none in this PR. The surface is published for Stage 6E-2d's `EnableRaftEnvelope` handler.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `golangci-lint run` clean (pre-commit hook ran)
- [x] `go test -race -run 'DynamicWrappedProposer|ShardGroup_Set|WrappedProposer|NoteRaftEnvelopeCutover' .` passes
- [x] `go test -race ./kv/` scoped tests pass

Refs: `docs/design/2026_05_31_partial_6e_enable_raft_envelope.md`

## Release Notes

* **New Features**
  * Introduced runtime-swappable raft payload envelope wrapping with dynamic wrapper support across all proposal types and shard groups.

* **Chores**
  * Added startup validation to detect and prevent incompatible payload envelope cutover states with clear operator guidance.

* **Tests**
  * Expanded test coverage for dynamic payload wrapper behavior, atomic pointer runtime swapping, integration with shard groups, and startup safety checks.

### DIFF
--- a/kv/raft_payload_wrapper.go
+++ b/kv/raft_payload_wrapper.go
@@ -136,18 +136,32 @@ type dynamicWrappedProposer struct {
 	wrapPtr *atomic.Pointer[RaftPayloadWrapper]
 }
 
-// newDynamicWrappedProposer wires a proposer that consults wrapPtr on
-// every Propose / ProposeAdmin. wrapPtr.Load() == nil keeps the path
-// as a pure pass-through to inner; a non-nil value applies wrap
-// before forwarding. wrapPtr must not be nil itself — the caller
-// owns the storage lifetime.
+// newDynamicWrappedProposer wires a proposer that consults wrapPtr
+// on every Propose / ProposeAdmin. wrapPtr.Load() == nil keeps the
+// path as a pure pass-through to inner; a non-nil value applies
+// wrap before forwarding.
+//
+// Contract: inner MUST be non-nil. A nil inner is a caller bug
+// (the construction site is responsible for handing in a valid
+// proposer) and the constructor deliberately does not silently
+// degrade — passing nil to Propose / ProposeAdmin would NPE at
+// the first call site, which is the same fail-fast shape as the
+// sibling newWrappedProposer(nil, wrap) and matches CLAUDE.md's
+// "don't validate for scenarios that can't happen" policy at
+// internal boundaries.
+//
+// wrapPtr MAY be nil — that path returns inner verbatim, mirroring
+// newWrappedProposer(inner, nil)'s degraded shape so a wiring site
+// that accidentally drops the pointer downgrades to a no-wrap
+// proposer rather than crashing the engine loop on first use. The
+// caller owns the storage lifetime of the pointer when non-nil.
 func newDynamicWrappedProposer(inner raftengine.Proposer, wrapPtr *atomic.Pointer[RaftPayloadWrapper]) raftengine.Proposer {
 	if wrapPtr == nil {
-		// Defensive: passing a nil pointer would NPE on the first
-		// Propose. Treat as static no-wrap so callers that pass nil
-		// by mistake degrade safely rather than crashing the engine
-		// loop. This matches newWrappedProposer(inner, nil) returning
-		// inner verbatim.
+		// Defensive degradation: passing a nil pointer would NPE
+		// on the first Propose. Treat as static no-wrap so callers
+		// that pass nil by mistake see the Stage 3 default rather
+		// than crashing the engine loop. This matches
+		// newWrappedProposer(inner, nil) returning inner verbatim.
 		return inner
 	}
 	return &dynamicWrappedProposer{inner: inner, wrapPtr: wrapPtr}
--- a/kv/raft_payload_wrapper.go
+++ b/kv/raft_payload_wrapper.go
@@ -2,6 +2,7 @@ package kv
 
 import (
 	"context"
+	"sync/atomic"
 
 	"github.com/bootjp/elastickv/internal/raftengine"
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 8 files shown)


bootjp:
@claude review

Round-1 round-up:

- **Gemini medium ×2** (defensive nil check on `newDynamicWrappedProposer(inner)` + `NewLeaderProxyForShardGroup(g)`): rejected per CLAUDE.md's "don't validate for scenarios that can't happen at internal boundaries". Silently returning nil would only defer the panic to the first `Propose` / `Commit` call site with worse diagnostics. Documented the contract on both constructors so the rejection rationale lives in-code rather than review history (d2900262).

The `wrapPtr == nil` degradation on `newDynamicWrappedProposer` is preserved — that one matches `newWrappedProposer(inner, nil)` returning `inner` verbatim, and `TestDynamicWrappedProposer_NilPointerReturnsInnerVerbatim` pins the shape against regressions.

Head: d2900262.

bootjp:
@claude review

Round-1 round-up:

- **Codex P1** (fail closed on restored raft cutovers): `noteRaftEnvelopeCutoverStartup` was WARN-only; if a sidecar restored from a post-cutover deployment hit a binary without the wrap closure factory, fresh writes would land cleartext above the cutover and halt the apply loop. Renamed to `refuseStartupOnActiveRaftCutover` and now returns a fail-closed error so `buildShardGroupsWithEncryptionWiring` refuses startup BEFORE buildShardGroups runs. No half-built runtime set to tear down.
- **Codex P2** (direct shard proposals bypass wrapper): `ShardedCoordinator.RunHLCLeaseRenewal` called `group.Engine.Propose` directly, which would skip the dynamicWrappedProposer once SetRaftPayloadWrap was populated post-cutover. Cached the dynamicWrappedProposer on `ShardGroup.proposer` when `NewLeaderProxyForShardGroup` constructs it; published as `ShardGroup.Proposer()` with an Engine fallback for legacy/test paths. HLC renewal now routes through `group.Proposer().Propose`.
- **Gemini medium ×2** (round-0): defensive nil checks rejected per CLAUDE.md, doc-only clarification in d2900262.

Caller audit (per /loop): all production `.Engine.Propose` sites surveyed. `sharded_coordinator.go:1864` migrated to `Proposer()`. `coordinator.go:637` (Coordinate.ProposeHLCLease) is test-only, not migrated. `main_encryption_registration.go:503` is deliberately raw-engine (the cutover marker must be cleartext); the post-cutover gap there is already tracked alongside the 6E-2c/2e wiring per the call-site comment.

Head: 9a33e678.

bootjp:
@claude review

Round-1 r3 round-up:

- **Codex P1** (forwarded internal writes bypass wrapper): the leader's Internal.Forward gRPC service builds its own TransactionManager via `kv.NewTransactionWithProposer(rt.engine, ...)`, which is the receive side of `LeaderProxy.forwardWithRetry`. Round-1's fix made LOCAL leader writes wrap-aware via NewLeaderProxyForShardGroup, but the forwarded-write path stayed on the raw engine — post-cutover writes from non-leader clients would land cleartext above the cutover. Threaded `shardGroups` through `serversInput → runtimeServerRunner → startRaftServers` and added `proposerForGroup` helper so the Internal trx now routes through the wrap-aware ShardGroup proposer. 83e7b3f8.

Caller audit (per /loop): every production `.Engine.Propose` site is now either (a) wrap-aware (LeaderProxy.Commit, Internal.Forward, RunHLCLeaseRenewal) or (b) explicitly raw-engine with a documented reason (cutover marker / RegisterEncryptionWriter, where the post-cutover gap is tracked alongside 6E-2c/2e).

Head: 83e7b3f8.

bootjp:
@claude review

Round-1 r4 round-up:

- **Codex P2 r3** (sidecar epoch bumped before refusal): the round-1 r2 ordering put `refuseStartupOnActiveRaftCutover` AFTER `buildEncryptionWriteWiring`, which internally calls `prepareStorageNonceEpoch → BumpLocalEpoch → durable sidecar write`. An orchestrator restart loop on a sidecar with active storage DEK + non-zero RaftEnvelopeCutoverIndex would consume one local_epoch per boot and eventually hit `ErrLocalEpochExhausted` — turning a recoverable misconfiguration into a non-recoverable DEK-rotation event. Moved the refusal to run BEFORE `buildEncryptionWriteWiring`. ad8b2147.
- Added `TestRefuseStartupOnActiveRaftCutover_NonZeroCutoverDoesNotBumpEpoch` to pin the ordering invariant (sidecar LocalEpoch is unchanged after refusal).

Note re Claude round-2 review (4615200467): the "Internal.Forward bypass" item it flagged on commit 9a33e678 was already addressed in 83e7b3f8 (the r3 commit Claude hadn't seen yet) via `proposerForGroup` helper + `shardGroups` plumbing through `startRaftServers`. All audit findings should now be resolved.

Head: ad8b2147.

bootjp:
Force-push: rebased onto origin/main to drop an unrelated commit (`5846c88e ci(jepsen): add DynamoDB workload to the dedup-mode scheduled workflow`) that had leaked into this branch from a local-ahead main. The 5 Stage 6E-2c commits are preserved; only their SHAs changed.

New head: 9fe4125f.

This makes the Codex P2 r4 finding on `.github/workflows/jepsen-test-scheduled-dedup.yml:209` ("keep the DynamoDB dedup workload single-item") no longer applicable to this PR — that file is no longer part of the change set. The finding still applies to the upstream jepsen workflow on main and should be tracked there.

@claude review

bootjp:
@claude review

Round-1 r5 round-up:

- **Coderabbit nitpick** (main.go:1573 — Internal.Forward proposer selection unverified): added focused unit-test file `main_proposer_for_group_test.go` pinning all three branches of `proposerForGroup` — wrap-aware via `sg.Proposer()`, fallback to `rt.engine` on missing entry, fallback on explicit nil entry.
- **Coderabbit nitpick** (main_encryption_write_wiring_test.go:215 — non-ENOENT read branch uncovered): added `TestRefuseStartupOnActiveRaftCutover_BadSidecarPropagates` with a malformed (non-JSON) sidecar to pin the fail-closed propagation contract.

Both are defense-in-depth regression tests; no production code changed. Head: 47bb81a8.

bootjp:
@claude review please verify TLA+ spec divergence per the checklist above.

PR #923, opened 2026-06-03
fix(adapter/sqs): standard SendMessageBatch double-send under leader churn

bootjp:
## Summary

Fixes a **double-send** bug in the SQS standard-queue `SendMessageBatch` path — the same `:duplicate-elements` class fixed for DynamoDB in PR #920, found by an adversarial S3/SQS audit prompted by the "extend dedup to S3/SQS" follow-up.

## The bug

`sendMessageBatchWithRetry` → `trySendMessageBatchOnce` → `sendBatchStandardOnce` runs an **in-process OCC retry loop** that, on every attempt, re-ran `buildBatchSendRecord` → `buildSendRecord` → `newMessageIDHex()`, minting a **fresh random MessageID (and send timestamp) per entry per attempt**, and derived the `data`/`vis`/`byage` storage keys from them.

Under a Raft leader-election storm, attempt 1 can **commit** its writes yet surface a self-inflicted `WriteConflict` to the adapter (the mechanism PR #920 documented). The retry then mints **new** MessageIDs and writes a **second copy of every entry** at new keys:

1. Attempt 1: mints `{M1,M2,M3}`, commits, but returns `WriteConflict` under churn.
2. Retry: mints `{M1',M2',M3'}`, commits cleanly, returns `{M1',M2',M3'}` + HTTP 200.
3. Queue holds **both** sets → every batch entry duplicated; the client only learns the second set.

Standard queues carry no dedup identity, and the OCC `ReadKeys` are only `[meta, gen]` (the random data keys never self-conflict), so nothing fenced this.

**Why only batch:** single `SendMessage` has **no** in-process retry (a conflict becomes a normal at-least-once SDK retry), and FIFO send is fenced by the `MessageDeduplicationId` dedup record. The batch path added an in-process retry without either fence.

## The fix

Pre-generate **one stable `sqsSendIdentity`** (MessageID + receipt token + send timestamp) **per entry, before the retry loop**, and reuse it on every attempt (`buildSendRecordWithIdentity`). The storage keys then stay constant across retries, so a committed-but-conflicted attempt is **overwritten idempotently** instead of duplicated.

No FSM probe / `PrevCommitTS` / gate is needed — unlike DynamoDB's `list_append` (which re-read a *growing* value), the batch write is a whole-value PUT at a now-stable key, so re-applying is a plain idempotent overwrite. The record's `QueueGeneration` is still re-derived per attempt, so a concurrent `DeleteQueue`/`PurgeQueue` generation bump is followed correctly (the stale-gen attempt-1 write is orphaned/unreachable).

## S3/SQS audit outcome (recorded in the design doc)

A handler-by-handler adversarial sweep found the original "S3/SQS share the same latent gap" note was over-broad:
- **S3 — safe.** Whole-value PUT at a stable key; `PutObject` is OCC-fenced by a caller-supplied `StartTS` (a leader-move retry that races an interleaved write **fails the fence rather than clobbering**), and has no in-process re-read-recompute. `CompleteMultipartUpload`'s manifest is re-assembled deterministically from the **immutable** uploaded parts (computed once before its retry loop), not from a re-read of the object value → no duplicate.
- **SQS single `SendMessage` / FIFO — safe** (random-UUID key + no in-process retry; FIFO dedup record).
- **SQS standard `SendMessageBatch` — fixed here.**

## Tests (`adapter/sqs_batch_send_dedup_test.go`)

- `TestSendMessageBatchStandard_RetryReusesStableKeys` — first dispatch commits-then-conflicts; asserts the retry dispatches the **same** storage keys (fresh MessageIDs would fail it → reproduces the bug pre-fix).
- `TestSendMessageBatchStandard_ReturnedMessageIdsStableAcrossRetry` — the returned MessageId is the one actually written on both attempts (client id == stored id).

## Validation

- `go test ./adapter/ -run 'SendMessage|Batch|FIFO|Send'` pass; new dedup tests pass.
- `golangci-lint run ./adapter/...` 0 issues; `go build ./...` clean.

## Self-review (5 passes)

1. **Data loss** — the fix prevents a double-send (extra messages), and keeps the generation re-derivation so a concurrent queue delete/recreate still routes to the live generation; no committed message is dropped.
2. **Concurrency / distributed** — stable keys make the leader-churn retry an idempotent overwrite; the `[meta,gen]` ReadKeys fence against `DeleteQueue`/`PurgeQueue` is unchanged.
3. **Performance** — identities are minted once instead of once-per-attempt (strictly fewer `crypto/rand` calls on the retry path); hot path (no retry) is one mint per entry, same as before.
4. **Data consistency** — eliminates the duplicate-element anomaly for standard batch send; FIFO/exactly-once and single-send semantics unchanged.
5. **Test coverage** — new co-located tests pin key-stability across the retry (the regression), the load-bearing property of the fix.

### DIFF
--- a/adapter/sqs_batch_send_dedup_test.go
+++ b/adapter/sqs_batch_send_dedup_test.go
@@ -18,9 +18,15 @@ import (
 type recordingBatchCoordinator struct {
 	stubAdapterCoordinator
 	dispatchKeys [][]string
+	// beforeDispatch, if set, runs at the start of each Dispatch with the
+	// 1-based dispatch number — lets a test mutate queue state between attempts.
+	beforeDispatch func(n int)
 }
 
 func (c *recordingBatchCoordinator) Dispatch(_ context.Context, req *kv.OperationGroup[kv.OP]) (*kv.CoordinateResponse, error) {
+	if c.beforeDispatch != nil {
+		c.beforeDispatch(len(c.dispatchKeys) + 1)
+	}
 	keys := make([]string, 0, len(req.Elems))
 	for _, e := range req.Elems {
 		keys = append(keys, string(e.Key))
@@ -74,6 +80,41 @@ func TestSendMessageBatchStandard_RetryReusesStableKeys(t *testing.T) {
 	require.NotEmpty(t, coord.dispatchKeys[0])
 	require.Equal(t, coord.dispatchKeys[0], coord.dispatchKeys[1],
 		"the retry MUST reuse the same storage keys; fresh MessageIDs would double-send every entry under leader churn")
+}
+
+// TestSendMessageBatchStandard_VisKeyStableAcrossDelayChange pins codex P2:
+// if a SetQueueAttributes changes the queue's DelaySeconds between a
+// committed-but-conflicted attempt and its retry, the retry must still write
+// the SAME vis/by-age keys (AvailableAtMillis is captured in the identity at
+// first mint, not recomputed from the changed delay) — otherwise the first
+// attempt's vis index entry is orphaned and can redeliver the message.
+func TestSendMessageBatchStandard_VisKeyStableAcrossDelayChange(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	st := store.NewMVCCStore()
+	seedStandardQueue(t, st, "q") // DelaySeconds = 0
+
+	coord := &recordingBatchCoordinator{stubAdapterCoordinator: stubAdapterCoordinator{clock: kv.NewHLC()}}
+	coord.beforeDispatch = func(n int) {
+		if n != 2 {
+			return
+		}
+		// SetQueueAttributes raises DelaySeconds to 900 before the retry.
+		meta := &sqsQueueMeta{Name: "q", Generation: 1, MaximumMessageSize: 262144, DelaySeconds: 900}
+		body, err := encodeSQSQueueMeta(meta)
+		require.NoError(t, err)
+		require.NoError(t, st.PutAt(ctx, sqsQueueMetaKey("q"), body, 100, 0))
+	}
+	srv := &SQSServer{store: st, coordinator: coord}
+
+	entries := []sqsSendMessageBatchEntryInput{{Id: "a", MessageBody: "alpha"}} // no per-message DelaySeconds
+	successful, _, err := srv.sendMessageBatchWithRetry(ctx, "q", entries)
+	require.NoError(t, err)
+	require.Len(t, successful, 1)
+
+	require.Len(t, coord.dispatchKeys, 2)
+	require.Equal(t, coord.dispatchKeys[0], coord.dispatchKeys[1],
+		"vis/by-age keys must stay stable across a mid-retry DelaySeconds change (cached AvailableAtMillis)")
 }
 
 // TestSendMessageBatchStandard_ReturnedMessageIdsStableAcrossRetry pins that
--- a/adapter/sqs_batch_send_dedup_test.go
+++ b/adapter/sqs_batch_send_dedup_test.go
@@ -1,0 +1,103 @@
+package adapter
+
+import (
+	"context"
+	"testing"
+
+	"github.com/bootjp/elastickv/kv"
+	"github.com/bootjp/elastickv/store"
+	"github.com/stretchr/testify/require"
+)
+
+// recordingBatchCoordinator records the mutation keys of each dispatch and
+// fails the FIRST dispatch with a retryable WriteConflict — modelling a
+// committed-but-conflicted attempt under leader churn (the apply may have
+// landed, but the dispatch surfaced WriteConflict). The second dispatch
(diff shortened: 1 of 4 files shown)


bootjp:
@claude review

bootjp:
## Round-1 review fixes pushed

### gemini (medium) — wasted identity minting on FIFO batches (FIXED)
Identities are now minted **lazily** on the first standard-path attempt (into a zero-valued, by-reference slice reused across retries) instead of all upfront. FIFO batches — which mint their own per-entry dedup-fenced identity — pay zero crypto/rand cost for identities they ignore.

### codex (P2) — vis key not stable across a mid-retry DelaySeconds change (FIXED)
`AvailableAtMillis` (which the vis/by-age keys embed) is now **captured in the identity at first mint** and reused, instead of being recomputed from a re-resolved delay each attempt. So a `SetQueueAttributes` that changes `DelaySeconds` between a committed-but-conflicted attempt and its retry no longer shifts the vis key / orphans the first attempt's vis index entry. Validation + delay resolution also run only on first mint, so an already-committed entry is never re-failed against a since-changed meta.

`buildSendRecordWithIdentity` drops its `delay` param (AvailableAtMillis lives in the identity); `buildSendRecord`'s external signature is unchanged → single-send and FIFO callers unaffected (audited: only `buildSendRecord` callers are sqs_messages.go:515 and sqs_fifo.go:208, both pass the same `(meta,in,delay)`).

New test: `TestSendMessageBatchStandard_VisKeyStableAcrossDelayChange` (a mid-retry DelaySeconds bump must not move the keys). All batch/send/FIFO tests pass; `golangci-lint ./adapter/...` 0 issues.

@claude review

bootjp:
## Round-2 review fix pushed

### codex (P2 round-2) — re-validate batch entries against current meta on every attempt (FIXED)
Round-1 moved per-entry validation into the first-mint path, so an entry whose identity was minted by a prior **uncommitted** attempt (lost a normal OCC race) skipped validation on retry — if a concurrent `SetQueueAttributes` lowered `MaximumMessageSize` meanwhile, the retry could commit a message the fresh meta rejects.

Split the two concerns:
- **`validateBatchEntry` runs against the CURRENT meta on EVERY attempt** (body required, MaximumMessageSize, attributes, FIFO params, delay bounds).
- The **stable identity is still minted once** and reused, so keys stay stable across retries (round-1 vis-key stability preserved). `AvailableAtMillis` is seeded from the first attempt's delay; re-resolving delay per attempt only re-checks bounds, it never shifts the key.

Caller audit: `validateBatchEntry` ← `buildBatchSendRecord` ← `sendBatchStandardOnce` (single callers each); `mintBatchSendIdentity` removed (no refs); single-send/FIFO unaffected (`buildSendRecord` unchanged).

New test: `TestSendMessageBatchStandard_RevalidatesAgainstCurrentMetaOnRetry` — a mid-flight `MaximumMessageSize` drop makes the retry reject the now-too-big entry. All batch/send/FIFO tests pass; `golangci-lint ./adapter/...` 0 issues.

@claude review

bootjp:
## Round-3 (docs only) pushed — addresses the two non-blocking nits from the round-2 approval

claude[bot] **approved** round-2; these close the remaining observations:
- **Stale doc comment removed**: dropped the obsolete first paragraph of `buildBatchSendRecord`'s doc (it referenced a `*sqsAPIError` return that no longer exists and duplicated the accurate description below).
- **Residual edge documented**: added a note to the design doc — if attempt 1 *commits* and a concurrent `SetQueueAttributes` tightens a limit before the retry, the retry re-validates and reports the (already-stored) entry as `failed` — an inconsistent client view within the standard-queue at-least-once contract, never a double-send.

No functional change (comment + markdown only); `go build` + `golangci-lint ./adapter/...` clean.

@claude review

bootjp:
## codex P2 ×2 → deferred to #931 (decision: merge this fix, follow up on the full dedup)

Both new codex P2 findings on `65fe7d0d` stem from the same root: the simple stable-key reuse assumes attempt 1's keys are untouched between the committed-but-conflicted attempt and the retry, which a concurrent **consumer** (receive/delete) or **SetQueueAttributes** can violate. Closing them correctly requires the full **option-2 dedup** (local `commitTS` + leader-only + `PrevCommitTS` exact-ts probe + data-key fence + result cache) — the same gated machinery used for DynamoDB in #920, i.e. a substantially larger change than this double-send correctness fix.

Per maintainer decision, this PR ships the double-send fix (the Jepsen-caught `:duplicate-elements`, which it fully eliminates and which claude approved), and the two P2 edges are tracked in **#931** for the gated dedup follow-up:
- **P2-1** (retry clobbers a concurrent consumer receive/delete) — narrower window than the original double-send, one message vs every entry; real but deferred.
- **P2-2** (committed entry reported `Failed[]` after retry revalidation) — within the standard-queue at-least-once contract; already documented in the design doc.

Net: strict improvement over `main` (eliminates the severe, common double-send) with the residual narrow edges tracked. Enabling auto-merge (squash) once required checks pass.

PR #933, opened 2026-06-05
feat(encryption): Stage 6E-2d — §7.1 quiescence barrier scaffold

bootjp:
## Summary

Stage 6E-2d ships the §7.1 quiescence-barrier scaffold inside
`EnableRaftEnvelope`. The barrier mechanism is fully wired and
tested but unreachable from operators — `raftEnvelopeWrapEnabled`
stays `false` in this slice (6E-2e wires the closure source from
`main.go`; 6E-2f atomically flips the gate).

### What ships

- `raftengine.ErrEnvelopeCutoverInProgress` sentinel error
  returned on `Propose` under the barrier. `ProposeAdmin` is
  exempt by interface contract so the cutover marker itself (and
  ConfChange-time `RegisterEncryptionWriter`) cross the barrier.
- `dynamicWrappedProposer` barrier state (`barrierMu` +
  `inflightUser` counter + `drainSig` channel) gating user-
  `Propose` only.
- `BeginCutoverBarrier` / `WaitInflightDrained` /
  `EndCutoverBarrier` trio on `dynamicWrappedProposer` driving
  steps 1 / 2 / 6 of the §7.1 sequence.
- `ShardGroup` forwarder methods with degraded fast-path on raw-
  Engine fallback so test fixtures degrade gracefully (immediate-
  success drain, pre-closed Begin channel, no-op End).
- `CutoverBarrierController` interface +
  `WithEncryptionAdminCutoverBarrier` option on
  `EncryptionAdminServer` (no production impl yet — 6E-2e).
- `EnableRaftEnvelope` handler runs the §7.1 6-step sequence
  (Begin → WaitDrained → ProposeAdmin → awaitCutoverApply →
  InstallWrap → End-via-defer) when the gate is true. Refuses
  with `FailedPrecondition` when the controller or latest-
  applied-index callbacks are unwired so a misconfigured 6E-2f
  release fails closed before any side effect.
- `raftEnvelopeWrapEnabled` converts from `const` to `var` so
  unit tests can flip it via `t.Cleanup`-based override (production
  value unchanged at `false`; 6E-2f atomically flips it).

### Behavior change, risk, test evidence

- **Behavior change**: production-observable change is **none**.
  `raftEnvelopeWrapEnabled = false` short-circuits before any
  6E-2d code path runs. The new error / barrier / state-machine
  are dead code under operator use until 6E-2e/2f land.
- **Risk**: low. The barrier mechanism is exercised end-to-end
  in unit tests but never engaged from production wiring.
- **Test evidence**:
  - `go test -race ./adapter/... ./kv/... ./internal/raftengine/...`
    targeting the encryption test surface — all pass.
  - `golangci-lint --config=.golangci.yaml run` (pre-commit hook
    fired during `git commit`) — 0 issues.
  - New tests pin: barrier blocks `Propose` / allows
    `ProposeAdmin`; `End` reopens `Propose`; `WaitDrained`
    respects ctx; drain semantics under in-flight Propose; the
    `ShardGroup` forwarder degraded fallback + production
    delegation; the state machine refuses without barrier/applied-
    index, drives the expected sequence, and runs `End` even on
    propose error or drain timeout.

### Caller audit

- `ErrEnvelopeCutoverInProgress` is a new sentinel — no existing
  `errors.Is` callers. The documented future caller
  (`main_encryption_registration.go::RegisterEncryptionWriter`)
  routes through `ProposeAdmin` and is barrier-exempt as designed
  (comment at the call site already anticipated this).
- `raftEnvelopeWrapEnabled` const → var: production read sites
  unchanged (one read in `EnableRaftEnvelope`); no callers
  performed compile-time reasoning on the value. Tests
  serialize their writes via `t.Cleanup`-based restore so the
  parallel `TestEncryptionAdmin_EnableRaftEnvelope_GatedUntil6E2`
  (which reads the var) observes the original `false` value
  while paused at `t.Parallel()`.

### 5-lens self-review

1. **Data loss** — none. `ErrEnvelopeCutoverInProgress` returns
   BEFORE `engine.Propose` so the entry was never accepted.
2. **Concurrency / distributed failures** — `barrierMu`
   serializes (read-barrier, inc-counter) atomically with
   (open-barrier, sample-counter); see the
   `dynamicWrappedProposer` type comment for the race that
   motivates a mutex over three atomics. Leader change mid-
   cutover: the handler's `requireLeader` precheck guards the
   entry; `ProposeAdmin` from a former leader returns
   `NotLeader` / `LeadershipLost`, the deferred `End` fires, the
   barrier closes, and the operator's retry restarts the
   sequence on the new leader.
3. **Performance** — one `sync.Mutex` acquire+release per
   `Propose` call (~50 ns), negligible vs Raft consensus
   latency. No new metrics, no new Raft round-trips, no new
   Pebble reads.
4. **Data consistency** — MVCC/OCC/HLC visibility unchanged.
   HLC lease renewal (which calls `Propose`, not `ProposeAdmin`)
   does observe the barrier; a single missed renewal tick
   during the few-ms barrier window is within the existing
   renewal tolerance (`hlcRenewalInterval = 1s`,
   `hlcPhysicalWindowMs = 3s`). The `awaitCutoverApply` step-4
   guarantees the cutover entry has been applied locally before
   step-5 `InstallWrap` runs, closing the propose-side race
   where the wrap closure could be installed before the engine
   sees the cutover index.
5. **Test coverage** — six new tests in `kv/` covering the
   barrier semantics; five new tests in `adapter/` covering the
   state machine. The barrier-incapable fallback path is
   covered alongside the production delegation path so a
   regression that swapped one for the other surfaces
   immediately.

## Test plan

- [x] `go test -race` on `adapter`, `kv`, `internal/raftengine`
      (targeted encryption surface)
- [x] `golangci-lint run` clean
- [ ] Full `make test` in CI
- [ ] No Jepsen suite — barrier is unreachable from production
      paths until 6E-2f flips the gate. Jepsen coverage will
      attach to 6E-2f / Stage 9.


bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

bootjp:
@claude review

PR #934, opened 2026-06-05
feat(snapshot-skip B3): cold-start restoreSnapshotState skip gate + CRC verifier + metrics

bootjp:
## Summary

Implements **Branch 3** of the cold-start snapshot-restore skip optimisation designed in PR #910, building on PR #915 (B2 — `metaAppliedIndex` plumbing). This is the **user-visible perf win** of the series: after this lands, the engine restores the FSM only when the on-disk state is actually behind the snapshot pointer; the common case (local restart with a fresh fsm.db) collapses to a header read + CRC verification, skipping the multi-GiB body restore that PR #909's `HEALTH_TIMEOUT_SECONDS=300` band-aid was sized to absorb.

## Reading order (6 commits, designed for one-at-a-time review)

| # | commit | scope |
|---|---|---|
| 1 | `8312f4ba` | `raftengine.SnapshotHeaderApplier` two-phase interface (`ParseSnapshotHeader` + `ApplySnapshotHeader`) |
| 2 | `024eb43c` | `kvFSM` implements both phases + compile-time guard |
| 3 | `933dab86` | `applyHeaderStateOnSkip` (3-step CRC, mirrors `openAndRestoreFSMSnapshot`) + `restoreSnapshotState` skip gate |
| 4 | (folded into #3) | `fsmAlreadyAtIndex` + skip-gate wiring |
| 5 | (next) | metrics + INFO log: `raftengine.ColdStartObserver` + `monitoring.ColdStartMetrics` + `Registry.ColdStartObserver()` |
| 6 | `ce277e66` | 9 new tests (skip-gate behaviour ×5 + CRC failure modes ×3 + kvFSM contract ×1) |

(Commit 5 was merged into the metrics commit; the actual git log is 5 commits.)

## What this does

`restoreSnapshotState` branches on `decideSkipOutcome`:

```go
decision, have := decideSkipOutcome(fsm, tok.Index)
reportColdStart(obs, logger, decision, tok.Index, have)
if decision == coldStartSkip {
    return applyHeaderStateOnSkip(fsm, fsmSnapPath(fsmSnapDir, tok.Index), tok.CRC32C)
}
return openAndRestoreFSMSnapshot(...) // unchanged
```

`decideSkipOutcome` returns one of 5 outcomes:

| outcome | trigger | action |
|---|---|---|
| `skip` | `LastAppliedIndex >= snap.Index` | header-only path |
| `execute` | `LastAppliedIndex < snap.Index` | full restore |
| `fallback_not_reader` | FSM doesn't implement `AppliedIndexReader` | full restore |
| `fallback_missing_meta` | meta key absent (pre-upgrade fsm.db) | full restore |
| `fallback_read_err` | `LastAppliedIndex` returned an error | full restore |

The three fallback paths preserve the strictly-additive guarantee from the design doc §4: over-restoring is always strictly safer than skipping incorrectly. Errors from `LastAppliedIndex` **do not abort cold start** — they collapse to fallback.

## CRC verification (mirrors `openAndRestoreFSMSnapshot`)

`applyHeaderStateOnSkip` runs the same three-step safety contract as the full restore path. Failure of any step propagates a typed error WITHOUT mutating FSM state:

| Step | Check | Error |
|---|---|---|
| 1 | `info.Size() < fsmMinFileSize` | `ErrFSMSnapshotTooSmall` |
| 2 | `readFSMFooter` vs `tokenCRC` | `ErrFSMSnapshotTokenCRC` |
| 3 | full-body `crc32.TeeReader` vs footer | `ErrFSMSnapshotFileCRC` |

The FSM's `ApplySnapshotHeader` (HLC ceiling + Stage 8a cutover) only fires after all three pass.

## Metrics + INFO log

`raftengine.ColdStartObserver` interface lives in the parent `raftengine` package so monitoring can implement it without circular import. `monitoring.Registry` gains `ColdStartObserver()`. The engine receives it through `OpenConfig.ColdStartObserver` — nil disables metrics, the skip itself still runs.

Two Prometheus series:

```
elastickv_fsm_cold_start_restore_total{outcome,fallback_reason}
elastickv_fsm_cold_start_applied_index_gap{outcome}
```

Plus a structured zap.Info log on every cold-start with fsm_applied / snapshot_index / gap and (for fallback) the reason enum.

## Design constraints honoured

- **§4 strictly-additive fallback**: `decideSkipOutcome` collapses every uncertainty to a fallback variant; `LastAppliedIndex` errors do NOT abort cold start.
- **§5 two-phase seam (round-7)**: `ParseSnapshotHeader` reads the v1/v2 header from a caller-supplied reader and drains the rest; `ApplySnapshotHeader` is pure assignment. Splits the parse from the side-effect so the engine can verify the CRC between them.
- **§5 CRC mirroring (round-6)**: 3-step verification (size + footer-vs-tokenCRC + full-body-CRC) exactly matches `openAndRestoreFSMSnapshot`. The full-body CRC pass IS the slow part of this PR (~6s for a 6 GiB FSM at 1 GiB/s SSD read) — but still strictly cheaper than the restore it replaces, which also reads the file once AND writes a temp Pebble database via `restoreBatchLoopInto`.
- **§9 observability**: three outcomes + gap label; the design's `not_reporter` label is named `not_reader` in the actual impl (matching the round-5 PR #915 rename).
- **Non-Goals respected**: `Engine.applySnapshot` (engine.go:1641 InstallSnapshot hot path) is untouched. The TODO sentinel added in PR #915 round-6 still names B3 as the follow-up site for the peer-after-InstallSnapshot bump — that's not in scope here either.

## Test results

```text
go vet ./internal/raftengine/etcd/ ./internal/raftengine/ ./kv/ ./store/ ./monitoring/   → exit 0
golangci-lint run ... (same packages)                                                     → 0 issues
go test ./internal/raftengine/etcd/ ./kv/ ./store/ ./monitoring/ -short                   → ok ~57s total
go test ./internal/raftengine/etcd/ -run 'TestSkipGate|TestApplyHeaderStateOnSkip'        → ok 0.028s
```

## What this does NOT do

- **Does NOT change** `HEALTH_TIMEOUT_SECONDS=300`. That's Branch 4, gated on production data showing steady-state skip rate ≥ 90%.
- **Does NOT** wire `Engine.applySnapshot` to populate `metaAppliedIndex` for peers receiving InstallSnapshot. The TODO sentinel from PR #915 round-6 still flags this; addressing it requires a separate design pass.
- **Does NOT** add an idle-cluster integration test (`ELASTICKV_RAFT_SNAPSHOT_COUNT=10` end-to-end). The unit tests cover the gate decision matrix; an integration test would prove the codex round-3 P2 scenario is closed end-to-end in production-like setup, which is most valuable as a follow-up after B3 has soaked.

## Soak plan

Same as B2: deploy and observe `elastickv_fsm_cold_start_restore_total{outcome="skipped"}` ramp up across cold starts. The design target is steady-state skip rate ≥ 90%; the per-outcome label distribution and gap distribution are the primary signals.

After one release of soak, Branch 4 can lower `HEALTH_TIMEOUT_SECONDS`.

## Refs

- PR #910 — design doc (rounds 1-7)
- PR #915 — Branch 2 plumbing (merged); this PR depends on `metaAppliedIndex` being populated
- PR #909 — `HEALTH_TIMEOUT_SECONDS` band-aid this series eventually obviates

## Summary by CodeRabbit

* **New Features**
  * Cold-start lifecycle callbacks added and wired through the raft engine to report skip/execute/fallback outcomes and applied-index gaps.
  * Prometheus metrics for cold-start outcomes and applied-index gap.
  * Snapshot-header read/apply contract introduced; KV FSM implements header apply and volatile-entry classification to support header-only skips and volatile-only replay on duplicates.

* **Tests**
  * Extensive unit/integration tests covering skip/execute/fallback decisions, header verification, volatile duplicate replay, and updated restore-call signatures.

### DIFF
--- a/internal/raftengine/cold_start.go
+++ b/internal/raftengine/cold_start.go
@@ -1,0 +1,40 @@
+package raftengine
+
+// ColdStartObserver receives cold-start snapshot-restore lifecycle
+// events from restoreSnapshotState. Implementations live in the
+// monitoring package and wire to Prometheus counters/gauges; the
+// engine receives a value through OpenConfig and treats nil as
+// "no metrics emitted" (preserves the byte-for-byte cold-start
+// behaviour for tests and callers that do not wire monitoring).
+//
+// Three outcomes match the design's strictly-additive policy
+// (docs/design/2026_06_02_idempotent_snapshot_restore.md §9):
+//
+//   - RestoreSkipped: the gate fired. `gap = haveAppliedIndex -
+//     snapshot.Metadata.Index` (how far ahead the live store was).
+//     This is the user-visible perf win.
+//
+//   - RestoreExecuted: the gate did NOT fire because the live store
+//     was genuinely stale (haveAppliedIndex < snapshot.Metadata.Index).
+//     `gap = snapshot.Metadata.Index - haveAppliedIndex` (the work
+//     the full restore re-did).
+//
+//   - RestoreFallback: the strictly-additive fallback path — the
+//     FSM did not expose AppliedIndexReader, LastAppliedIndex
+//     reported the meta key missing, or it returned an error. The
+//     full restore runs but the skip was never even attempted.
+//     `reason` carries a stable short label so Prometheus can
+//     surface why the optimisation could not engage:
+//
+//     not_reader   — FSM does not implement AppliedIndexReader
+//     missing_meta — meta key absent (pre-upgrade fsm.db)
+//     read_err     — LastAppliedIndex returned an error
+//
+// Implementations MUST NOT block; the engine calls these on the
+// cold-start critical path. Treat all label/string arguments as
+// untrusted enum values from the engine's enumeration above.
+type ColdStartObserver interface {
+	RestoreSkipped(snapIndex, haveAppliedIndex uint64)
+	RestoreExecuted(snapIndex, haveAppliedIndex uint64)
+	RestoreFallback(snapIndex uint64, reason string)
+}
--- a/internal/raftengine/etcd/cold_start_volatile_replay_test.go
+++ b/internal/raftengine/etcd/cold_start_volatile_replay_test.go
@@ -1,0 +1,161 @@
+package etcd
+
+import (
+	"io"
+	"sync/atomic"
+	"testing"
+
+	"github.com/bootjp/elastickv/internal/raftengine"
+	raftpb "go.etcd.io/raft/v3/raftpb"
+)
+
+// volatileTagFakeFSM is a StateMachine that classifies payloads by
+// their leading byte: 0x02 (kv.raftEncodeHLCLease) is volatile, every
+// other tag is data-mutating. Mirrors the kvFSM contract closely
+// enough that the engine's cold-start duplicate guard can be tested
+// without pulling in the full kv package.
+type volatileTagFakeFSM struct {
+	calls       atomic.Int32
+	lastPayload []byte
+}
+
+func (f *volatileTagFakeFSM) Apply(data []byte) any {
+	f.calls.Add(1)
+	cp := make([]byte, len(data))
+	copy(cp, data)
+	f.lastPayload = cp
+	return nil
+}
+
+func (f *volatileTagFakeFSM) Snapshot() (Snapshot, error) { return nil, nil }
+func (f *volatileTagFakeFSM) Restore(_ io.Reader) error   { return nil }
+
+func (f *volatileTagFakeFSM) IsVolatileOnlyPayload(payload []byte) bool {
+	return len(payload) > 0 && payload[0] == 0x02
+}
+
+var _ raftengine.VolatileEntryClassifier = (*volatileTagFakeFSM)(nil)
+
+// TestApplyNormalCommitted_VolatileEntryReplayedOnDuplicate pins the
+// codex P1 #934 round 7 fix: after the cold-start skip gate seeds
+// e.applied at the WAL committed tail, entries delivered by Raft at
+// indices <= e.applied that are volatile-only (HLC lease, tag 0x02)
+// MUST still reach fsm.Apply so the post-snapshot ceiling raise is
+// reconstructed. Data-mutating duplicates (any other tag) MUST NOT
+// reach fsm.Apply (OCC re-validation; ceiling regression). A
+// regression would either silently drop the lease (the bug) or
+// silently re-execute KV writes (idempotency violation).
+func TestApplyNormalCommitted_VolatileEntryReplayedOnDuplicate(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name      string
+		tag       byte
+		wantApply bool
+	}{
+		{"volatile HLC lease (tag 0x02) replays past e.applied", 0x02, true},
+		{"data-mutating single KV (tag 0x00) is dropped", 0x00, false},
+		{"data-mutating batch KV (tag 0x01) is dropped", 0x01, false},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			t.Parallel()
+			fsm := &volatileTagFakeFSM{}
+			e := newTestEngine(fsm, nil, nil)
+			e.applied = 200
+
+			payload := []byte{c.tag, 0xde, 0xad, 0xbe, 0xef, 0x00, 0x00, 0x00, 0x00}
+			entry := raftpb.Entry{
+				Type:  raftpb.EntryNormal,
+				Index: 150, // <= e.applied → duplicate path
+				Data:  encodeProposalEnvelope(42, payload),
+			}
+
+			if err := e.applyNormalCommitted(entry); err != nil {
+				t.Fatalf("applyNormalCommitted: %v", err)
+			}
+
+			got := fsm.calls.Load()
+			switch {
+			case c.wantApply && got != 1:
+				t.Fatalf("volatile duplicate: fsm.Apply call count = %d, want 1 (lost in-memory effect)", got)
+			case !c.wantApply && got != 0:
+				t.Fatalf("data-mutating duplicate: fsm.Apply call count = %d, want 0 (re-applied; OCC will re-validate against post-tail state)", got)
+			}
+
+			// In every case e.applied must NOT advance — the entry's
+			// index is below the gate-seeded value and resolveProposal
+			// is intentionally not called either.
+			if e.applied != 200 {
+				t.Fatalf("e.applied advanced to %d, want it pinned at 200 for duplicate-path entries", e.applied)
+			}
+		})
+	}
+}
+
+// TestApplyNormalCommitted_FreshEntryAlwaysAppliesAndAdvances pins the
+// non-duplicate path: entries past e.applied always reach fsm.Apply
+// regardless of the volatile/data classification, and e.applied
+// advances. Locks in the asymmetry — the classifier ONLY gates the
+// duplicate arm.
+func TestApplyNormalCommitted_FreshEntryAlwaysAppliesAndAdvances(t *testing.T) {
+	t.Parallel()
+	for _, tag := range []byte{0x00, 0x01, 0x02} {
+		t.Run("tag=0x0"+string("0123"[tag])+"_fresh", func(t *testing.T) {
+			t.Parallel()
+			fsm := &volatileTagFakeFSM{}
+			e := newTestEngine(fsm, nil, nil)
+			e.applied = 100
+
+			entry := raftpb.Entry{
+				Type:  raftpb.EntryNormal,
+				Index: 150,
+				Data:  encodeProposalEnvelope(7, []byte{tag, 0x99}),
+			}
+
+			if err := e.applyNormalCommitted(entry); err != nil {
+				t.Fatalf("applyNormalCommitted: %v", err)
+			}
+			if got := fsm.calls.Load(); got != 1 {
+				t.Fatalf("fsm.Apply call count = %d, want 1 (fresh entry, all tags)", got)
+			}
+			if e.applied != 150 {
+				t.Fatalf("e.applied = %d, want 150 (fresh entry must advance)", e.applied)
+			}
+		})
+	}
+}
+
+// TestApplyNormalCommitted_DuplicateWithoutClassifier_DropsAll guards
+// the FSM-doesn't-opt-in path: a StateMachine that does NOT implement
+// VolatileEntryClassifier (existing FSMs, third-party engines) must
+// keep the pre-PR behavior — every duplicate is dropped, including
+// any that happen to be volatile. The strictly-additive opt-in keeps
+// the engine compatible with FSMs that have not been updated.
+func TestApplyNormalCommitted_DuplicateWithoutClassifier_DropsAll(t *testing.T) {
+	t.Parallel()
+	// fakeStateMachine (from encryption_test.go) does NOT implement
+	// VolatileEntryClassifier — that absence is the contract under
+	// test.
+	fsm := &fakeStateMachine{}
+	if _, ok := any(fsm).(raftengine.VolatileEntryClassifier); ok {
+		t.Fatal("fakeStateMachine unexpectedly implements VolatileEntryClassifier; test contract broken")
+	}
+	e := newTestEngine(fsm, nil, nil)
+	e.applied = 200
+
+	entry := raftpb.Entry{
+		Type:  raftpb.EntryNormal,
+		Index: 150,
+		Data:  encodeProposalEnvelope(13, []byte{0x02, 0xff}), // would-be volatile
+	}
+
+	if err := e.applyNormalCommitted(entry); err != nil {
+		t.Fatalf("applyNormalCommitted: %v", err)
+	}
+	if got := fsm.calls.Load(); got != 0 {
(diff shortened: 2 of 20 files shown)


bootjp:
@claude review

Branch 3 of the PR #910 design — the user-visible cold-start perf win. Follows the merged Branch 2 (PR #915 — metaAppliedIndex plumbing). 6 commits, focused review pathways:

Review observation surface:
- §4 strictly-additive fallback: `decideSkipOutcome` collapses every uncertainty (FSM not a reader, meta key missing, read error) to a fallback enum; `LastAppliedIndex` errors MUST NOT abort cold start; tests `TestSkipGate_Fallback*` cover all three reasons.
- §5 two-phase seam (round-7): `ParseSnapshotHeader(r io.Reader) (ceiling, cutover uint64, err error)` + `ApplySnapshotHeader(ceiling, cutover uint64)`. Engine wraps r in a crc32 TeeReader; FSM Parse reads + drains; engine verifies CRC; engine calls Apply ONLY on verification success. The body bytes are read but discarded (fsm.db already has equivalent state).
- §5 CRC three-step (round-6): size + footer-vs-tokenCRC + full-body-CRC, byte-for-byte matching `openAndRestoreFSMSnapshot`'s safety contract. Tests `TestApplyHeaderStateOnSkip_*` inject each failure mode and assert FSM state unchanged.
- §6 nosync mode: not relevant here (B3 reads only; B2 handled the write side).
- §9 observability: 3 outcomes + gap label + structured INFO log. `raftengine.ColdStartObserver` interface in the parent package so monitoring can implement without circular import.
- Compile-time guards: `var _ raftengine.SnapshotHeaderApplier = (*kvFSM)(nil)` added next to the existing AppliedIndexReader/Writer guards — a rename or signature drift fails `go build` rather than silently degrading the skip to full-restore-always.
- Non-Goals respected: `Engine.applySnapshot` untouched (peer-after-InstallSnapshot bump remains a B4 / follow-up; the TODO sentinel from PR #915 round-6 still names it).
- Caller audit (per /loop semantic-change rule): `restoreSnapshotState` signature changed (added obs + logger). All in-package callers updated; 9 test call sites updated to pass (nil, nil) — tests opt out of metrics.

bootjp:
@claude review

Round-1 codex P2 addressed (commit 320d06ef):

`monitoring.Registry.ColdStartObserver()` now flows through to `OpenConfig.ColdStartObserver` in production:

- `etcd.FactoryConfig` gained `ColdStartObserver raftengine.ColdStartObserver`
- `Factory.Create` copies it into the `OpenConfig{}` literal at `factory.go:68`
- `main.go` reorders so `metricsRegistry` is built before `newRaftFactory`, then passes `metricsRegistry.ColdStartObserver()` in
- Three `newRaftFactory` callers in tests pass `nil` (those scopes do not exercise metrics)

Kept the change inside the **engine-specific** `etcd.FactoryConfig` rather than the shared `raftengine.FactoryConfig` to avoid touching `buildShardGroups`, `buildRuntimeForGroup`, and unrelated test wiring.

Verified: `go vet ./...`, `golangci-lint run` (0 issues across raftengine/etcd, monitoring, raftengine, root), `go test ./internal/raftengine/etcd/` green, all 9 `TestSkipGate_*` / `TestApplyHeaderStateOnSkip_*` still pass.

bootjp:
@claude review

Addresses **codex P1 #934** (`internal/raftengine/etcd/wal_store.go`) — cold-start skip gate would re-apply post-snapshot WAL entries.

**Bug:** when FSM applied > tok.Index (normal interval between snapshots — metaAppliedIndex advances each Apply), the skip path fired on `have >= tok.Index`. But `Open` still initializes raft's applied floor from `LocalSnap.Metadata.Index` and `openAndReadWAL` replays the post-snapshot WAL entries. Those entries (`tok.Index+1 .. have`) re-applied onto a Pebble store that already contained them → OCC conflicts in transaction commit + HLC ceiling left below timestamps already on disk.

**Fix:** reorder `loadWalState` so the WAL is opened BEFORE the skip-gate. Compute `lastWalIndex` (max entry Index in the WAL tail, or `snapshot.Metadata.Index` when empty) and gate skip on `have >= lastWalIndex`. When the WAL still carries entries beyond `have`, the gate falls through to full restore which runs through the normal apply path without double-applies.

Caller audit: `restoreSnapshotState` gains `lastWalIndex uint64` param. Single production call site (`loadWalState`) + 14 test call sites updated. `reportColdStart` extended to log both `snapshot_index` (operator context) and `last_wal_index` (actual threshold).

Regression: `TestSkipGate_ExecutesWhenWALCarriesPostSnapshotEntries` — snap=100, applied=150, lastWalIndex=200; would previously skip (have>snapIndex) but must execute because have<lastWalIndex.

go test + golangci-lint clean on `./internal/raftengine/etcd/`.

bootjp:
@claude review

Addresses all 4 findings on review 4434778631 + the coderabbit review on the same head:

1. **Codex P1 (`wal_store.go:293`) — advance Raft's applied floor on skip.** Round-1 fix gated skip on `have >= lastWalIndex` but Engine.Open still seeded `e.applied` from `maxAppliedIndex(LocalSnap) = snap.Index`. Raft would still deliver entries snap.Index+1..have to applyCommitted and replay them onto a Pebble store already containing them.

   **Fix:** `restoreSnapshotState` now returns `(effectiveApplied uint64, error)`. `diskState` carries `EffectiveApplied`. `Engine.Open` uses `coldStartApplied(disk) = max(maxAppliedIndex(LocalSnap), EffectiveApplied)` to seed `e.applied` and the atomic mirror. Belt+braces: `applyNormalCommitted` skips entries with `Index <= e.applied` (idempotent no-op); `setApplied` is now advance-only so conf changes in [snap.Index+1, have] don't regress the counter.

2. **Codex P2 / coderabbit Major (observer args).** ColdStartObserver contract is `(snapshotIndex, haveAppliedIndex)` — Prometheus computes `have - snapIndex`. Round-1 passed `target=lastWalIndex` instead. Reverted; lastWalIndex stays as a structured-log field.

3. **Coderabbit Major (emit-after-success).** `reportColdStart` moved to after `applyHeaderStateOnSkip` / `openAndRestoreFSMSnapshot` succeeds — CRC/header failures no longer register a 'successful' outcome.

Caller audit: `restoreSnapshotState` signature change (14 test sites updated). `setApplied` advance-only audited against all 4 apply paths. `coldStartApplied` new helper, single caller.

Regression: `TestSkipGate_ReturnsEffectiveAppliedOnSkip` (skip returns have, not snap.Index) + `TestSkipGate_EmitsAfterSuccess` (CRC failure does not record skipped/executed metrics).

go test + golangci-lint clean on `./internal/raftengine/etcd/`.

bootjp:
@claude review

Addresses **codex P2 #934 round 3** (review 4434778631 → 2 new inline findings):

1. **`wal_store.go:160` — gate on the committed index.** Round-2 used `entries[n-1].Index` (WAL tail) as the skip threshold. A follower's WAL can carry an UNCOMMITTED suffix that raft will not deliver via `Ready.CommittedEntries` until the leader confirms it. Followers with such suffixes were forced into a full restore on every restart.

   **Fix:** new `coldStartSkipThreshold(snapshot, hardState) = max(snapshot.Metadata.Index, hardState.Commit)`. Raft's invariant guarantees `Commit ≤ walTail`, so the walTail comparison was redundant for safety and harmful for the follower optimization.

2. **`monitoring/cold_start.go:80` — avoid underflowing the executed gauge.** Round-1's gate change means executed restores no longer imply `have < snapIndex` (crashed follower can have FSM ahead of snapshot but behind committed tail). The `snapIndex - have` subtraction underflowed into ~2^64.

   **Fix:** `RestoreExecuted` computes `|snapIndex - have|` (absolute distance). Comment updated.

Regression tests:
- `TestColdStartSkipThreshold` — 4-case table for helper.
- `TestColdStartObserver_ExecutedAbsoluteGap` — 3-case table covering FSM behind / FSM ahead / FSM equal.
- Plus `TestColdStartObserver_SkippedGap`, `_NilSafe`, `_FallbackReasonLabel` for monitoring coverage.

bootjp:
@claude review

Addresses both round-5 findings:

1. **Finding 1 — recordingObs test double drift.** `recordingObs.RestoreExecuted` updated to mirror `monitoring.ColdStartObserver`'s absolute-value computation (`|snapIndex - have|`). Tightened `TestSkipGate_ExecutesWhenWALCarriesPostSnapshotEntries` to assert the semantically correct gap (50) instead of a length-only check.

2. **Finding 2 — log/metric baseline drift.** Renamed log fields to disambiguate the two baselines:
   - Skipped: `gap_ahead_snapshot` (= metric baseline) + `gap_ahead_committed` (= committed-tail relative).
   - Executed: `gap_to_snapshot` (absolute, no underflow) + `gap_behind_committed`.
   - `last_wal_index` → `last_committed_index` since the helper now returns `max(snap.Index, hardState.Commit)`.

   Extracted the executed arm into `reportColdStartExecute` to keep `reportColdStart` under cyclop.

Finding 3 (defensive Commit < snapIndex guard) — confirmed correct as-is by claude; no change.

go test + golangci-lint clean on `./internal/raftengine/etcd/` + `./monitoring/`.

bootjp:
@claude review

Round 7 addresses codex P1: post-snapshot HLC lease effects were lost when the cold-start skip gate fired.

**Root cause.** The Round-1 idempotency guard
(`if entry.Index <= e.applied { return nil }`) treated every duplicate uniformly,
but HLC lease entries (`raftEncodeHLCLease`, tag 0x02) carry purely in-memory
effects (`HLC.SetPhysicalCeiling`). After the skip gate seeded `e.applied` at the
WAL committed tail, the guard dropped all post-snapshot lease entries.
`ApplySnapshotHeader` restored only the older snapshot-time ceiling, so the next
leader-issued fenced timestamp could collide with persisted commit_ts values
stamped under the lost lease.

**Fix.**
- New optional FSM seam `raftengine.VolatileEntryClassifier` —
  classifies cleartext FSM payloads as volatile-only or data-mutating.
- `kvFSM.IsVolatileOnlyPayload` returns `true` iff `payload[0] == raftEncodeHLCLease`.
  Static iface assertion pins the contract.
- `applyNormalEntry` now takes a `dropIfNonVolatile` flag. On the cold-start
  duplicate path the engine still decodes the envelope and decrypts
  (post-cutover HLC leases are encrypted), then either passes the payload to
  `fsm.Apply` (volatile — monotonic in-memory effect, no `setApplied`, no
  `resolveProposal`) or returns `(nil, nil)` (data-mutating — dropped, preserving
  the Round-1 OCC / sidecar idempotency).
- `applyNormalCommitted` forwards `duplicate` to `applyNormalEntry` and only
  advances `setApplied` + `resolveProposal` on the fresh path.

**Test.** `cold_start_volatile_replay_test.go` pins three contracts:
1. duplicate HLC lease → `fsm.Apply` called, `e.applied` unchanged
2. duplicate KV/MVCC → `fsm.Apply` NOT called, `e.applied` unchanged (Round-1 regression locked down)
3. FSM that doesn't implement the classifier → drop-all (no opt-in regression)

Plus the fresh-entry asymmetry: every tag still reaches `fsm.Apply` when
`entry.Index > e.applied`.

**Caller audit.** `applyNormalEntry` signature widened; every existing call site
(`applyNormalCommitted` + 10 `encryption_test.go` sites) passes `false`,
preserving the pre-PR semantics.

**Open items not addressed in this commit.**
- codex P2 (factory plumbing for `ColdStartObserver`) — separate concern
  scoped at `Factory.Create` / `OpenConfig`; will follow up if you want it
  bundled or in a follow-up PR.
- codex P2 (per-group metric labels) — same call; meaningful only after the
  factory plumbing lands so the registry observer is actually fed.

CI note: previous run hit a 504 on reviewdog install — transient infra, not a
lint failure on the code.

bootjp:
@claude review

Round 7 F1+F2 follow-up — both Minor findings from your last review are now closed.

**R7-F1 — `SetApplyIndex` stale-index on volatile replay.** Moved
`SetApplyIndex` inside an `if !dropIfNonVolatile {…}` guard so it only
fires on the fresh-entry arm. Today's only volatile entry (HLC lease)
doesn't read `pendingApplyIdx` so this is observably a no-op, but it
future-proofs the seam against a new volatile entry type that would
also persist (the seam's documented purpose).

**R7-F2 — Encrypted-volatile path untested.** Added
`TestApplyNormalCommitted_VolatileDuplicate_PostCutoverEncrypted`
which uses `raftCipherFixture` + `envelopeEntry` to wrap an HLC lease
above the cutover index but BELOW `e.applied`. Verifies:
1. `fsm.Apply` reached once (decrypt → classify → replay ordering),
2. FSM receives the cleartext payload byte-for-byte (classifier must
   see post-decrypt bytes),
3. `e.applied` pinned at 200 (no advance on duplicate arm).

**Refactor note.** Pulled the classifier check into
`Engine.isVolatilePayload` to stay under the cyclop budget — same
semantics, cleaner call site.

Tests + lint clean locally:
- `go test -race ./internal/raftengine/etcd/ ./kv/...` ok
- `golangci-lint run ./internal/raftengine/... ./kv/...` 0 issues

R1-F1 (header-error-before-CRC ordering doc comment) is the only
remaining open item; you flagged it as doc-only, no safety impact.

bootjp:
@claude review

Final cleanup: R1-F1 doc comment in `applyHeaderStateOnSkip` explaining the header-error-before-CRC ordering. No code change. Every finding across 8 rounds is now closed.

PR #943, opened 2026-06-10
feat(redis): flip one-phase txn dedup default to on (closes #937)

### LINKED ISSUE #937: Jepsen Scheduled Stress (Redis): duplicate-elements + future-read + G-single/G2-item-realtime on main
# Jepsen Scheduled Stress (Redis workload) — duplicate-elements + future-read + G-single/G2-item-realtime on main

## Summary

The `Jepsen Scheduled Stress Test` run [#27033231956](https://github.com/bootjp/elastickv/actions/runs/27033231956/job/79790587770) (Redis list-append workload, scheduled cron on `main`) failed validation with a hard anomaly set. The prior scheduled stress run on main at 12:57 UTC the same day ([#27016167818](https://github.com/bootjp/elastickv/actions/runs/27016167818)) passed.

The single commit landed on `main` between the two scheduled runs is PR #928 — strictly inside `internal/backup/` (S3 backup encoder); it does not touch the Redis adapter, the Raft engine, the coordinator, or any path the Redis workload exercises. So this is **not introduced by PR #928** — it is a pre-existing edge case that the stress harness occasionally exposes.

Per-PR Jepsen runs (`jepsen-test.yml`) use `--time-limit 5 --rate 5 --concurrency 5`. The scheduled stress uses `--time-limit 150 --rate 10 --concurrency 8 --key-count 16 --max-writes-per-key 250 --max-txn-length 4` — ~30× wall-clock time, 2× throughput per client, longer transactions. The duplicate / future-read patterns surface only at that scale.

## Verdict (from the failed run)

```clojure
{:valid? false,
 :anomaly-types (:G-single-item-realtime
                 :G2-item-realtime
                 :cycle-search-timeout
                 :duplicate-elements
                 :future-read)}
```

Notable anomaly entries:

| Type | Evidence |
|---|---|
| `:duplicate-elements` | `{:duplicates {187 2}}` on op index 525, process 1, `:f :txn` reading key 9 = `[1 2 3]`, appending `[13 50]`, reading key 15. Element 187 appears twice in some list. |
| `:future-read` | op index 1073, process 18, transactional read returned a list whose tail contained values not yet appended in real-time at the read-start instant. |
| `:G-single-item-realtime` | first witness at op index 1667 (~84.2s into the run); second around 1669. SCC traversed; cycle confirmed. |
| `:G2-item-realtime` | op index 2797 (~137.7s into the run). Multi-key anti-dependency cycle. |
| `:cycle-search-timeout` | `{:scc-size 495 :anomaly-spec-type :G-single-item-realtime}` — Elle exhausted its budget exploring one SCC. Indicates the history has many candidate cycles to check, consistent with the realtime violations above. |

The two anomalies that are most diagnostic are `:duplicate-elements` and `:future-read` — these are not "noise" from real-time analysis alone:

- **`:duplicate-elements`** in a list-append workload means *the same value appended twice to the same list*. The standard cause is **client retry of a transaction that actually committed at the server but whose ACK was lost** (timeout / connection close / leader change mid-RPC). Without an idempotency token or matching commit-deduplication path, the retry produces a second commit.
- **`:future-read`** = a successful read returned a value before the value's write started in wall-clock real-time. Combined with `:duplicate-elements`, the most likely explanation is **the same retry-committed-twice value being observable on a path that is not tied to the first write's commit timestamp**.

Both are consistent with an **ambiguous-commit window**: the Redis adapter, the Lua bridge, or the txn dispatch returns an uncertain error (`io.unknown`) on the path back to the client, the client retries, the original commit had already been applied → duplicate append + tail of the list reordered relative to other clients' reads.

## Reproduction profile

- Branch: `main`, HEAD `00a8d0bc` ("backup: Phase 0b M4-2b impl - S3 collision-rename reversal (#928)").
- Workflow: `.github/workflows/jepsen-test-scheduled.yml` (display name `Jepsen Scheduled Stress Test`).
- Workload: `elastickv.redis-workload` (list-append, Elle checker).
- Args: `--time-limit 150 --rate 10 --concurrency 8 --key-count 16 --max-writes-per-key 250 --max-txn-length 4 --ports 63791,63792,63793 --host 127.0.0.1`.
- Reproducibility unknown — only two scheduled stress runs exist in the visible history; 12:57 passed, 18:37 failed. Recommend manually re-dispatching once or twice to confirm whether this is reliably reproducible or sporadic.

## Suggested investigation order

1. **Audit the Redis adapter's RPC-failure → client-error → retry boundary** for any commit-applied-but-not-acked window. The relevant entry points are `adapter/redis.go` and the Lua dispatch (`adapter/redis_lua_pool.go`); the Raft propose ACK path is in `kv/sharded_coordinator.go` / `kv/shard_store.go`.
2. **Check the EXEC reuse / dedup path** (PR #46 — `runTransactionWithDedup`, `prepareDispatch`, `onePhaseTxnDedup`). The dedup probe was meant to make exactly this class of retry idempotent for one-phase txns; verify the same protection covers the path that produced the duplicate here (`:r 9 → :append 13 50 → :r 15`).
3. **Reproduce locally** with `./scripts/run-jepsen-local.sh` using the same stress parameters once it's clear which adapter code path is suspect. The stress params are at the top of `jepsen-test-scheduled.yml` — paste them into the local script.
4. **If reproduced**, add an Elle regression test under `jepsen/` (or a Go-side targeted test under `adapter/`) that exercises the ambiguous-commit retry directly so the next regression is caught at PR time, not at scheduled-stress time.

## Out of scope

- **Not introduced by PR #934** (`impl/snapshot-skip-b3-skip-gate`, cold-start FSM snapshot-restore skip metrics plumbing) — different branch, per-PR Jepsen passed, metrics-only change.
- **Not introduced by PR #928** (S3 backup collision-rename encoder) — only touches `internal/backup/`.

## Links

- Failing run: https://github.com/bootjp/elastickv/actions/runs/27033231956/job/79790587770
- Prior passing scheduled stress: https://github.com/bootjp/elastickv/actions/runs/27016167818
- PR #928 (in-window main commit, S3-only): https://github.com/bootjp/elastickv/pull/928

bootjp:
## Summary

Flips `adapter.RedisServer.onePhaseTxnDedup` from default-off to default-on, closing the rollout of the option-2 idempotency design landed by [`docs/design/2026_05_21_proposed_txn_secondary_idempotency.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_05_21_proposed_txn_secondary_idempotency.md). No new mechanism — the FSM probe (`dedupProbeOnePhase`), the wire change (`TxnMeta.PrevCommitTS`, V2-only when non-zero), and every adapter-side reuse path already ship and are exercised every day by the dedup-mode Jepsen workflow.

Closes #937.

## Authorization

The parent design's `M4` 7-day criterion ([source](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_05_21_proposed_txn_secondary_idempotency.md#m4--validation)) reads:

> 7 consecutive days without `:duplicate-elements` / `:G-single-item-realtime` in the dedup-mode workflow.

[`jepsen-test-scheduled-dedup.yml`](https://github.com/bootjp/elastickv/blob/main/.github/workflows/jepsen-test-scheduled-dedup.yml) (Redis stress profile, `ELASTICKV_REDIS_ONEPHASE_DEDUP=1`) has produced **12 consecutive green runs over 10 calendar days, 2026-05-31 → 2026-06-10**. Run IDs are listed in the new design doc; the most recent are:

| Date (UTC) | Run | Conclusion |
|---|---|---|
| 2026-06-10 04:50 | [27253991270](https://github.com/bootjp/elastickv/actions/runs/27253991270) | success |
| 2026-06-09 04:44 | [27184402445](https://github.com/bootjp/elastickv/actions/runs/27184402445) | success |
| 2026-06-08 04:57 | [27116904871](https://github.com/bootjp/elastickv/actions/runs/27116904871) | success |
| 2026-06-07 04:54 | [27083142341](https://github.com/bootjp/elastickv/actions/runs/27083142341) | success |
| 2026-06-06 04:40 | [27052820868](https://github.com/bootjp/elastickv/actions/runs/27052820868) | success |
| 2026-06-05 04:49 | [26996058409](https://github.com/bootjp/elastickv/actions/runs/26996058409) | success |
| 2026-06-04 04:57 | [26931749014](https://github.com/bootjp/elastickv/actions/runs/26931749014) | success |
| 2026-06-04 04:16 | [26930308744](https://github.com/bootjp/elastickv/actions/runs/26930308744) | success |
| 2026-06-03 04:58 | [26864667493](https://github.com/bootjp/elastickv/actions/runs/26864667493) | success |
| ...(2 more) | | success |

Exceeds the 7-day threshold. The parent's R5 (FSM determinism across a rolling upgrade) is discharged — the probe code has shipped on every production node for months.

## Change

| File | Change |
|---|---|
| `docs/design/2026_06_10_proposed_redis_onephase_dedup_default_on.md` | New design doc (commit 1) recording M4 evidence and the rollout. |
| `adapter/redis.go` | Flips env-var sense from `== "1"` to `!= "0"`; updates the comment, godoc, and field comment. |
| `.github/workflows/jepsen-test-scheduled.yml` | Adds explicit `ELASTICKV_REDIS_ONEPHASE_DEDUP: "0"` to the job env so the control baseline keeps its meaning across the default change. Retirement of the control workflow is a 30-day follow-up. |

## Behaviour change

- **Default**: dedup gate is now **on** for any process that does not opt out.
- **Opt-out**: `ELASTICKV_REDIS_ONEPHASE_DEDUP=0` (single-env-var rollback, no binary revert). The constructor option `WithOnePhaseTxnDedup(false)` still trumps the env.
- **Existing opt-in users**: `=1` (the dedup-mode workflow's existing setting) continues to resolve to `true`. No breakage.

## Risk

| Concern | Disposition |
|---|---|
| Data loss | None — the dedup path only short-circuits on a confirmed prior commit at the exact `prev_commit_ts` (the parent design's R1 + the existing `kv/fsm.go` probe). No write is dropped. |
| Concurrency / distributed failures | None — every node already runs the probe code; no new fanout, no new round-trip. R5 discharged as above. |
| Performance | Default-on adds the `PrevCommitTS` probe on each retryable attempt. Cost is one `CommittedVersionAt` MVCC lookup at the primary key + stale `commit_ts`; this is the same cost the dedup-mode workflow has been paying for 12 days without regression. No hot-path allocation change. |
| Data consistency | Improved — the failure modes the parent design predicted (`:duplicate-elements`, `:future-read`, `:G-single-item-realtime`) become unreachable on the protected path. Issue #937 is the most recent control-baseline reproduction. |
| Test coverage | No new branches added; existing `adapter/redis_*_dedup_test.go` suites cover both gate states. Tests that construct `&RedisServer{...}` directly (zero-value `onePhaseTxnDedup`) still observe the legacy path, so the off-path remains covered at unit level. |

## Verification

- `go vet ./...` — clean
- `go build ./...` — clean
- `go test -run 'Dedup|OnePhase|PrevCommit|Idempot' ./adapter/` — pass (1.2s)
- `go test -run 'TestRedis|TestList|TestSet|TestZSet|TestStream|TestExec|TestMulti' ./adapter/` — pass (169s)
- `golangci-lint run ./adapter/...` — 0 issues

## Follow-ups

- **30 days post-merge** — review whether `jepsen-test-scheduled.yml` (now explicit dedup-off) still adds signal. If not, retire it.
- **DynamoDB default-on** — parallel work tracked by [`2026_06_03_partial_dynamodb_onephase_dedup.md`](https://github.com/bootjp/elastickv/blob/main/docs/design/2026_06_03_partial_dynamodb_onephase_dedup.md)'s `M2`. Not part of this PR.
- **S3, SQS dedup paths** — not yet routed through one-phase reuse; out of scope for the parent design.

## Summary by CodeRabbit

* **Documentation**
  * Added a design doc describing a planned default-on change for Redis transaction dedup and rollout/verification plans.

* **Chores**
  * Flipped adapter default behavior so transaction dedup is enabled by default, with an environment opt-out preserved.
  * Introduced a separate opt-in gate for standalone SET dedup to avoid unintended behavior changes.

* **Tests**
  * Updated standalone SET tests and added a regression test to verify overwrite semantics remain correct under current defaults.

### DIFF
--- a/.github/workflows/jepsen-test-scheduled.yml
+++ b/.github/workflows/jepsen-test-scheduled.yml
@@ -39,6 +39,19 @@ jobs:
     runs-on: ubuntu-latest
     env:
       GOCACHE: /tmp/go-build
+      # Explicit dedup-OFF control baseline. The Redis adapter's
+      # onePhaseTxnDedup flipped to default-on in
+      # docs/design/2026_06_10_proposed_redis_onephase_dedup_default_on.md;
+      # this workflow is preserved as the legacy-path coverage so anomalies
+      # the dedup gate prevents (`:duplicate-elements`, `:future-read`,
+      # `:G-single-item-realtime`) continue to be measured against an
+      # unprotected build. Pair with the dedup-ON workflow
+      # (.github/workflows/jepsen-test-scheduled-dedup.yml) which sets
+      # ELASTICKV_REDIS_ONEPHASE_DEDUP=1 explicitly. Retirement of this
+      # workflow is a follow-up after 30 days of post-flip data; until
+      # then, do NOT remove this env var — without it the two workflows
+      # would exercise the same path under the new default.
+      ELASTICKV_REDIS_ONEPHASE_DEDUP: "0"
     steps:
       - uses: actions/checkout@v6
         with:
--- a/adapter/redis.go
+++ b/adapter/redis.go
@@ -251,20 +251,48 @@ type RedisServer struct {
 	// to opt out — kept as a one-env-var operator rollback.
 	onePhaseTxnDedup bool
 
+	// standaloneSetDedup gates whether the *standalone* SET command (not SET
+	// inside MULTI/EXEC) routes through runTransactionWithDedup. Default off
+	// because the dedup path's applySet does not match the legacy
+	// executeSet/replaceWithStringTxn semantics for SET-over-collection: a
+	// `SET k v` after `RPUSH k x` returns WRONGTYPE on the dedup path but
+	// overwrites correctly on the legacy path (PR #943 round-1 codex P1).
+	// Bringing applySet to parity (collection-deletion + string write inside
+	// the dedup txn) is tracked as a follow-up; until that lands, the
+	// standalone SET path stays on the legacy default-on-flip code path
+	// regardless of onePhaseTxnDedup's value. Enable explicitly via
+	// WithStandaloneSetDedup(true) or ELASTICKV_REDIS_ONEPHASE_DEDUP_SET=1
+	// only when applySet's parity is verified for the workload at hand.
+	standaloneSetDedup bool
+
 	route map[string]func(conn redcon.Conn, cmd redcon.Command)
 }
 
 type RedisServerOption func(*RedisServer)
 
 // WithOnePhaseTxnDedup enables (or disables) the option-2 one-phase
-// idempotency dedup on list-push, MULTI/EXEC, and standalone-write retries
+// idempotency dedup on list-push and MULTI/EXEC retries
 // (see RedisServer.onePhaseTxnDedup). On by default since the rollout
 // recorded in docs/design/2026_06_10_proposed_redis_onephase_dedup_default_on.md;
 // pass false to opt out from code, or set ELASTICKV_REDIS_ONEPHASE_DEDUP=0
 // to opt out from the environment. The constructor option trumps the env var.
+// Standalone SET requires the separate WithStandaloneSetDedup gate; see
+// RedisServer.standaloneSetDedup.
 func WithOnePhaseTxnDedup(enabled bool) RedisServerOption {
 	return func(r *RedisServer) {
 		r.onePhaseTxnDedup = enabled
+	}
+}
+
+// WithStandaloneSetDedup enables the option-2 dedup path on the *standalone*
+// SET command (not SET inside MULTI/EXEC). Off by default because the dedup
+// path's applySet does not yet match the legacy executeSet semantics for
+// SET-over-collection — see RedisServer.standaloneSetDedup. Enable only
+// after verifying applySet parity for the workload (no SET-over-list /
+// SET-over-hash / SET-over-set / SET-over-zset / SET-over-stream issued).
+func WithStandaloneSetDedup(enabled bool) RedisServerOption {
+	return func(r *RedisServer) {
+		r.standaloneSetDedup = enabled
 	}
 }
 
@@ -506,10 +534,13 @@ func NewRedisServer(listen net.Listener, redisAddr string, store store.MVCCStore
 		// ELASTICKV_REDIS_ONEPHASE_DEDUP=0 opts out; the WithOnePhaseTxnDedup
 		// constructor option still trumps the env var.
 		onePhaseTxnDedup: os.Getenv("ELASTICKV_REDIS_ONEPHASE_DEDUP") != "0",
-		baseCtx:          baseCtx,
-		baseCancel:       baseCancel,
-		streamWaiters:    newKeyWaiterRegistry(),
-		zsetWaiters:      newKeyWaiterRegistry(),
+		// standaloneSetDedup defaults off; see field comment for the
+		// applySet-vs-executeSet parity gap that gates this separately.
+		standaloneSetDedup: os.Getenv("ELASTICKV_REDIS_ONEPHASE_DEDUP_SET") == "1",
+		baseCtx:            baseCtx,
+		baseCancel:         baseCancel,
+		streamWaiters:      newKeyWaiterRegistry(),
+		zsetWaiters:        newKeyWaiterRegistry(),
 	}
 	r.relay.Bind(r.publishLocal)
 
@@ -1137,7 +1168,13 @@ func (r *RedisServer) set(conn redcon.Conn, cmd redcon.Command) {
 	// SET there is exactly one element with the same redisResult shape as
 	// the standalone reply (resultString OK / resultNil for NX/XX miss /
 	// resultBulk for GET).
-	if r.onePhaseTxnDedup {
+	// Both gates must be on to route standalone SET through the dedup path.
+	// onePhaseTxnDedup covers the MULTI/EXEC and list-push retries that the
+	// parent design's M4 validated; standaloneSetDedup is a separate sub-gate
+	// (default off) because applySet diverges from executeSet on SET-over-
+	// collection — flipping onePhaseTxnDedup default-on without this guard
+	// would change normal Redis overwrite behaviour (PR #943 round-1 codex P1).
+	if r.onePhaseTxnDedup && r.standaloneSetDedup {
 		// Call runTransactionWithDedup directly instead of going through
 		// runTransaction. runTransaction re-checks the same
 		// r.onePhaseTxnDedup gate and routes here anyway; the indirection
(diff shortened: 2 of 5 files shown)


bootjp:
@claude review

bootjp:
@claude review

Round-1 responses (commit a5909eb8):

### codex P1 — Preserve SET overwrites before defaulting dedup on ✅ FIXED

You're right — txnContext.applySet (adapter/redis.go:2356) returns WRONGTYPE for SET-over-collection, while the legacy setLegacy → executeSet → replaceWithStringTxn path correctly deletes the collection's logical elements and writes the string. Flipping onePhaseTxnDedup default-on without a separate gate on standalone SET would have broken normal Redis overwrite semantics.

Fix: introduced a per-path sub-gate `standaloneSetDedup` (default off). The standalone SET branch in `set()` now requires **both** `onePhaseTxnDedup` **and** `standaloneSetDedup` to be on. With the default config the SET path falls back to `setLegacy`, preserving legacy semantics. MULTI/EXEC and list-push paths (the parent design's actual M4-validated workloads) are unaffected.

Regression test `TestRedis_SET_OverwritesList_UnderDefaultGate` (new file adapter/redis_set_overwrite_default_test.go) seeds `RPUSH k x`, runs `SET k v` under default config, asserts OK + GET returns v. Fails on the pre-fix build (WRONGTYPE), passes on the fixed build.

Bringing applySet to parity with executeSet is a larger surgery; tracking as a follow-up before re-flipping the sub-gate.

### gemini medium — broken [#937] reference link ✅ FIXED

Replaced with an inline link on line 83.

### gemini medium — accept "false" as opt-out ❌ DECLINED

elastickv's env-var convention across the codebase is binary "0"/"1" (e.g. ELASTICKV_REDIS_TRACE == "1" two lines above the SET dedup site). Accepting "false" without also accepting "no" / "off" / "FALSE" would be inconsistent; accepting all of them would diverge from every other elastickv env-var read. The opt-out value "0" is documented in the godoc and in-file comment. Project-wide convention shift toward truthy-string parsing is a separate refactor — out of scope here.

### CI Analyze (go) failure — workflow permissions, not code 🚧 NOT FIXED HERE

CodeQL ran fine (287/605 files, 7 diagnostics, SARIF exported) but the upload step failed with `security-events: write` missing from GITHUB_TOKEN scope. Workflow-configuration concern, not a code change for #943. Will track separately.

### coderabbit minor — same [#937] link as gemini ✅ FIXED (subsumed)

### Verification

- go vet ./... clean
- go build ./... clean
- golangci-lint run ./adapter/... 0 issues
- TestStandaloneSetDedup + TestRedis_SET_OverwritesList_UnderDefaultGate (5 tests) — all pass
- Dedup/OnePhase/PrevCommit test family — all pass

PR #947, opened 2026-06-10
perf(lease-read): cache LeaseProvider type assertion on Coordinate and ShardGroup

### LINKED ISSUE #555: perf(lease-read): cache LeaseProvider type assertion (deferred from #549)
### Issue
`Coordinate.LeaseRead` and `groupLeaseRead` perform `c.engine.(raftengine.LeaseProvider)` on every call. Each type assertion is ~1-3 ns; not a bottleneck but trivially cacheable.

### Fix
- Add `lp raftengine.LeaseProvider` field on `Coordinate` and `ShardGroup`.
- Set once in `NewCoordinatorWithEngine` / `NewShardedCoordinator` (after the type assertion).
- `LeaseRead` checks `c.lp != nil` instead of asserting.

### Acceptance
- Same observable behavior; `LeaseRead` benchmark shows reduced cycles per op.

References: PR #549 (deferred).

bootjp:
Closes #555.

## What

`Coordinate.LeaseRead` and `groupLeaseRead` performed the interface type assertion `engine.(raftengine.LeaseProvider)` on every call (a hot path: Redis GET and all lease reads). This caches the assertion once at construction:

- Added `lp raftengine.LeaseProvider` field to `Coordinate` (`kv/coordinator.go`), set once in `NewCoordinatorWithEngine` in the same block that already asserts for the leader-loss callback.
- Added `lp raftengine.LeaseProvider` field to `ShardGroup` (`kv/sharded_coordinator.go`), set once per group in the `NewShardedCoordinator` loop that already asserts for the per-shard leader-loss callback.
- `LeaseRead`, `refreshLeaseAfterDispatch`, `groupLeaseRead`, and `leaseRefreshingTxn.maybeRefresh` now test `lp == nil` instead of re-asserting. `groupLeaseRead` keeps a nil-group guard so `engineForGroup`'s nil-safety is preserved.
- Added `BenchmarkLeaseRead` / `BenchmarkGroupLeaseRead` (`kv/lease_read_benchmark_test.go`) exercising the fast path; both assert the slow path (`LinearizableRead`) never runs so the benchmark can't silently measure the wrong path.

The `engine` field is never reassigned after construction, so the cached value stays valid for the object's lifetime without a lock.

## Behavior change

None expected. The assertion result is identical; it is now computed once instead of per call. All existing lease tests pass unchanged.

## Risk

Low. The only subtlety is construction ordering in `NewShardedCoordinator`: the `leaseRefreshingTxn` wrapper is created earlier in the same loop iteration than `g.lp` is assigned, but it holds a `*ShardGroup` pointer (not a copy), and `maybeRefresh` only runs at Commit time — long after the loop finishes — so `g.lp` is always populated by then.

## Scope note / deviation

`leaseReadEngineCtx` (`kv/raft_engine.go`) also does a per-call `LeaseProvider` assertion, but it takes a bare `raftengine.LeaderView` and is shared across the internal storage read path (`shard_store.go`); it has no `Coordinate`/`ShardGroup` to cache on. The issue's accepted direction scopes the field to `Coordinate` and `ShardGroup`, so `leaseReadEngineCtx` is intentionally left unchanged.

## Test evidence

`go test -race ./kv/...`:
```
ok  	github.com/bootjp/elastickv/kv	10.499s
```

`golangci-lint --config=.golangci.yaml run ./kv/...`:
```
0 issues.
```

Benchmarks (Apple M1 Max, fast path, 0 allocs, LinearizableRead never invoked):
```
BenchmarkLeaseRead-10         	 9749521	       209.1 ns/op	       0 B/op	       0 allocs/op
BenchmarkGroupLeaseRead-10    	11888186	        89.07 ns/op	       0 B/op	       0 allocs/op
```

## Self-review

1. **Data loss** — No persistence/Raft propose/apply/snapshot path touched. Lease-read semantics (fast vs. slow path selection) are byte-for-byte identical; only the cached assertion replaces the per-call one. No new error-swallowing. No risk.
2. **Concurrency / distributed failures** — `lp` is written once during construction and only read afterward; `engine` is never reassigned, so no lock is needed and there is no data race (confirmed by `go test -race`). Construction ordering of `leaseRefreshingTxn` vs. `g.lp` assignment is safe (pointer aliasing; `maybeRefresh` runs only at Commit time). Leader-loss callback registration is unchanged.
3. **Performance** — Removes an interface type assertion from the lease-read hot path, replacing it with a single nil check on a struct field. Benchmarks show 0 allocs and the fast path is preserved. No extra Raft round-trips, no new lock contention.
4. **Data consistency** — No change to MVCC visibility, OCC commit-ts ordering, HLC ceiling, or the lease freshness bound (`engineLeaseAckValid` and the secondary caller-side lease check are untouched). The fallback to `LinearizableRead` when the engine lacks `LeaseProvider` or the lease is disabled is preserved exactly.
5. **Test coverage** — Existing lease tests (`TestCoordinate_LeaseRead_*`, `TestShardedCoordinator_LeaseRead*`, fallback-when-engine-lacks-LeaseProvider, leader-loss registration) cover both the cached-present and nil paths and pass unchanged. Added two benchmarks that fail if the slow path is taken, locking in the "assertion off the hot path" claim.

### DIFF
--- a/kv/coordinator.go
+++ b/kv/coordinator.go
@@ -120,6 +120,11 @@ func NewCoordinatorWithEngine(txm Transactional, engine raftengine.Engine, opts 
 	for _, opt := range opts {
 		opt(c)
 	}
+	// Resolve the optional LeaseProvider capability once here so the
+	// LeaseRead / refreshLeaseAfterDispatch hot paths test a cached
+	// field instead of repeating the interface type assertion per call.
+	// engine is never reassigned after construction, so the cached value
+	// stays valid for the Coordinate's lifetime.
 	// Register a leader-loss hook so the lease is invalidated the instant
 	// the engine notices a state transition out of the leader role,
 	// rather than waiting for wall-clock expiry of the current lease.
@@ -128,6 +133,7 @@ func NewCoordinatorWithEngine(txm Transactional, engine raftengine.Engine, opts 
 	// one-shot tools) MUST call Close() to avoid leaking a closure
 	// pointing into this Coordinate.
 	if lp, ok := engine.(raftengine.LeaseProvider); ok {
+		c.lp = lp
 		c.deregisterLeaseCb = lp.RegisterLeaderLossCallback(c.lease.invalidate)
 	}
 	return c
@@ -169,10 +175,18 @@ type CoordinateResponse struct {
 type Coordinate struct {
 	transactionManager Transactional
 	engine             raftengine.Engine
-	clock              *HLC
-	connCache          GRPCConnCache
-	log                *slog.Logger
-	lease              leaseState
+	// lp caches the engine's optional LeaseProvider capability so the
+	// LeaseRead hot path (and refreshLeaseAfterDispatch) test a single
+	// field for nil instead of performing an interface type assertion on
+	// every call. It is set once in NewCoordinatorWithEngine and is nil
+	// when the engine does not implement raftengine.LeaseProvider. The
+	// engine field is never reassigned after construction, so this stays
+	// in sync without a lock.
+	lp        raftengine.LeaseProvider
+	clock     *HLC
+	connCache GRPCConnCache
+	log       *slog.Logger
+	lease     leaseState
 	// deregisterLeaseCb removes the leader-loss callback registered
 	// against engine at construction. Long-lived Coordinates don't
 	// need to call it (the engine will be closed after them), but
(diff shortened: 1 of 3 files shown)


bootjp:
@claude review

bootjp:
@claude review

PR #954, opened 2026-06-12
docs(design): propose scaling roadmap

bootjp:
## What

Adds `docs/design/2026_06_12_proposed_scaling_roadmap.md` — a doc-only scaling roadmap that surveys what exists, what is proposed, and what is **missing** for elastickv to scale further (more data, more throughput, more nodes).

This is a roadmap / implementation-plan doc, which `docs/design/README.md` explicitly permits ("Concrete implementation plans"). No code changes.

## Structure

1. **Current scaling envelope** — the seven things that bound a deployment today, each anchored to a file:line on `main`.
2. **Dimension-by-dimension analysis** — (a) data volume → split M2/M3 + blob offload + shared cache, with range MERGE flagged as a missing design; (b) write throughput → multi-group + leader balance (#953) + the multi-node bootstrap gap; (c) read throughput → lease reads (done), follower-reads (missing design, learner primitive present); (d) cross-group txn → per-group HLC vs centralized TSO assessment; (e) cluster size / membership; (f) operational scaling.
3. **Gap list → recommended new designs** ranked by leverage, each with depends-on edges.
4. **Sequencing** — dependency-ordered rollout tying in #945 / #951 / #953.
5. **Open Questions.**

## Verified-on-tree claims (selection)

- `RunHLCLeaseRenewal` proposes the ceiling to the **default group only** (`kv/sharded_coordinator.go:1915`) — the TSO doc's near-term "all led groups" fix is **not** yet implemented.
- `--raftBootstrapMembers` is rejected for multi-group (`main.go:746`, `len(groups) != 1`); each group bootstraps single-member (`multiraft_runtime.go:246-254`) — the multi-node multi-group gap PR #953 §1.1a (PR0) documents.
- Each Pebble store allocates a private 256 MiB block cache (`store/lsm_store.go:273-279`); shared-cache is a TODO comment (`:117-120`), not a design.
- FSM apply is serial per group (`kv/fsm.go:123-124`).
- The raft **learner primitive is implemented** in code (engine `AddLearner`/`PromoteLearner`, raftadmin CLI, `--raftJoinAsLearner`) even though the doc filename still reads `proposed` — but follower-served reads (its consumer) have no design.
- Snapshot disk offload and lease reads are implemented; TSO and range merge are not.

## Notes

- References a companion proposal being written in parallel, `2026_06_12_proposed_multinode_multigroup_bootstrap.md`, by name (it is the authoritative spec for Gap 1).
- All referenced design docs were confirmed to exist (or are named in-flight PR branches: #945 `docs/hotspot-split-m2-proposal`, #951 `design/hotspot-split-m3-automation`, #953 `design/leader-balance-scheduler`).

### DIFF
--- a/docs/design/2026_06_12_proposed_scaling_roadmap.md
+++ b/docs/design/2026_06_12_proposed_scaling_roadmap.md
@@ -219,22 +219,25 @@ memory each group's private cache/memtable pins.
   M6–M7) is a larger component, but the need for a shared ordering source is
   not a throughput/amortization question once cross-node cross-group
   transactions are possible: with the per-group fix in place, every node's
-  timestamps are self-monotonic, but **global** monotonicity across nodes
-  leading different groups is still not guaranteed without a shared oracle
-  (TSO doc §6 "Guarantee"). Cross-group transactions (`kv/transaction.go`,
-  `kv/txn_codec.go`) that span groups led by different nodes need a single
-  ordering source for OCC commit-ts comparability. (The shared-snapshot
+  timestamps are self-monotonic, but **global** monotonicity across coordinator
+  nodes is still not guaranteed without a shared oracle (TSO doc §6
+  "Guarantee"). Cross-group transactions (`kv/transaction.go`,
+  `kv/txn_codec.go`) whose timestamps can be allocated by different ingress /
+  coordinator nodes need a single ordering source for OCC commit-ts
+  comparability, regardless of where the participating groups' current leaders
+  live. `LeaderProxy.Commit` / `Internal.Forward` preserve non-zero timestamps,
+  so leader co-location does not prove single-clock allocation. (The shared-snapshot
   invariant — every operation in one txn reading at the *same* `startTS` — is
   already upheld: `nextStartTS` allocates one `startTS` for the whole txn and
   propagates it via `reqs.StartTS` to every participating group; the gap is the
-  cross-*node* comparability of the per-txn `commitTS`, not the per-txn
+  cross-*coordinator* comparability of the per-txn `commitTS`, not the per-txn
   `startTS`. See OQ-1.) So: per-group renewal fix is in-scope-soon and
-  load-bearing for one node leading multiple groups; before enabling the first
-  cross-group transaction whose participating groups are led by different
-  nodes, the roadmap must either pull the dedicated TSO group forward or land a
-  narrower single-oracle bridge that pins cross-group timestamp allocation to
-  one designated leader. Until such txns are enabled, the per-group fix plus
-  the shared-`*HLC` property remains adequate.
+  load-bearing for one node leading multiple groups; before enabling any
+  cross-group transaction mode in which more than one coordinator node can issue
+  `startTS` / `commitTS`, the roadmap must either pull the dedicated TSO group
+  forward or land a narrower single-oracle bridge that pins cross-group
+  timestamp allocation to one designated leader. Until such txns are enabled,
+  the per-group fix plus the shared-`*HLC` property remains adequate.
 
 ### (e) Cluster size / membership
 
@@ -393,8 +396,8 @@ The ordering is driven by unblock-edges, not by perceived value in isolation.
    change; closes the cross-group monotonicity gap (§1.5) *before* the
    topology that exposes it exists. Land first so each group remains safe when
    replicas move across nodes, but do not enable cross-group transactions whose
-   participating groups are led by different nodes until step 11 (or its
-   single-oracle bridge) lands.
+   timestamps can be allocated by more than one coordinator node until step 11
+   (or its single-oracle bridge) lands.
 2. **Multi-node multi-group bootstrap** (Gap 1 / **PR #955**,
    `2026_06_12_proposed_multinode_multigroup_bootstrap.md`). The root
    unblocker for (b), (c), (e), Gap 3, Gap 4. Nothing else multi-node-shaped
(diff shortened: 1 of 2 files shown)


bootjp:
@claude review

bootjp:
@claude review

Addressed in `09031487` (doc-only, `Status: Proposed` unchanged).

**Issue 1 — stale line anchors (fixed).** Re-verified against current `main`:
- keyviz sampler `§2(f)`: `observeMutation` is at `kv/sharded_coordinator.go:1841-1846` (was cited `1795-1824`, which now holds `rawLogs`/`txnLogs`). Corrected, and added the function name for stability.
- `RunHLCLeaseRenewal` `§1.5`: function is `kv/sharded_coordinator.go:1960-1985`, with `group, ok := c.groups[c.defaultGroup]` at `:1961` (was `1914-1953` / `:1915`). Corrected both citations.

The other spot-checked anchors (`main.go:746`, `multiraft_runtime.go:246-254`, `lsm_store.go`, `fsm.go`, etc.) re-verified accurate; left unchanged.

**Issue 2 — companion doc reference (softened, Option B).** The companion is PR #955 (`design/multinode-multigroup-bootstrap`), open and claude-approved but not on `main`. Softened all three references — `§2(b)`, `§3 Gap 1`, `§4 step 2` plus the in-flight-PR map line — to the same "in review as **PR #955** (`<file>`, branch `<branch>`)" form already used for #945/#951/#953. "Authoritative spec" is now conditioned on "once PR #955 lands."

**Issue 3 — OQ-1 / TSO sequencing (took the substantive route: sketched the commitTS logic AND annotated step 11).** Grounded in code:
- A cross-group txn issues both `startTS` (`nextStartTS`, `:1429`) and `commitTS` (`resolveTxnCommitTS` → `nextTxnTSAfter`, `:1102`/`:1376`) from a single `*HLC` (`c.clock`) on the coordinator's node. `maxLatestCommitTS`/`Observe` only floors on the keys the txn touches.
- Today this is safe only because every group shares one process-wide `*HLC`. The per-group renewal fix (step 1) buys per-*node* monotonicity only — TSO doc §6 "Guarantee" is explicit that cross-node monotonicity is not covered.
- So the moment step 2 makes the two groups of a cross-group txn coordinate on different nodes, two concurrent cross-group txns can draw `commitTS` from two independently-advanced clocks → non-comparable, breaking OCC. OQ-1 therefore cannot be answered affirmatively by a review deferred to "just before step 2."
- `§4 step 11` is re-labeled "deferred-pending-OQ-1-resolution, not settled-last," with the explicit trigger: **the first cross-group transaction whose participating groups are led by different nodes**. OQ-1 now also raises an interim option (pin cross-group ts allocation to one designated group's leader) to evaluate in the PR #955 design, short of the full TSO group.

**Inline findings:**
- **codex-connector (P2, learner reads, line 184) — fixed.** The doc said "learners forward `LinearizableRead` to the leader," which is wrong: `handleRead` returns `ErrNotLeader` for any non-leader state (`internal/raftengine/etcd/engine.go:1583`) and the *caller* must forward (`docs/raft_learner_operations.md`). Corrected, and noted Gap 3's follower-read design must supply the forwarding/proxy/replica-read API the primitive omits.
- **gemini (line 213, shared start timestamp) — folded in.** Added a note in `§2(d)` that the shared-snapshot invariant (one `startTS` per txn via `nextStartTS` → `reqs.StartTS`) is already upheld; the gap is cross-node `commitTS` comparability, not `startTS`.
- **gemini (line 399, live cutover / rolling upgrade) — folded in.** Added OQ-7 on the single-node→multi-node transition (live `AddVoter` expansion vs blue/green dual-write proxy vs fresh-cluster migration; version-skew/capability-gating), deferring the spec to PR #955 and pointing at the TSO doc §7 phased-cutover structure as the model.
- **gemini (line 309, interceptor/nil-safe pattern) and (line 356, layered mitigations / defer cluster-version Raft entries) — noted, not folded in.** These are implementation-pattern suggestions for the eventual Gap 3 follower-read design and the bootstrap rollout respectively; the roadmap explicitly defers mechanism to those docs and says it "does not duplicate the detail of the docs it references." Captured here so the downstream designs pick them up.

bootjp:
@codex review

Addressed the current-head roadmap review comments.

- narrowed Gap 1 to multi-group configs with per-group multi-node voter sets, while noting single-group multi-voter bootstrap already exists
- separated range ownership migration from replica placement in Gap 4, and made region balancing depend on both the M2 migration plane and a replica-placement / membership-change design

Verification:
- git diff --check

bootjp:
@codex review

Addressed the current rollout-guardrail review items:
- clarified that the per-group HLC ceiling fix is not sufficient to enable cross-group transactions led by different nodes; those remain gated on the TSO/single-oracle bridge.
- tightened in-place expansion to require AddLearner -> catch-up watermark -> promotion, reserving direct AddVoter for bootstrap/offline fully caught-up peers.

Validation: git diff --check.

bootjp:
@codex review

Addressed the new P1 guardrail findings:
- changed the TSO/single-oracle trigger from participant leader placement to timestamp allocation by multiple coordinator/ingress nodes, including forwarded requests that preserve non-zero timestamps.
- added the independent read-only-shard validation guardrail: a TSO/single-oracle bridge is necessary but not sufficient while validateReadOnlyShards has a TOCTOU window, so those txn shapes stay gated until a read-validate FSM phase or equivalent lands.

Validation: git diff --check; commit hook golangci-lint --config=.golangci.yaml run --fix.

bootjp:
@codex review

Addressed the current scaling-roadmap followups:
- renamed the proposal file/header to the add-commit date (`2026_06_23`)
- scoped the bootstrap limitation to multi-group configs; single-group multi-voter bootstrap is no longer described as unsupported
- clarified that M2 migration only reduces per-node data when replica placement differs, so Gap 1 / replica placement / region balance are prerequisites for actual volume relief
- added the keyviz read-path caveat for adapter-direct reads
- made Gap 1's rollout dependency on the HLC renewal fix explicit
- required region balance to depend on a replica-placement / membership-change design
- changed the cross-group bridge requirement to allocate both `startTS` and `commitTS`, and kept the read-only-shard validation gate explicit
- expanded capability checks to learners/new replicas and rephrased live expansion as learner-add/promote instead of direct live AddVoter

Validation: `git diff --check HEAD` passes. Docs-only change.

PR #989, opened 2026-07-03
Fix DynamoDB Jepsen transaction cancel classification

bootjp:
## Summary
- Treat DynamoDB `TransactionCanceledException` as an indeterminate Jepsen outcome in append workloads.
- Apply the same classification to the multi-table DynamoDB append workload.
- Add regression tests for the client exception handling path.
- Apply the gofmt correction required by the PR lint workflow.

## Root Cause
Scheduled run 28297833001 failed in the DynamoDB Jepsen workload with `:G1a` / `:dirty-update` anomalies. The history recorded a transaction as `:fail` with `TransactionCanceledException`, but later reads observed that transaction's appended value. Marking this response as a definite failed transaction made Elle assert the write was absent, producing a false positive.

## Tests
- `HOME=$(pwd)/tmp-home LEIN_HOME=$(pwd)/.lein LEIN_JVM_OPTS="-Duser.home=$(pwd)/tmp-home" /opt/homebrew/bin/lein test elastickv.dynamodb-workload-test elastickv.dynamodb-multi-table-workload-test`
- Rechecked the run 28297833001 DynamoDB history with `TransactionCanceledException` failures rewritten to `:info`: `{:valid? true}`
- `gofmt -l monitoring/hlc_test.go`
- `git diff --check`

### DIFF
--- a/jepsen/src/elastickv/dynamodb_multi_table_workload.clj
+++ b/jepsen/src/elastickv/dynamodb_multi_table_workload.clj
@@ -452,11 +452,9 @@
             (assoc op :type :info :error :network-error)
 
             (contains? #{"ConditionalCheckFailedException"
-                         "InternalServerError"} err-type)
+                         "InternalServerError"
+                         "TransactionCanceledException"} err-type)
             (assoc op :type :info :error (str err-type))
-
-            (= "TransactionCanceledException" err-type)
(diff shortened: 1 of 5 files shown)


bootjp:
@codex review

bootjp:
@codex review

PR #1071, opened 2026-07-11
kv: refine pebble backpressure guard

bootjp:
## Summary
- Refine the Pebble backpressure guard shared by the FSM compactor and lock resolver.
- Allow stable wide-L0 shapes, such as many L0 files and existing compaction debt with `L0 sublevels=1` and no active compactions, so maintenance does not skip forever.
- Keep yielding on high L0 sublevels and on high L0 file/debt pressure while Pebble compactions are active.
- Add the relevant Pebble pressure fields to skip logs.

## Tests
- `GOCACHE=$(pwd)/.cache GOTMPDIR=$(pwd)/.cache/tmp go test ./kv`
- `GOCACHE=$(pwd)/.cache GOTMPDIR=$(pwd)/.cache/tmp go test ./store ./internal/raftengine/...`
- `GOCACHE=$(pwd)/.cache GOTMPDIR=$(pwd)/.cache/tmp go test ./adapter -timeout=20m`
- `GOCACHE=$(pwd)/.cache GOTMPDIR=$(pwd)/.cache/tmp go test ./...` fails because `adapter` exceeds the default 10m package timeout; `./adapter` passes with `-timeout=20m`.

### DIFF
--- a/kv/compactor.go
+++ b/kv/compactor.go
@@ -17,6 +17,7 @@ const (
 	defaultFSMCompactorTimeout         = 5 * time.Second
 	defaultFSMCompactorLeaderTimeout   = 500 * time.Millisecond
 	defaultFSMCompactorMaxL0Files      = 256
+	defaultFSMCompactorMaxL0Sublevels  = 20
 	defaultFSMCompactorMaxLSMDebtBytes = 512 << 20
 )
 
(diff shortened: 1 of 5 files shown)


bootjp:
@codex[agent] review

PR #1115, opened 2026-07-18
Implement synchronous S3 blob offload

bootjp:
## Summary
- store content-addressed S3 chunk blobs outside Raft while committing chunk references and metadata atomically through Raft
- fsync the local blob and push to peer replicas concurrently, requiring the configured durable target with a two-copy floor
- proxy GET misses from the source peer and randomized alternates, verify SHA-256, and repair the local copy
- gate rollout on every Raft group endpoint capability with a 30-second cache and keep the feature disabled by default
- bound capability polling and peer blob RPCs so unavailable peers fail closed without indefinite request stalls

## Safety
- legacy BlobKey reads and writes remain available while the flag is disabled or any peer lacks capability
- authenticated peer RPC remains available during rollback so existing offloaded objects stay readable
- chunk references become reachable only in the same transaction as object or part metadata
- peer-local writes preserve the shared MVCC LastCommitTS watermark

## Verification
- make tla-tools
- make tla-check
- go test ./adapter -run ^Test\(S3Blob\|GRPCS3Blob\) -count=10 -timeout=300s
- go test ./kv -run ^TestShardedCoordinatorRaftMembersIncludesEveryGroupEndpoint$ -count=10 -timeout=180s
- go test . -run ^TestConfigureAdminService -count=5 -timeout=180s
- go test -race ./adapter ./kv . -run ^Test\(S3Blob\|GRPCS3Blob\|ShardedCoordinatorRaftMembers\|ConfigureAdminService\) -count=1 -timeout=300s
- golangci-lint run ./... --timeout=5m
- git diff --check

## Stack
Base: #1063


bootjp:
Current HEAD: `611737a0bb9abc0bb0a2e372cf3ee1b8dc1b32fe`

Fixed the current-head CI failure in `TestRuntimeServerRunner_PrepareAdminForwardServersDoesNotBindPublicListeners` by wiring the test runner with the same `startupGatedCoordinator` used by production startup. This keeps the production fail-closed membership-discovery requirement unchanged while making the listener test exercise the real runtime dependency shape.

Caller audit:
- `prepareAdminForwardServers`
- `newS3BlobCluster`
- `startupGatedCoordinator.RaftMembers`
- `startupGatedCoordinator.RaftMembersForKey`
- runtime runner construction in `startServers`

Verification:
- failing test x10
- root package full test
- focused race test
- full lint
- diff check
- signed commit verified for bootjp <contact@bootjp.me>

@codex review

bootjp:
Current HEAD: `395a2553ec964f23689b07eff1de8121a0749e04`

Addressed the current-head durability and metadata findings:
- multipart chunkrefs now use immutable attempt-versioned keys selected by the committed part descriptor
- chunkref V2 records the peers that durably acknowledged the blob; reads try those write-time replicas plus current membership
- learner acknowledgements no longer satisfy the voter quorum target
- offloaded chunkref proposals flush at `s3MetaBatchOps`; only the final tail and manifest/descriptor share the final transaction
- failed requests and object/multipart cleanup remove staged chunkrefs; content-addressed chunkblob abort orphans remain assigned to the M3 orphan scanner in design section 3.5

Caller audit covered `storeS3UploadPart`, `commitS3UploadPart`, multipart completion, object read/fetch, failed PUT/part cleanup, overwrite/delete/abort cleanup, `persistS3ChunkBlob`, `prepareS3BlobReplication`, and all chunkref encode/decode/key callers. Existing V1 values and unversioned manifests continue through version zero.

Verification:
- original listener CI regression test x10: pass
- focused S3/key tests x10: pass
- focused race: pass
- full lint: 0 issues
- `make tla-check`: pass
- diff check: pass
- full `adapter` package run reached its 600s timeout in the existing multi-node Raft integration tests; changed S3 tests and CI full race remain the authoritative gates
- signed commit verified as `bootjp <contact@bootjp.me>`

@codex review

bootjp:
Current HEAD: `e9efbb173ea61cb88bf8ec8b11256de5cc38e4c7`

Addressed both current-head findings from `395a2553ec`:
- blob fetch now preserves every current and write-time source endpoint, prioritizes the current source address, and deduplicates only identical `(node ID, address)` endpoints
- logical backup now routes `!s3|chunkblob|` and `!s3|chunkref|`, validates content hashes and declared sizes, distinguishes legacy `PartVersion` from offloaded `ChunkRefVersion`, and assembles mixed manifests fail-closed

Caller audit:
- `orderedS3BlobPeers` is called only by `fetchS3ChunkBlob`
- offload backup handlers are called only by the snapshot prefix dispatcher
- offload reference resolution feeds the existing manifest-driven `flushObjectWithCollision -> assembleObjectBody` path
- all `assembleObjectBody`, `filterChunksForManifest`, and `verifyChunkCompleteness` callers were audited

Verification:
- original listener CI regression test x10: pass
- focused adapter S3 tests x10: pass
- full `internal/backup` suite: pass
- focused race: pass
- full lint: 0 issues
- `make tla-check`: pass
- `git diff --check`: pass
- aggregate `go test ./...` passed changed packages but hit a timing-sensitive `internal/raftengine/etcd` failure under full parallel load; isolated package rerun passed
- signed commit verified as `bootjp <contact@bootjp.me>`

@codex review

bootjp:
Current HEAD: `635c5299426c4fc4bc168e73bd2c968f2609a6b1`

Addressed the five findings from the `e9efbb17` review:
- `AdminGetObject` now uses the shared offloaded chunk read and proxy-on-miss path
- M1 waits for every current Raft member, including learners, before publishing a chunkref, so every member-local logical backup is self-contained; M2 may restore quorum acknowledgement after apply/backfill is available
- logical backup now validates manifest `ChunkSizes` against chunkref size and chunkblob size
- production offload remains fail-closed until M3 GC workers set runtime readiness, even when `ELASTICKV_S3_BLOB_OFFLOAD=true`
- full chunkref batches remain immutable staged metadata by design section 3.2; only the final manifest or part descriptor makes them discoverable, and failed requests remove the staged attempt prefix

Caller audit:
- durability target: `persistS3ChunkBlob -> prepareS3BlobReplication -> s3BlobDurableTarget`
- object reads: public GET range hydration and admin GET both use `readS3ObjectChunk`
- backup: snapshot prefix dispatch feeds `HandleObjectManifest` and `resolveOffloadedChunkPaths`; legacy parts are unchanged
- GC readiness is read only by `s3BlobOffloadDecision`; the default remains false and tests inject readiness explicitly

Verification:
- focused S3, backup, and original listener regression tests x10: pass
- focused race: pass
- full `internal/backup` suite: pass
- full lint: 0 issues
- `make tla-check`: all outcomes match the design contract
- `git diff --check`: pass
- local full `adapter` suite reached its existing package-wide 10 minute timeout; current-head CI Test/race is the authoritative full-suite gate
- signed commit verified as `bootjp <contact@bootjp.me>`

@codex review

PR #1116, opened 2026-07-18
jepsen: preserve uncertain ZSet dependencies

bootjp:
## Summary
- Preserve pre-read `:info` mutations when an intervening ZINCRBY or ZREM may depend on them.
- Treat only an unconditional committed ZADD as a superseding reset when no state-dependent mutation can linearize in between.
- Add a regression test for an uncertain ZADD feeding a committed ZINCRBY before a later ZADD.

## Root cause
The ZSet safety checker discarded an uncertain mutation whenever a later committed state change existed. In the failing scheduled history, an uncertain ZADD was required to explain an intervening successful ZINCRBY reply. Removing it made the mutation chain impossible and caused 157 false `unexpected-presence` errors.

## Impact
Valid ZSet histories containing response-loss uncertainty are no longer rejected, while later ZADDs still supersede uncertainty when no state-dependent operation can rely on it.

## Tests
- `cd jepsen && lein test` (143 tests, 335 assertions)
- Replayed the failing run artifact (493 reads, 971 mutations, 0 errors, `valid? true`)
- Commit hook `golangci-lint --config=.golangci.yaml run --fix` (0 issues)
- `git diff --check`

### DIFF
--- a/jepsen/src/elastickv/redis_zset_safety_workload.clj
+++ b/jepsen/src/elastickv/redis_zset_safety_workload.clj
@@ -581,25 +581,36 @@
   (and (some? (:complete-idx earlier))
        (> (:invoke-idx later) (:complete-idx earlier))))
 
-(defn- superseded-by-preceding-state-change?
-  "True iff a committed mutation before the read strictly follows m and
-  changes this member's state. Such a later committed write/remove
-  determines the read's base state and makes m's pre-read uncertainty
-  irrelevant for this read.
-
-  Relative increments are not absolute overwrites. Earlier uncertainty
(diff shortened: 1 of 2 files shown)


bootjp:
@codex review

Please review the current HEAD `e57d74e2a90649f186e0817bd505157015035642`, especially the real-time ordering in `could-linearize-between?` and whether preserving uncertain ZADD/ZREM dependencies can admit impossible histories. CI is green, all 143 Jepsen unit tests pass, and replaying the failing artifact now yields 493 reads / 971 mutations / 0 errors.

PR #1135, opened 2026-07-22
Avoid full scans for Lua stream appends

bootjp:
## Summary
- Avoid full stream entry scans for Lua XADD and XLEN when the stream is already in the wide-column layout.
- Track stream metadata, appended entries, and bounded head trims inside the Lua script context, then commit only the delta where safe.
- Materialize the full stream only for Lua commands that need arbitrary entry access, such as XRANGE and XREVRANGE.
- Add OCC read fences so absent-stream creation still rejects concurrent cross-type creation.

## Root cause
The production CPU profile under Redis proxy traffic showed Lua EVAL/EVALSHA dominated by streamState -> loadStreamAt -> scanStreamEntriesAt. XADD in Lua was loading every existing stream entry before appending one row, which turned Redis proxy secondary replay into O(stream length) scans.

## Validation
- go test ./adapter -run TestLuaStream
- go test ./adapter -run 'TestRedisEval|TestRedisLua|TestRedisScript|TestRedisStream|TestRedisX'
- go test -race ./adapter -run 'TestLuaStream'
- go test ./... -run '^$'
- golangci-lint run ./adapter --timeout=5m
- go test ./adapter was attempted in the original worktree but stopped after 449s with no completion; targeted adapter coverage and full-package compile passed in this branch.

### DIFF
--- a/adapter/redis_lua_context.go
+++ b/adapter/redis_lua_context.go
@@ -34,6 +34,7 @@ type luaScriptContext struct {
 	redisCallCount    int
 
 	touched     map[string]struct{}
+	readKeys    map[string][]byte
 	deleted     map[string]bool
 	everDeleted map[string]bool
 	// negativeType caches keys for which server.keyTypeAt() has already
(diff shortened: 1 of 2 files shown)


bootjp:
@codex review

bootjp:
@codex review

PR #1136, opened 2026-07-22
Add Redis delta compactor toggle

bootjp:
## Summary
- add ELASTICKV_REDIS_DELTA_COMPACTOR_ENABLED to disable the Redis DeltaCompactor background loop
- skip RedisServer compactor wiring when the toggle is false so urgent compaction requests are ignored
- keep the default enabled to preserve existing behavior

## Risk
- disabling the compactor can let Redis collection delta keys accumulate, so it should be used as an operational relief switch while investigating scan load

## Validation
- go test .
- golangci-lint run . --timeout=5m

Author: bootjp


bootjp:
@codex review

PR #1143, opened 2026-07-22
Cap Redis blocking replay pressure

bootjp:
## Summary
- Cap derived secondary blocking replay concurrency instead of using all remaining backend connections
- Add a short delay before BZPOP ZREM replay and bound no-effect retries
- Count blocking replay deadline expiry as a replay miss instead of a secondary write failure

## Tests
- go test ./cmd/redis-proxy -count=1
- go test ./proxy -count=1
- golangci-lint run ./proxy ./cmd/redis-proxy --timeout=5m


bootjp:
@codex review

bootjp:
Addressed the BZPOP replay review feedback. The initial delay is skipped when the remaining secondary deadline is too short, so short timeouts still issue at least one ZREM. No-effect retries are now derived from the remaining deadline and capped by a bounded retry window instead of a fixed attempt count.

@codex review

bootjp:
Addressed the retry-window budget finding by counting the maximum jittered backoff duration before allowing each no-effect retry. Added a unit test that verifies the computed retry budget stays within the configured blocking replay window.

@codex review

PR #1145, opened 2026-07-22
Preserve empty Lua stream metadata

bootjp:
## Summary
- Keep Lua `XTRIM MAXLEN 0` as an empty stream instead of treating it as a logical delete.
- Preserve stream TTL, last generated ID, and trim cursor metadata on the empty stream path.
- Add a regression test covering a future last ID, TTL preservation, stream type preservation, and the next `XADD *` result.

## Root cause
Lua stream trimming had a special `MAXLEN 0` branch that cleared the stream state, marked the key deleted, and cleared TTL metadata. That diverged from the normal Redis stream `XTRIM` path, which keeps the stream metadata record even when all entries are removed.

## Tests
- `GOCACHE=$(pwd)/.gocache GOTMPDIR=$(pwd)/.gotmp go test ./adapter -run 'TestLuaStreamXTrimMaxLenZeroKeepsEmptyStreamMetadata|TestLuaStreamXAddMaxLenZeroKeepsLastIDWithoutRows|TestLuaStreamXAddMaxLenUsesTrimCursorForNextHeadScan' -count=1`
- `GOCACHE=$(pwd)/.gocache GOTMPDIR=$(pwd)/.gotmp go test ./adapter -run 'TestLuaStream|TestTrimLuaStreamHead' -count=1`
- `GOCACHE=$(pwd)/.gocache GOTMPDIR=$(pwd)/.gotmp go test ./adapter -run 'TestLua|TestRedisStream|TestXAdd|TestXTrim|TestXRange|TestXRead' -count=1`

### DIFF
--- a/adapter/redis_lua_context.go
+++ b/adapter/redis_lua_context.go
@@ -3490,21 +3490,7 @@ func (c *luaScriptContext) cmdXTrim(args []string) (luaReply, error) {
 		return luaIntReply(0), nil
 	}
 	removed := st.meta.Length - int64(maxLen)
-	if maxLen == 0 {
-		st.value.Entries = nil
-		st.exists = false
-		c.deleted[string(key)] = true
-		c.everDeleted[string(key)] = true
-		c.clearTTL(key)
-		st.materialized = true
-		st.requiresFullRewrite = true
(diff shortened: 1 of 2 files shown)


bootjp:
@codex review

